PR #89, opened 2026-02-01
fix(mcp): resolve re-exported symbol usages in get_dependents

### LINKED ISSUE #86: get_dependents: symbol usage count always 0 for re-exported symbols
## Description

When using `get_dependents` with a `symbol` parameter, `totalUsageCount` is always `0` even when the symbol is widely used across the codebase.

## Steps to Reproduce

```
get_dependents({ filepath: "packages/core/src/vectordb/lancedb.ts", symbol: "VectorDB" })
```

**Expected:** `totalUsageCount` reflects actual usage (10+ files import and use `VectorDB`)
**Actual:** `totalUsageCount: 0`, though the file-level `dependents` list is correct (9 dependents found)

## Analysis

The likely cause is that most files import `VectorDB` from `@liendev/core` (the package re-export) rather than from the direct path `../vectordb/lancedb.ts`. The symbol resolver doesn't follow re-export chains to match usage back to the source file.

```
grep confirms usage:
- packages/cli/src/cli/complexity.ts: import { VectorDB } from '@liendev/core'
- packages/cli/test/helpers/test-db.ts: import { VectorDB } from '@liendev/core'
- packages/cli/test/e2e/real-projects.test.ts: import { VectorDB, ... } from '@liendev/core'
- packages/cli/test/benchmarks/performance.test.ts: import { VectorDB, ... } from '@liendev/core'
```

## Impact

Symbol-level impact analysis is effectively broken for any symbol accessed through package re-exports, which is the common pattern in monorepo setups.

Found during dogfooding session (2026-02-01).

alfhen:
## Summary

- Fixes `get_dependents` returning `totalUsageCount: 0` for symbols accessed through re-exports (e.g., `import { VectorDB } from '@liendev/core'` instead of directly from the source file)
- Adds re-export chain resolution: detects barrel/index files that re-export the target symbol, then finds indirect consumers that import the symbol by name
- Adds unit tests for the dependency analyzer's re-export resolution logic

## How it works

The fix adds a second phase to `findSymbolUsages`:

1. **Phase 1 (existing):** Find direct dependents where the import path matches the target file
2. **Phase 2 (new):** If any direct dependent also exports the target symbol (i.e., it's a re-exporter/barrel file), scan all chunks for additional files that import the symbol by name from any path

This handles the common monorepo pattern where symbols are accessed through package entry points (`@scope/pkg`) rather than direct file paths.

## Test plan

- [x] Typecheck passes (`npm run typecheck`)
- [x] Build succeeds (`npm run build`)
- [x] All 517 tests pass (511 existing + 6 new)
- [ ] Manual verification with `get_dependents({ filepath: "packages/core/src/vectordb/lancedb.ts", symbol: "VectorDB" })`

Closes #86

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

### 👁️ Veille

*[Veille](https://lien.dev) by Lien*

### DIFF
--- a/packages/cli/src/mcp/handlers/dependency-analyzer.ts
+++ b/packages/cli/src/mcp/handlers/dependency-analyzer.ts
@@ -543,7 +543,7 @@ function findSymbolUsages(
 
   // Phase 2: Indirect dependents through re-export chains
   const indirect = collectReExportedSymbolDependents(
-    chunksByFile, allChunks, targetSymbol, direct.processedFiles, log
+    chunksByFile, allChunks, targetSymbol, normalizedTarget, normalizePathCached, direct.processedFiles, log
   );
 
   return {
@@ -595,10 +595,12 @@ function collectReExportedSymbolDependents(
   chunksByFile: Map<string, SearchResult[]>,
   allChunks: SearchResult[],
   targetSymbol: string,
+  normalizedTarget: string,
+  normalizePathCached: (path: string) => string,
   processedFiles: Set<string>,
   log: (message: string, level?: 'warning') => void
 ): { dependents: DependentInfo[]; totalUsageCount: number } {
-  const reExporterPaths = findReExporterPaths(chunksByFile, targetSymbol);
+  const reExporterPaths = findReExporterPaths(chunksByFile, targetSymbol, normalizedTarget, normalizePathCached);
   if (reExporterPaths.size === 0) {
     return { dependents: [], totalUsageCount: 0 };
   }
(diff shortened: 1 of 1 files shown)


alfhen:
Closing this PR — the re-export chain resolution adds significant complexity (~700 lines) for minimal practical value. The `totalUsageCount` is informational context for AI agents, and in practice agents read the code directly regardless of the count.

Instead, we'll document that symbol tracking only works for direct imports as a known limitation, and focus on fixing the underlying issues (#90, #91).

Issue #556, opened 2026-05-15
Index does not refresh on git branch switch — stale chunks for files not in current working tree


alfhen:
## Context

While using the Lien MCP tools (`semantic_search`, `list_functions`, `find_similar`, `get_files_context`) across multiple feature branches, I noticed that switching branches with `git checkout` does not invalidate or refresh the index. Chunks indexed while on the previous branch keep surfacing in tool results, including chunks from files that don't exist on the new branch's working tree.

## Reproduction

1. Start with a clean working tree on branch `A`. Let the MCP server build/refresh its index.
2. On branch `A`, a file `foo/bar.py` exists with some exported symbol `quux`.
3. `git checkout B` where branch `B` does **not** have `foo/bar.py` (file added on `A` only, or deleted on `B`).
4. From the new working tree (where `ls foo/bar.py` fails), run any of:
   - `semantic_search` with a query that previously surfaced `quux`
   - `list_functions` with a pattern that matched `quux`
   - `find_similar` against a snippet that resembles content from `foo/bar.py`

Observed: tool results still contain chunks from `foo/bar.py`, including `metadata.file` pointing at the file path and `startLine`/`endLine` that referenced its previous content. `get_files_context` on the missing path also returns its old chunks.

## What `indexInfo` looks like in this state

This is what surprised me — the index reports as healthy at first glance:

- `reindexInProgress: false`
- `pendingFileCount: 0`
- `lastReindexDurationMs`: populated (small, e.g. ~300ms)
- `msSinceLastReindex`: relatively recent (single-digit minutes)
- `indexDate`: an older date than today (matches the day the previous branch was active)

So the metadata suggests a recent reindex happened, but the underlying chunks have not been invalidated for files that left the working tree on branch switch.

## Expected

One of:

1. **Auto-detect the branch switch** (e.g. watch `.git/HEAD` for ref changes) and reconcile: re-index files whose content changed, drop chunks for files that no longer exist on the new ref.
2. **Expose a manual reindex tool via MCP** so the caller can force `Reindex now` when it knows the underlying state changed. (Currently I don't see one listed in the available MCP tools.)
3. **Surface the indexed branch / ref in `indexInfo`** so callers can detect a mismatch and react (re-run a tool with caveat, warn the user, etc.).

## Why it matters

AI coding assistants act on the file paths and line numbers returned by these tools. When the index keeps chunks for files that no longer exist on the current checkout, the assistant can:

- Recommend edits to files that aren't in the tree.
[...]

PR #562, opened 2026-05-15
fix(mcp): advance indexInfo.indexDate on incremental reindexes

alfhen:
Follow-up to [#556](https://github.com/getlien/lien/issues/556) — addresses the cosmetic [residual flagged in the closing comment](https://github.com/getlien/lien/issues/556#issuecomment-4458655180).

## Summary

After the v1/v2/v3 fixes landed, the reporter confirmed branch-switch reconciliation works end-to-end. One residual: \`indexInfo.indexDate\` stayed at its original value across reindexes, even though \`indexedCommit\`, \`lastReindexDurationMs\`, and \`msSinceLastReindex\` were correctly advancing.

## Root cause

\`indexDate\` was sourced from \`vectorDB.getVersionDate()\` → \`new Date(currentVersion).toLocaleString()\` (\`packages/core/src/vectordb/lancedb.ts:249\`). \`currentVersion\` is the numeric timestamp in \`.lien-index-version\`, written by \`writeVersionFile()\` only at the end of a **full** \`indexCodebase\` run (\`packages/core/src/indexer/index.ts:213,466\`). The incremental reindexes that file-watcher and the always-on git poll fire (\`indexSingleFile\` / \`indexMultipleFiles\`) intentionally don't bump that file — it's a cross-process invalidation token and bumping it on every change would cause needless reconnect storms in other clients tracking the same index.

So \`indexDate\` reflected "when was the index last fully rebuilt", not "when was content last reconciled". Two genuinely different concepts mashed into one display field.

## Fix

In \`getIndexMetadata\` (\`packages/cli/src/mcp/server.ts\`), prefer the existing \`reindexStateManager.lastReindexTimestamp\` (which tracks every reindex op, full or incremental, and is already used to derive \`msSinceLastReindex\`). Fall back to \`vectorDB.getVersionDate()\` pre-first-reindex.

\`indexVersion\` stays put — it's the cross-process token. Diverging semantics from \`indexDate\` is intentional: version = schema token, date = content freshness. The latter is what humans / AI assistants actually want to see.

## Test plan

- [x] \`server.test.ts\` gains two cases under \`getIndexMetadata indexDate source\`:
  - uses \`lastReindexTimestamp\` when set (proves the new path)
  - falls back to \`getVersionDate()\` when null (regression guard for first-boot)
- [x] \`npm run typecheck\` clean
- [x] \`npm run lint\` (0 errors)
- [x] \`npm run format:check\` clean
- [x] \`npm run test -w packages/cli\` — all green (17 tests in \`server.test.ts\`)
- [ ] Manual: dogfood loop → after a checkout-triggered reindex, \`indexInfo.indexDate\` should reflect "now" (matching the \`msSinceLastReindex\` window) instead of the original version-file timestamp.

## Out of scope

- Changing \`writeVersionFile\` semantics (would cause reconnect storms in cross-process consumers).
- Renaming \`indexDate\` to something more explicit. Possible follow-up but doesn't help any user today.

---

### Lien Review

> [!NOTE]
> **Low Risk**
>
> The PR correctly addresses the issue where `indexInfo.indexDate` remained stale during incremental reindexes. By taking the maximum of the version-file timestamp and the in-session reindex timestamp, the metadata now accurately reflects the most recent reconciliation regardless of whether it was a full or incremental update. The implementation includes robust fallbacks for cases where no reindex has occurred yet, and the added tests verify both the preferred path and the fallback logic.
>
> - Updated `getIndexMetadata` to use the maximum of `vectorDB.getCurrentVersion()` and `reindexStateManager.lastReindexTimestamp` as the effective index date.
> - Ensured `msSinceLastReindex` is calculated from the same effective timestamp, providing accurate freshness info even after server restarts.
> - Added comprehensive unit tests in `server.test.ts` covering various scenarios including external full reindexes and first-boot fallbacks.

<sup>Reviewed by [Lien Review](https://lien.dev) for commit 223fe38. Updates automatically on new commits.</sup>

### DIFF
--- a/packages/cli/src/mcp/server.test.ts
+++ b/packages/cli/src/mcp/server.test.ts
@@ -283,36 +283,76 @@ describe('startMCPServer', () => {
     });
   });
 
-  describe('getIndexMetadata indexDate source', () => {
-    it('uses lastReindexTimestamp when set (advances on incremental reindexes)', async () => {
-      const fixedTimestamp = 1714521600000; // 2024-05-01T00:00:00Z
+  describe('getIndexMetadata indexDate / msSinceLastReindex sources', () => {
+    function withReindexTimestamp(ts: number | null) {
       vi.mocked(createReindexStateManager).mockReturnValueOnce({
         getState: vi.fn().mockReturnValue({
           inProgress: false,
           pendingFiles: [],
-          lastReindexTimestamp: fixedTimestamp,
-          lastReindexDurationMs: 200,
+          lastReindexTimestamp: ts,
+          lastReindexDurationMs: ts === null ? null : 200,
         }),
         startReindex: vi.fn(),
         completeReindex: vi.fn(),
         failReindex: vi.fn(),
       } as unknown as ReturnType<typeof createReindexStateManager>);
+    }
+
+    it('uses lastReindexTimestamp when it is the most recent reconciliation', async () => {
+      const reindexTs = 1714521600000; // 2024-05-01
+      mockVectorDB.getCurrentVersion.mockReturnValue(1); // very old
+      withReindexTimestamp(reindexTs);
 
       await startMCPServer({ rootDir: '/test/project' });
 
       const toolContext = vi.mocked(registerMCPHandlers).mock.calls[0][1];
       const metadata = toolContext.getIndexMetadata();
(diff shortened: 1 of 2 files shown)


alfhen:
@coderabbit run

PR #566, opened 2026-05-15
feat(plugin): edit-gate hook — enforce Lien impact analysis before Edit/Write

### LINKED ISSUE #560: Ship a PreToolUse hook with the plugin: gate Edit/Write on Lien-indexed files until impact analysis has been done
## Context

The MCP-server instructions block already declares:

> **REQUIRED before Edit/Write on any file:** `get_files_context({ filepaths })`
> **REQUIRED before renaming/removing/changing the signature of any exported symbol:** `get_dependents({ filepath, symbol })`

In practice these are advisory — the model frequently goes straight to `Edit`/`Write`. A `PreToolUse` hook shipped *with the plugin* converts the "REQUIRED" wording into an actual gate, with zero per-user `~/.claude/settings.json` wiring.

The intent is **"did you do impact analysis on this file?"**, not **"did you call this specific tool?"** — three Lien tools can satisfy it.

## Proposal

Ship a plugin-level hook bundle (PreToolUse + PostToolUse + SessionStart) that:

1. On `Edit`/`Write`/`MultiEdit`, reads `tool_input.file_path` and `session_id` from stdin.
2. Skips if extension is outside Lien's indexed-extension set (no friction on `.md`/`.yml`/`.json`).
3. Looks up a session sentinel under `~/.lien/indices/<repoId>/gate-sessions/<session_id>/`:
   - existing-file edit: any of `fc-<hash>` / `dep-<hash>` / `fs-<hash>` for that path
   - new-file write: any recent `fs-*` (loose by design)
4. Applies a TTL (default 60 min) so stale sentinels don't satisfy the gate.
5. If unsatisfied → emits `systemMessage` naming the call that would satisfy it; advisory by default (exit 0 + stderr), blocking opt-in via plugin setting.

`PostToolUse` matched against `mcp__plugin_lien_lien__(get_files_context|get_dependents|find_similar)` writes the sentinels. `SessionStart` garbage-collects any `gate-sessions/<id>/` whose id ≠ the current one.

## Decisions

- **Storage path** — `~/.lien/indices/<repoId>/gate-sessions/<session_id>/`. `repoId` derives deterministically via `extractRepoId(cwd)` from `@liendev/parser` (`<basename>-<md5(path)[:8]>`). Reuses the same path the vectorDB already creates. Will add `lien path --store` (one-line subcommand) so the shell script doesn't reimplement repoId derivation.
- **Scope** — per-repo, keyed off cwd. Matches index scope; session sub-dirs avoid cross-session bleed.
- **Default mode** — **advisory** (exit 0, `systemMessage` only). Blocking is one config flag away. Rationale: a noisy first release will get the plugin uninstalled; advisory still reaches the model via `systemMessage`.
- **Override** — `LIEN_GATE=off` env var (one-shot for the next call) plus a `/lien-gate off` slash command shipped with the plugin for the duration of the session. No magic comments (fragile).
- **v2 deferred** — symbol-level demand on `Edit` (`get_dependents` for exported identifiers in `old_str`); tightening of `find_similar` looseness for new files.

## v1 scope (PR-ready)

- `plugins/claude/hooks/hooks.json` registering `PreToolUse` (matcher `Edit|Write|MultiEdit`), `PostToolUse` (matcher `mcp__plugin_lien_lien__(get_files_context|get_dependents|find_similar)`), `SessionStart`.
- `plugins/claude/hooks/gate.sh` — PreToolUse logic; reads `tool_input.file_path` + `session_id`; respects `LIEN_GATE=off`.
- `plugins/claude/hooks/sentinel.sh` — PostToolUse; parses `tool_input` (and for `find_similar`, scans `tool_result` JSON for cited filepaths).
- `plugins/claude/hooks/session-clean.sh` — SessionStart; removes stale `gate-sessions/<id>/`.
- `lien path --store` and `lien path --extensions` CLI subcommands.

## Verified against installed-plugin examples

- Plugins **can** ship hooks (`hooks` key in `plugin.json` or `hooks/hooks.json` wrapper). `${CLAUDE_PLUGIN_ROOT}` resolves at runtime.
- `SessionStart`/`SessionEnd` are both valid events.
- `PreToolUse` payload contains `session_id`, `cwd`, `tool_name`, `tool_input.file_path`.
- `PostToolUse` matchers handle `mcp__*` regex (see `security-guidance` and `hookify` plugins).

## Simpler alternatives considered

- **Sharper MCP instructions** — already tried; model still skips. Rejected as standalone, kept as a parallel improvement.
- **Status-line nag** — visible to the human, not the model. Doesn't change behavior on the next tool call. Useful adjacent; not a substitute.
- **`Stop` hook reminder** — fires too late; the edits already shipped.

## Risks & escape hatches

- **Misfire on hook self-edits / non-indexed dirs** → extension filter + `LIEN_GATE=off`.
- **False-positive rate** — estimate: in a 20-edit session, 2–4 trip without a satisfying call (most edits are preceded by a read/grep already). Advisory default keeps the cost low.
- **Stale-index drift** — orthogonal; tracked in #556.

## Related

- Existing MCP "REQUIRED" wording — this enforces it.
- #137 (closed), #556 — index staleness on git ops; same reliability cluster.

alfhen:
Closes #560.

## Summary

Ships a `PreToolUse`/`PostToolUse`/`SessionStart` hook trio with the Claude Code plugin that nudges the model to run `get_files_context` / `get_dependents` / `find_similar` before `Edit`/`Write` on Lien-indexed files.

**Blocking by default** — discovered during live dogfooding that Claude Code 2.1.142 records a PreToolUse `systemMessage` to the transcript but does not surface it to the model on subsequent turns. Only exit 2 + stderr (the blocking path) reaches the model. Advisory mode is preserved as an explicit opt-down for users who only want the UI signal.

State lives under `~/.lien/indices/<repoId>/gate-sessions/<session_id>/`, reusing `extractRepoId` after walking up for `.git` (so behavior is stable from any subdirectory). `SessionStart` GCs sessions idle >24h; concurrent Claude Code sessions for the same repo don't clobber each other.

Three new CLI subcommands expose the path/state machinery to the shell scripts:

- `lien path --store | --extensions | --root` — single source of truth for the bash hooks.
- `lien gate on | block | advisory | off | status` — toggle and inspect the gate; `/lien-gate` slash command wraps it.

## What ships

| Path | Purpose |
| --- | --- |
| `plugins/claude/hooks/hooks.json` | Registers the three hook events |
| `plugins/claude/hooks/gate.sh` | PreToolUse gate: TTL-bounded sentinel lookup, ext filter, blocking by default |
| `plugins/claude/hooks/sentinel.sh` | PostToolUse writer: `fc-<hash>` for `get_files_context`, `dep-<hash>` for `get_dependents`, `fs-any` for `find_similar`. Validates `session_id`, skips when `tool_response.isError` is true |
| `plugins/claude/hooks/session-clean.sh` | SessionStart GC: removes session dirs idle >24h; clears `gate-disabled` |
| `plugins/claude/commands/lien-gate.md` | `/lien-gate` slash command |
| `packages/cli/src/cli/path-cmd.ts` | `lien path` |
| `packages/cli/src/cli/gate-cmd.ts` | `lien gate` |
| `packages/cli/src/cli/store-paths.ts` | Shared `getStoreRoot()` (used by both commands) |
| `packages/cli/src/cli/project-root.ts` | `resolveProjectRoot()` — walks up for `.git` |

## Test plan

- [x] `npm run format:check`
- [x] `npm run lint` (0 errors; pre-existing warnings unchanged)
- [x] `npm run typecheck`
- [x] `npm run build`
- [x] `npm run test -w packages/cli` (664/664), `-w @liendev/parser` (747/747), `-w @liendev/review` (348/348). Core has one pre-existing flake in `embeddings/worker-embeddings.test.ts` (worker exit), unrelated to this change.
- [x] Shell smoke tests on each hook script (indexed/non-indexed extension, env overrides, sentinel satisfied/aged-out, sentinel-writer paths, SessionStart GC, multi-file `get_files_context`, traversal-attempt `session_id` rejected, `isError: true` suppresses sentinel).
- [x] **Live dogfooding through Claude Code** (see findings below).
- [ ] **Reviewer**: install/refresh the plugin locally and confirm the gate fires on a fresh `Edit`. Note that `/reload-plugins` reads from `~/.claude/plugins/cache/...`, not the source directory — to test changes from this PR you'll need to either `/plugin uninstall lien && /plugin install lien@lien` or copy hook files into the cache snapshot.

## Live dogfooding findings

Six findings from running the gate end-to-end inside a real Claude Code session. Each is reflected in the current PR head.

1. **`systemMessage` is silent to the model.** Probed via a discriminating PostToolUse hook on `Read` that emitted a directive ("emit a specific sentinel string"). The marker landed in the transcript as a `hook_system_message` attachment but never appeared in the model's next-turn input, and the model did not emit the string. Same channel mechanism is used by gate.sh in advisory mode — so the original "advisory by default" shipped a hook that the model didn't actually see. Default flipped to blocking.

2. **Path canonicalization mismatch was a real bug.** Initial design hashed file paths verbatim. Claude Code passes `tool_input.file_path` as **absolute**; the model passes relative paths to MCP tools. Hashes never matched → gate fired even after a satisfying call. Fixed by normalizing both sides to a project-root-relative form before hashing (commit `83db553`).

3. **Project-root resolution.** Running `lien gate off` from a subdir wrote the flag into a different store than the hook (which uses Claude Code's `cwd`) read from. `resolveProjectRoot()` walks up for `.git`; `lien path --root` exposes it to shell scripts. Hooks now canonicalize against `$root` with `$cwd` as fallback (commit `4b96691`).

4. **Multi-session safety.** Original `SessionStart` cleanup wiped all other-session dirs unconditionally. With two concurrent Claude Code sessions on the same repo, the second's startup would have destroyed the first's sentinels. Now only sessions idle >24h get GC'd (commit `c0f7b73`).

5. **`fc-*` was satisfying the new-file gate.** The model calls `get_files_context` constantly for context-gathering, so unrelated `fc-*` sentinels were silently satisfying the new-file branch and defeating the `find_similar` nudge. Tightened to `fs-*` only (commit `0d1da1e`). Paired with an `fs-any` marker so an empty `find_similar` result doesn't cause an infinite nudge loop.

6. **Per-cited `fs-<hash>` from `find_similar` was dead code.** Live dogfooding showed Claude Code's PostToolUse payload for MCP tool results doesn't use the `tool_response.content[0].text` shape sentinel.sh was reading. Only `fs-any` was getting written in practice. The per-cited semantic was also conceptually weak — `find_similar` is search, not impact analysis on a specific file — so dropped entirely (commit `1f0851f`).

## CodeRabbit feedback addressed

Fixed: filesystem try/catch in `gateCommand`; shared `getStoreRoot()` (also Windows-safe via forward-slash normalization); `session_id` sanitization in all three hooks; `isError` check before sentinel write; test isolation via mocked `os.homedir()`; tightened new-file gate; relative-vs-cwd preference for `target_path` resolution.

Skipped (with reasons): try/catch around `extractRepoId`/`getSupportedExtensions` (don't throw); `Bash(lien gate *)` rewrite (CodeRabbit wrong — every installed plugin uses the colon form); `MultiEdit` payload "iteration" (MultiEdit is multi-edit-single-file, `tool_input.file_path` is the right field); null-terminated jq for newline-in-filename (vanishingly rare; the literal kept being collapsed and broke the multi-path case for normal users).

## Notes for review

- **Default is blocking.** Users who want UI-only (no enforcement) can run `lien gate advisory`. Users who want it off entirely: `lien gate off`.
- **`jq` is a soft prereq.** Every hook checks `command -v jq` and silently exits 0 if missing — the gate degrades to a no-op rather than failing edits.
- **Cache-refresh wrinkle.** `/reload-plugins` reads cached snapshots; full `/plugin uninstall && /plugin install` (or marketplace refresh) is needed to pick up hook source changes during dev.

### DIFF
--- a/packages/cli/src/cli/gate-cmd.test.ts
+++ b/packages/cli/src/cli/gate-cmd.test.ts
@@ -19,14 +19,14 @@ describe('gateCommand', () => {
   let tmpHome: string;
   let STORE: string;
   let DISABLED: string;
-  let BLOCKING: string;
+  let ADVISORY: string;
+  let LEGACY_BLOCKING: string;
   let logSpy: ReturnType<typeof vi.spyOn>;
   let errSpy: ReturnType<typeof vi.spyOn>;
   let exitSpy: ReturnType<typeof vi.spyOn>;
   let homeSpy: ReturnType<typeof vi.spyOn>;
 
   beforeEach(async () => {
-    // Redirect $HOME so the test never touches the developer's real ~/.lien.
     tmpHome = await fs.mkdtemp(path.join(os.tmpdir(), 'lien-gate-test-'));
     homeSpy = vi.spyOn(os, 'homedir').mockReturnValue(tmpHome);
 
@@ -37,7 +37,8 @@ describe('gateCommand', () => {
       extractRepoId(resolveProjectRoot(process.cwd())),
     );
     DISABLED = path.join(STORE, 'gate-disabled');
-    BLOCKING = path.join(STORE, 'gate-blocking');
+    ADVISORY = path.join(STORE, 'gate-advisory');
+    LEGACY_BLOCKING = path.join(STORE, 'gate-blocking');
 
     logSpy = vi.spyOn(console, 'log').mockImplementation(() => undefined);
     errSpy = vi.spyOn(console, 'error').mockImplementation(() => undefined);
@@ -54,37 +55,64 @@ describe('gateCommand', () => {
     await fs.rm(tmpHome, { recursive: true, force: true });
   });
 
-  it('off writes gate-disabled and clears gate-blocking', async () => {
-    await fs.mkdir(STORE, { recursive: true });
-    await fs.writeFile(BLOCKING, '');
-    await gateCommand('off');
-    expect(await exists(DISABLED)).toBe(true);
-    expect(await exists(BLOCKING)).toBe(false);
-  });
-
-  it('on clears both flags', async () => {
+  it('on clears every flag (blocking is the default, no flag needed)', async () => {
     await fs.mkdir(STORE, { recursive: true });
     await fs.writeFile(DISABLED, '');
-    await fs.writeFile(BLOCKING, '');
+    await fs.writeFile(ADVISORY, '');
+    await fs.writeFile(LEGACY_BLOCKING, '');
     await gateCommand('on');
     expect(await exists(DISABLED)).toBe(false);
-    expect(await exists(BLOCKING)).toBe(false);
+    expect(await exists(ADVISORY)).toBe(false);
+    expect(await exists(LEGACY_BLOCKING)).toBe(false);
   });
 
-  it('block writes gate-blocking and clears gate-disabled', async () => {
+  it('block is an alias of on', async () => {
     await fs.mkdir(STORE, { recursive: true });
-    await fs.writeFile(DISABLED, '');
+    await fs.writeFile(ADVISORY, '');
     await gateCommand('block');
-    expect(await exists(BLOCKING)).toBe(true);
+    expect(await exists(ADVISORY)).toBe(false);
     expect(await exists(DISABLED)).toBe(false);
   });
 
-  it('status reports the current mode without mutating state', async () => {
+  it('off writes gate-disabled and clears advisory/legacy flags', async () => {
+    await fs.mkdir(STORE, { recursive: true });
+    await fs.writeFile(ADVISORY, '');
+    await fs.writeFile(LEGACY_BLOCKING, '');
+    await gateCommand('off');
+    expect(await exists(DISABLED)).toBe(true);
+    expect(await exists(ADVISORY)).toBe(false);
+    expect(await exists(LEGACY_BLOCKING)).toBe(false);
+  });
+
+  it('advisory writes gate-advisory and clears disabled/legacy flags', async () => {
+    await fs.mkdir(STORE, { recursive: true });
+    await fs.writeFile(DISABLED, '');
+    await fs.writeFile(LEGACY_BLOCKING, '');
+    await gateCommand('advisory');
+    expect(await exists(ADVISORY)).toBe(true);
+    expect(await exists(DISABLED)).toBe(false);
+    expect(await exists(LEGACY_BLOCKING)).toBe(false);
+  });
+
+  it('status reports off when gate-disabled exists', async () => {
     await fs.mkdir(STORE, { recursive: true });
     await fs.writeFile(DISABLED, '');
     await gateCommand('status');
     expect(logSpy).toHaveBeenCalledWith(expect.anything(), 'off');
     expect(await exists(DISABLED)).toBe(true);
+  });
+
+  it('status reports advisory when gate-advisory exists', async () => {
+    await fs.mkdir(STORE, { recursive: true });
+    await fs.writeFile(ADVISORY, '');
+    await gateCommand('status');
+    expect(logSpy).toHaveBeenCalledWith(expect.anything(), 'advisory (UI-only)');
+  });
+
+  it('status reports on (blocking) when no flags are set', async () => {
+    await fs.mkdir(STORE, { recursive: true });
+    await gateCommand('status');
+    expect(logSpy).toHaveBeenCalledWith(expect.anything(), 'on (blocking)');
   });
 
   it('rejects unknown actions with exit 1', async () => {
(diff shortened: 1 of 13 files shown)


alfhen:
Closing in favor of #567.

Live dogfooding inside Claude Code 2.1.142 surfaced a finding that changes the design: `hookSpecificOutput.additionalContext` from PostToolUse actually reaches the model on the next turn (verified with a behavioral probe). That makes a much simpler architecture viable — surface Lien impact analysis as a post-edit annotation rather than gating edits beforehand.

The annotator approach is the same outcome (model sees impact data) without the gate's structural baggage: no sentinels, no TTL, no path canonicalization, no multi-session GC, no advisory-vs-blocking dichotomy, no ceremony-vs-comprehension proxy problem.

`resolveProjectRoot` and `lien path` will be salvaged from this branch when #567 lands. Everything else (gate.sh, sentinel.sh, session-clean.sh, `lien gate` subcommand, `/lien-gate` slash command, gate-* tests) is retired.

PR #573, opened 2026-05-16
feat(plugin): daemon-backed lien annotate with in-memory suppression

### LINKED ISSUE #569: Daemon-backed lien annotate: in-memory suppression + amortized index connect
## Context

PR #568 / issue #567 shipped the post-Read impact annotator (`lien annotate <file>` invoked from a PostToolUse hook). Each annotation today costs ~500–1500ms — dominated by Node startup + LanceDB connect + `scanAll`. Mitigations in v1: trivial-skip suppression for low-impact files, plus a 5-min per-session TTL touchfile to avoid re-annotating the same file in close succession.

That works, but two related quality-of-experience issues remain:

1. **Cold-start cost on every annotation.** A 1-second pause on every fresh Read is at the edge of acceptable for an agentic loop where Reads are frequent. The 5-min suppression hides the cost for repeat Reads of the same file, but cross-file Reads still pay full cost.
2. **Disk-based suppression as a workaround for process boundaries.** The current design uses touchfiles under `~/.lien/indices/<repoId>/annotated-sessions/<sid>/` because bash hooks are short-lived processes with no shared memory. Filesystem semantics are correct and cheap, but the implementation has session-id sanitization, TTL math, and a SessionStart GC step that only exist because we don't have a persistent process.

Both go away with one architectural shift: a daemon that holds the index connection and suppression state in memory, with thin hooks RPC'ing into it.

## Proposal

Introduce a long-running `lien annotate-daemon` (or fold it into `lien serve` — see Open questions). The hook script becomes a thin client that sends `{ session_id, file_path }` over stdio (or a Unix socket) and reads the annotation back. The daemon:

- Holds one `VectorDB` connection open across all annotation requests (no per-call `lancedb.connect`).
- Keeps a `Map<session_id, Map<file_path_hash, timestamp>>` in memory for suppression.
- Background-refreshes the chunk scan when the index changes (currently the dependency-analyzer caches by `indexVersion` per-process — that cache becomes session-wide for free).
- Exits on idle (e.g., 30 min) so it doesn't leak.

Per-call cost drops from ~500–1500ms to ~5–20ms (essentially `findDependents` over a hot in-memory index plus formatter overhead).

## What goes away

- `plugins/claude/hooks/annotate-clean.sh` — daemon GCs its own state on idle/exit
- `<store>/annotated-sessions/*` on disk — superseded by in-memory map
- `LIEN_ANNOTATE_TTL_MIN` as a coarse 5-min default — the daemon can be smarter (e.g., suppress only within the same MCP turn, since the model has fresh context anyway)
- The `session_id` sanitization defensive code (we still validate it but no longer interpolate it into a path)

## What stays

- `plugins/claude/hooks/annotate-read.sh` — but shrinks to a 5-line stdio bridge
- `lien annotate <file>` CLI subcommand — kept as a one-shot for users who want to invoke from the terminal without the daemon (cold-start cost acceptable for ad-hoc use)
- Trivial-skip suppression (cheap files don't get annotated regardless)
- The annotation format itself

## v1 scope

- `lien annotate-daemon` subcommand: starts a long-running process listening on a Unix socket at `~/.lien/indices/<repoId>/annotate-daemon.sock`.
- Daemon protocol: line-delimited JSON, `{ session_id, file_path }` → `{ annotation?: string }`. One in-flight request at a time per connection.
- `annotate-read.sh` connects to the socket if it exists, else falls back to spawning `lien annotate` (graceful degradation while the daemon is still rolling out).
- Daemon auto-spawn: if the hook can't connect, it spawns the daemon in the background, retries once after a short delay. Subsequent hooks reuse the running daemon.
- Idle-shutdown: daemon exits after 30 min of no requests.
- In-memory suppression: `Map<session_id, Map<hash, timestamp>>` with the same 5-min TTL, expired entries cleaned lazily on next access for that session.

## Open questions

- **Fold into `lien serve`, or separate daemon?** `lien serve` is already long-running for users who installed Lien for the MCP integration. Folding in keeps the dependency surface flat. But `serve` has a different lifecycle (Claude Code spawns and supervises it). A separate `annotate-daemon` is independent and survives MCP server restarts. Probably separate — easier to reason about lifecycles independently.
- **Unix socket vs stdio?** Socket lets the daemon serve multiple sessions concurrently. stdio is simpler but binds the daemon to one client. Socket is the right answer; the complexity is small.
- **Windows?** Named pipes instead of Unix sockets. Same protocol, different transport.
- **Concurrent requests?** A daemon serving multiple sessions can get simultaneous requests. The annotator is mostly CPU/IO-bound on the index; LanceDB scan should be fine to share. Need to confirm the dependency-analyzer's `chunkScanCache` is reentrant — quick code read.

## Deferred to v3

- **Pre-warming on SessionStart.** Daemon could start eagerly when a session begins, paying the LanceDB connect once. Saves ~500ms on the very first Read.
- **Cross-session annotation cache.** A daemon could remember "we already analyzed `foo.ts` at commit `X` 2 minutes ago" and reuse the result for another session reading the same file. v3 polish.
- **Streaming annotations.** For files with dozens of dependents, stream the listing as the daemon computes it instead of buffering. Probably not worth it — annotations are short.

## References

- PR #568 — the v1 implementation this issue extends
- Issue #567 — the original annotator design
- `packages/cli/src/cli/annotate-cmd.ts` — the one-shot CLI to wrap
- `plugins/claude/hooks/annotate-read.sh` — the hook to thin out
- `packages/cli/src/mcp/handlers/dependency-analyzer.ts:55-63` — the existing per-process scan cache that becomes per-daemon (session-wide) for free

alfhen:
## Summary

- Per-call cost dropped from ~500–1500ms to ~5–20ms by holding the `VectorDB` + `scanCache` in a long-running per-repo `lien annotate-daemon`.
- Suppression state moved from disk touchfiles into an in-memory `Map<sessionId, Map<file, ts>>` with the same 5-min TTL — no more `<store>/annotated-sessions/` machinery, no more SessionStart/SessionEnd GC hooks, no more session_id path interpolation.
- Hook (`annotate-read.sh`) is now a thin socket bridge that auto-spawns the daemon on a missing socket and falls through to today's one-shot `lien annotate` if the daemon path is broken — zero user-visible regression.

Closes #569.

## What's in the diff

| Area | Change |
|------|--------|
| `packages/cli/src/daemon/` (new) | `annotate-daemon.ts` (Unix-socket server, `SuppressionTracker`, idle timer, version-poll → `reconnect` + `clearDependencyCache`, stale-socket unlink), `protocol.ts` (NDJSON), `client.ts` (promise-based test client). |
| `packages/cli/src/cli/annotate-daemon-cmd.ts` (new) | `lien annotate-daemon [--detach]` subcommand. `--detach` respawns into background and exits 0; bare invocation runs foreground with SIGINT/SIGTERM shutdown. |
| `packages/cli/src/cli/annotate-cmd.ts` | Core extracted as `runAnnotateOnce(vectorDB, paths)`. One-shot CLI keeps working for terminal use. |
| `packages/cli/src/mcp/handlers/dependency-analyzer.ts` | `getOrScanChunks` now single-flights cache rebuilds — concurrent daemon connections share one scan. |
| `plugins/claude/hooks/annotate-read.sh` | Socket bridge (nc → python3 → spawn-and-retry → fall-through to `lien annotate`). |
| `plugins/claude/hooks/hooks.json` | SessionStart + SessionEnd entries removed; daemon owns lifecycle. |
| `annotate-clean.sh`, `annotate-end.sh` | Deleted. |

## Design notes

- **One daemon per repo.** Socket at `~/.lien/indices/<repoId>/annotate-daemon.sock`, mirrors the existing per-repo store-path scheme. Lifecycles independent across repos.
- **Idle shutdown:** 30 minutes of inactivity → graceful exit.
- **Version invalidation:** polls `.lien-index-version` every `VERSION_CHECK_INTERVAL_MS` (same constant `lien serve` uses). On change → `vectorDB.reconnect()` + `clearDependencyCache()`.
- **Crash recovery:** stale socket files are `unlink`ed on daemon startup before bind.
- **Concurrency:** `findDependents` is read-only against the shared `importIndex`; the single-flight on the initial scan rebuild handles the only contention point.
- **Suppression policy unchanged at 5 min** for v1 — validating the latency win before tuning behavior.

## Test plan

- [x] Unit: `protocol.test.ts` (7 tests — encode/parse round-trip, malformed-input rejection)
- [x] Unit: `suppression-tracker.test.ts` (4 tests — TTL, expiry with injectable clock, per-session and per-file isolation)
- [x] Integration: `annotate-daemon.test.ts` (6 tests — empty-index null response, invalid-session error, concurrent connections, stale-socket unlink, graceful close, idle-shutdown)
- [x] Pre-commit gate: `format:check` ✓, `lint` ✓ (28 pre-existing warnings, 0 errors), `typecheck` ✓, `build` ✓
- [x] Full test suite passes (1065 across cli/parser/core/runner; the 40 qdrant failures are pre-existing and environmental — they require a running Qdrant server, see CLAUDE.md)
- [ ] Manual E2E (post-merge dogfooding):
  - `npm run build && npm link`
  - Fresh Claude session reads a non-trivial file → annotation lands via `additionalContext`
  - Re-read same file inside 5 min → silent (in-memory suppression hit)
  - Different file → annotation in <50ms (cache warm)
  - `pkill -f annotate-daemon` mid-session → next Read auto-respawns daemon
  - `lien index` mid-session → next annotation reflects new index (version-poll triggers reconnect)

---

### Lien Review

🔴 **Review required** - 1 new function is too complex.
🔗 **Impact**: 1 high-risk file(s) with 3 total dependents
| ⏱️ time to understand | 1 | +165 |

> [!WARNING]
> **2 issues found** · Medium Risk
>
> The PR introduces a daemon-backed annotation system to significantly reduce latency. While the architecture is sound and includes robust fallbacks, there are two logic errors that can lead to daemon crashes or failure to auto-restart after a crash.
>
> - Introduced `annotate-daemon` to hold VectorDB and suppression state in memory, reducing per-call overhead.
> - Updated `annotate-read.sh` hook to use the daemon with a one-shot fallback for reliability.
> - Implemented single-flighting for dependency scans in `getOrScanChunks` to handle concurrent requests efficiently.

<sup>Reviewed by [Lien Review](https://lien.dev) for commit 429c0ea. Updates automatically on new commits.</sup>


alfhen:
Closing in favor of a broader profiling pass. The daemon delivered the latency win but is overkill for a single hook path — pivoting to measure where Lien actually spends time across indexing, MCP tools, and annotate, then targeting the highest-leverage bottlenecks (suspected: pushing `findDependents` filtering into LanceDB rather than full `scanAll` in JS). Branch stays on remote for reference.

PR #606, opened 2026-06-28
feat(parser): add Swift AST support

alfhen:
## What

Upgrades **Swift** from line-based chunking to full Tree-sitter AST parsing — symbols, imports, call sites, complexity, and test associations — bringing it in line with the other AST languages (now **11**: TS, JS, PHP, Python, Rust, Go, Java, C#, Ruby, Kotlin, **Swift**). Follows the Ruby (0.46.0) / Kotlin (0.47.0) playbook.

This is **PR 1 of the arc** (AST support). PR 2 will add the e2e project + docs-site move + changeset (the changeset auto-triggers the full e2e suite and opens the release).

## Grammar

`tree-sitter-swift@^0.7.1` (alex-pinkus). Loads against the repo's `tree-sitter@0.25.0`; ships prebuilds for all 6 target platforms (incl. linux-x64/arm64 for CI) so there's **no native compile**.

Unlike Kotlin, the grammar exposes field names (`name`/`body`/`return_type`/`declaration_kind`), so `swift.ts` follows the `java.ts` field-based pattern. Swift-specific handling:
- `class` / `struct` / `actor` / `enum` / `extension` all share **one** `class_declaration` node, distinguished by the `declaration_kind` keyword → all map to `SymbolInfo.type: class` with the real keyword kept in the signature (`struct Foo`, `enum Color`, `extension AFError`).
- `protocol_declaration` → `interface` (its `protocol_function_declaration` members are abstract → not chunked).
- `init` / `deinit` / `subscript` are methods.
- Only the `Swift` stdlib module is filtered from imports; `Foundation`/`UIKit`/`Combine`/etc. are kept as real edges.
- `isTestFile` now recognises Swift conventions (`Tests/` dirs, `*Tests.swift`), **scoped to `.swift`** so other languages are unchanged (updated in both the parser and core copies).

## Validation

- ✅ `npm run typecheck` / `lint` / `format:check` / `build` (parser + core) clean
- ✅ Full parser suite **852 tests** green (28 new Swift cases + registry flips), no regressions
- ✅ **Dogfood on Alamofire**: 98 files, **0 parse errors**, 3183 chunks (~15.6ms/file). Symbol mix: 523 class / 1645 method (all with `parentClass`) / 116 function / 27 interface. Imports correctly capture Foundation/UIKit/Combine/XCTest and filter `Swift`; 44 test files detected.

## ⚠️ Distribution decision needed before merge

[...]

PR #608, opened 2026-06-29
fix(review): never report a no-verdict agent run as a clean review

alfhen:
**Follow-up #2** (independent of the Swift work; branched from `main`). Fixes the agent-review reliability bug surfaced while dogfooding #606.

## The bug

On #606 the agent-review **silently reported "0 findings"** (a clean PR) on one run, then a real 2-finding review on an identical re-run. Pulling the agent trace (via the now-default-on logging) showed why:

The agent investigates to its turn/budget ceiling, is force-nudged to emit, and ends `finish_reason: 'stop'`. Whether that forced final turn yields valid findings JSON vs. prose is **non-deterministic** (the Kimi provider intermittently ignores `response_format: json_object`). When it emitted prose:

```
incomplete = stopReason !== 'completed' && !parsed.summary
```

`stopReason` was `'completed'` (the model chose to stop), so `incomplete` computed **false** even though no verdict was parsed and the summary-retry also failed → the run was shown as a **clean 0-findings review**. A false "all clear" is the worst failure mode for a review tool.

## Fix (both OpenAI/OpenRouter and Anthropic clients)

1. **Verdict-based `incomplete`** — a run is complete **iff it produced a summary**, regardless of stop reason. A "completed but no parseable verdict" run is now flagged incomplete, so the existing *"Lien Review did not finish — re-run"* notice surfaces instead of a false clean.
2. **Lenient `extractResponse`** — after the ` ```json ` fence and the raw body, also recover a JSON object **embedded in surrounding prose** (model ignored `json_object` and prefixed reasoning). Turns a would-be silent-empty into a real verdict.
3. `index.ts` incomplete-notice reason now covers the `completed`-no-verdict case (*"ended without emitting a parseable JSON verdict"*).

## Tests

385 review tests pass, including two new regression tests:
- a `stop` turn with prose (+ retry also prose) → `incomplete: true` (would be `false` under the old formula — the silent-clean bug)
- a prose-wrapped JSON verdict → recovered on the same turn, `incomplete: false`

Not a rule/prompt change (doesn't alter what findings the agent produces — only whether a no-verdict run is flagged), so no harness calibration; no changeset (the review package is built from source, not published).

> Note: this reduces the *blast radius* of the non-determinism (no more false clean; degraded runs are now visible). The deeper tuning — keeping the agent from exhausting its budget mid-investigation — is separate; happy to follow up if you want to raise/retune the turn & token budgets.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of review runs that end without a usable summary, so incomplete results are now detected more reliably.
  * Expanded response parsing to recover valid JSON even when it appears in plain text or embedded in surrounding prose.
  * Updated the incomplete-review message to better reflect when a verdict could not be parsed.

* **Tests**
  * Added coverage for incomplete runs and for JSON verdicts embedded in non-JSON text.

---

### Lien Review

🔴 **Review required** - 2 new functions are too complex.
🔗 **Impact**: 2 high-risk file(s) with 4 total dependents
| 🔀 test paths | 2 | +0 |
| 🧠 mental load | 4 | +21 |
| ⏱️ time to understand | 7 | +199 |
| 🐛 estimated bugs | 2 | +0.115 |

> [!WARNING]
> ⚠️ **Review did not complete**
>
> Lien Review did not finish — it stopped unexpectedly while investigating. Any findings shown are partial; re-run the review to retry.

<sup>Reviewed by [Lien Review](https://lien.dev) for commit 6b18519. Updates automatically on new commits.</sup>

### DIFF
--- a/packages/review/src/plugins/agent/anthropic-client.ts
+++ b/packages/review/src/plugins/agent/anthropic-client.ts
@@ -535,28 +535,37 @@ function extractResponse(content: Anthropic.Messages.ContentBlock[]): {
   // so the verdict may not live in the last one.
   const text = textBlocks.map(b => b.text).join('\n');
 
-  // Try, in order: a ```json fence (normal turns); the raw body (a forced
-  // verdict turn may emit bare JSON); and a JSON object embedded in prose. The
-  // last guards against a model that prefixed reasoning — without it that turn
-  // parses to an empty result and the run is silently treated as clean.
-  const fenced = text.match(/```json\s*\n([\s\S]*?)\n\s*```/)?.[1];
-  const candidates = [fenced, text.trim(), embeddedJsonObject(text)];
+  // Candidate JSON strings, in priority order: each ```json fence LAST first (a
+  // few-shot/example fence precedes the real verdict), then the raw body, then a
+  // JSON object embedded in prose. Prefer a candidate carrying a `summary` (the
+  // verdict marker) so an example can't beat the real verdict; fall back to the
+  // first findings-only candidate if nothing carries a summary.
+  const fences = [...text.matchAll(/```json\s*\n([\s\S]*?)\n\s*```/g)].map(m => m[1]).reverse();
+  const candidates = [...fences, text.trim(), embeddedJsonObject(text)];
 
+  let fallback: { findings: AgentFinding[] } | undefined;
   for (const candidate of candidates) {
     if (!candidate) continue;
+    let parsed: unknown;
     try {
-      const parsed = JSON.parse(candidate);
-      const rawFindings = Array.isArray(parsed) ? parsed : parsed.findings;
-      const findings: unknown[] = Array.isArray(rawFindings) ? rawFindings : [];
-      const summary = isValidSummary(parsed.summary) ? parsed.summary : undefined;
-      if (summary || findings.length > 0) {
-        return { findings: findings.filter(isValidFinding), summary };
-      }
+      parsed = JSON.parse(candidate);
     } catch {
-      // not parseable — try the next candidate
+      continue; // not parseable — try the next candidate
     }
+    const { findings, summary } = readVerdict(parsed);
+    if (summary) return { findings, summary };
+    if (findings.length > 0 && !fallback) fallback = { findings };
   }
-  return { findings: [] };
+  return fallback ?? { findings: [] };
+}
+
+/** Pull validated findings + summary out of one parsed JSON verdict (array or object). */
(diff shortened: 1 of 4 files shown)


alfhen:
**Added a second, distinct failure mode** found while checking #607's review run.

#607's agent-review didn't silently report clean — it **threw** `Unexpected end of JSON input` and failed (non-blocking → 0 findings). Root cause: `chatCompletion` did `(await response.json())` unguarded, so a 200 with an empty/truncated body (an OpenRouter/Kimi hiccup) crashed the whole review. The original silent-clean fix in this PR does **not** cover this.

New commit (ddc8b99): `chatCompletion` reads the body as text, parses defensively, and **retries transient failures (empty/unparseable body, 429, 5xx) up to 3× with backoff** — the same hiccup recovered on a re-run, so an in-process retry recovers it. Genuine 4xx still throws immediately; the Anthropic client is unaffected (uses the SDK). Added a regression test (a `null` mock response = transient empty body → client retries and recovers).

So #608 now hardens the agent against the **three** ways a flaky model/provider turned into a bad review: silent false-clean (no verdict), prose-wrapped JSON, and an empty-body crash. 387 review tests pass.

alfhen:
**Two updates:**

**1. Third failure mode fixed — `fetch failed` (network error / hung connection).** Distinct from the silent-clean and empty-body cases: `fetch()` itself *rejected* (or a hung connection stalled the turn ~5 min before failing), which my earlier empty/5xx retry didn't cover. Now:
- `fetch` is wrapped with an `AbortController` **120s timeout** and retried on rejection (network error / abort), alongside the existing empty/429/5xx retries.
- If chat requests still fail after retries, the run loop ends gracefully (`stopReason: 'error'` → incomplete notice) instead of crashing the plugin.
- Tests: fetch-rejection retries & recovers; persistent failure degrades to an incomplete review.

**2. Rebased onto #606** (base changed to `worktree-swift-ast`). You were right that #608's runs showed no verbose agent trace — the default-on logging change lives on #606, and #608 was branched from `main` without it. Rebasing onto #606 means #608's own CI runs now emit the full per-turn trace (and it stacks cleanly like #607; GitHub will retarget to `main` once #606 merges). No conflict — the two PRs touch different regions.

**Also addressed CodeRabbit** on the extraction code: guard `findings` is an array before `.length`/`.filter`; concatenate all Anthropic text blocks (not just the last).

388 review tests pass. #608 now hardens the agent against four flaky-provider failure modes: silent false-clean, prose-wrapped JSON, empty body, and fetch rejection/timeout.

PR #629, opened 2026-07-02
fix(core): close alternation-based ReDoS bypass in safeRegex

alfhen:
## What

`packages/core/src/utils/safe-regex.ts`'s `REDOS_PATTERN` heuristic only
rejected a quantifier immediately before a closing paren (e.g. `(a+)+`).
It missed alternation-based ReDoS: `safeRegex('(a|a)+$')` returned a live
`RegExp`, and `.test()` on a 26-char string took ~3.5s, growing
exponentially per added character (confirmed via live repro, not just
static analysis).

`safeRegex` is reachable from external input via the `list_functions` MCP
tool's `pattern` arg (`packages/cli/src/mcp/handlers/list-functions.ts`)
and the vector DB pattern filters (`packages/core/src/vectordb/query.ts`,
`filterByPattern`/`matchesPattern`). `lien serve` is single-threaded, so
one bad pattern from a client is a full denial of service.

## Why this approach / library choice

Task started from `redos-detector` as the first candidate. I evaluated it
against a benign+malicious corpus before committing:

- **`redos-detector`**: correctly flags `(a|a)+$` (it's literally their
  README's canonical "bad" example) but at default budget also flags
  common, actually-safe patterns like `.*Controller.*` and
  `a.*b.*c.*d.*e.*` as unsafe (`hitMaxScore`) — these are polynomial, not
  exponential, but the detector's own path-exploration algorithm can't
  distinguish "big but finite" from "infinite" within its budget. Raising
  the budget to reduce false positives made the detector itself **OOM** on
  a 5-wildcard pattern (`a.*b.*c.*d.*e.*` with `maxScore: 100000` crashed
  the Node process). Both the default (too many false positives on
  everyday search patterns) and the raised-budget config (DoS in the
  analyzer itself) are disqualifying for a hot-path dependency.
- **`recheck`**: ships native per-platform binaries
  (`recheck-linux-x64`, `recheck-macos-arm64`, ...) via
  `optionalDependencies` — disqualified by the pure-JS/no-native-deps
  requirement.
- **`safe-regex2`** (Fastify org, `ret`-based star-height analysis):
  zero false positives on the full existing benign-pattern test suite,
  tiny (11.5kB + one zero-dependency sub-dependency `ret`), no install
  scripts, fast, linear-time analysis (no OOM risk). But it's a
  star-height/nested-quantifier checker, and **it doesn't catch `(a|a)+$`
  either** (returns "safe") — that's an inherent blind spot of the whole
  "safe-regex" checker lineage: they look at repetition *nesting*, not
  ambiguity *within* one repeated alternation.

Given neither pure-JS library alone closes the demonstrated bypass without
either missing it or regressing common usage, I shipped `safe-regex2` as
the primary (well-maintained, fast, zero false positives on existing
tests) plus a small, targeted, unit-tested structural check
(`hasDuplicateAlternationBranch`) that scans quantified groups for
alternation branches that are identical once case-folded (patterns always
compile with the `i` flag) — e.g. `(a|a)+`, `(a|A)+$`, `(?:a|a)+$`. I
verified empirically that the *exact-duplicate* shape is what causes
exponential blowup in V8 (`(a|a)+$` on 26 chars: ~3.5s; `(a|A)+$`: ~5.5s)
while a same-length *prefix-overlap* shape (`(a|aa)+$`) does **not**
(~22ms) — so scoping the check to exact duplicates avoids over-rejecting
patterns like `(get|getAll)+` that merely share a prefix.

## Changes

- `packages/core/src/utils/safe-regex.ts`: replaced the hand-rolled regex
  heuristic with `safe-regex2` + `hasDuplicateAlternationBranch`, plus a
  256-char pattern length cap enforced before any analysis runs (defense
  in depth against pathologically long input). Function signature and
  null-return contract unchanged — no call-site changes needed.
- `packages/core/package.json`: added `safe-regex2` (pure JS, one
  zero-dep sub-dependency `ret`, no install scripts).
- `packages/core/src/utils/safe-regex.test.ts`: added coverage for the
  `(a|a)+$` bypass and variants (case-folded, non-capturing group),
  confirms nested-quantifier patterns are still rejected, confirms benign
  alternations (`^get.*Service$`, `foo|bar`, `(foo|bar)+`,
  `(get|set)Value`) still compile, and confirms the length cap. All
  ReDoS-prone assertions check `toBeNull()` only — no dangerous pattern is
  ever executed against a real string in tests.
- `.changeset/safe-regex-redos.md`: patch `@liendev/core` +
  `@liendev/lien`.

## Hot-path latency

`safeRegex` runs once per `list_functions` call and once per candidate
record in `query.ts`'s content-scan fallback (`filterByPattern` /
`matchesPattern`, pre-existing per-record call pattern, unrelated to this
change). Benchmarked on typical short patterns:

- 7 realistic patterns × 1000 calls each (7000 total): **10.16ms total,
  ~0.0015ms/call**.
- Same pattern called 10,000 times (worst-case repeated-recompute in a
  large scan): **6.80ms total, ~0.0007ms/call**.

Trivial relative to any DB scan or MCP round-trip.

## Verification

```
npm run format:check   # pass
npm run lint            # pass, 0 errors (22 pre-existing warnings, unrelated)
npm run typecheck       # pass
npm run build           # pass
npm run test -w @liendev/core -- --exclude '**/worker-embeddings.test.ts'
  # 32 files / 494 tests passed
npm run test -w @liendev/lien
  # 40 files / 677 tests passed, incl. list-functions.test.ts (23 tests)
```

`worker-embeddings.test.ts` excluded per the known environmental
fork-worker teardown crash in symlinked worktrees (reproduces on
unmodified main) — unrelated to this change.

## Caveats

- Dependency install used the worktree two-step (package-lock-only in the
  worktree + `npm install --no-save safe-regex2 -w @liendev/core` in the
  main checkout). On conflict, rebase and re-run
  `npm install --package-lock-only`.
- `packages/review/src/plugins/agent/rules.ts` has its own, separately
  defined `safeRegex` with the same original heuristic (same blind spot),
  but it only compiles rule-authored keyword strings baked into
  `BUILTIN_RULES` — not attacker-reachable — so it's out of scope here.
- The duplicate-alternation check is intentionally narrow (exact-duplicate
  branches only, not general prefix/overlap ambiguity) to avoid
  over-rejecting legitimate patterns; broader alternation ambiguity is a
  known residual gap, consistent with using this as defense-in-depth
  rather than a formal verifier.

---

### Lien Review

> [!NOTE]
> **Low Risk**
>
> This PR replaces the hand-rolled ReDoS heuristic in `safeRegex` with `safe-regex2` plus an AST-based duplicate-alternation check and a 256-character length cap. The implementation correctly handles the reported bypass `(a|a)+$` and its variants, including nested wrapper groups like `((a|a))+` and case-folded duplicates. Callers in `list-functions.ts` and `query.ts` already treat a `null` return as a graceful fallback, so the changed rejection behavior is non-breaking. The test suite covers the new bypass cases, nested-quantifier rejection, benign alternations, character-class/escaped-pipe false-positive avoidance, and the length boundary.
>
> - Replaced `REDOS_PATTERN` heuristic with `safe-regex2` + `hasDuplicateAlternationBranch`
> - Added 256-character pattern length cap
> - Added regression tests for alternation-based ReDoS bypasses and false-positive cases

<sup>Reviewed by [Lien Review](https://lien.dev) for commit 33ed497. Updates automatically on new commits.</sup>


alfhen:
Addressed the two agent-review findings in 33ed497.

| Finding | Verdict | Notes |
|---|---|---|
| [`:40` nested duplicate alternation bypasses the check](https://github.com/getlien/lien/pull/629#discussion_r3512785916) (e.g. `((a|a))+`) | **Real — fixed** | Confirmed by direct repro: `safeRegex('((a|a))+')` returned a live `RegExp` pre-fix. `QUANTIFIED_GROUP` only matched non-nested `(...)+` groups, so an extra wrapping group defeated it entirely, even though `((a|a))+` is exploitable exactly like `(a|a)+`. |
| [`:56` branch splitter treats every raw `\|` as a boundary](https://github.com/getlien/lien/pull/629#discussion_r3512785923) (char classes / escapes) | **Real — fixed** | The bug mechanism is real, though I couldn't reproduce the review's own two examples as stated (`([a|b]\|[a|b])+` was correctly rejected even pre-fix, since those two branches genuinely are identical; `(a\\|a)+` was already allowed pre-fix, not rejected). I found and verified independent, genuine false positives caused by the same mechanism: `([a|c]\|[b|c])+` and `(a\\|x\|b\\|x)+` — both have two *distinct* branches but were incorrectly rejected because naive `.split('\|')` fragments the character class / escaped pipe. |

**Fix**: rewrote `hasDuplicateAlternationBranch` on top of `ret`'s AST (already a transitive dependency of `safe-regex2`, per the review's own suggestion) instead of regex/string-splitting. This walks nested groups/repetitions properly (closes the bypass) and compares real alternation branches instead of `|`-delimited substrings (closes the false positives), while still correctly rejecting genuinely duplicate branches that happen to contain `|` (e.g. `([a|b]\|[a|b])+`, `(a\\|a\|a\\|a)+`).

Added `ret` as an explicit direct dependency of `@liendev/core` (it was already resolving transitively, but we now import it directly). Extended `safe-regex.test.ts` with regression coverage for the bypass, the two verified false positives, and the "still correctly rejected" duplicate-despite-pipe cases — assertions only, no pattern is ever executed against a string.

All gates pass: format, lint, typecheck, build, `@liendev/core` tests (497 passed, incl. safe-regex.test.ts), `@liendev/lien` tests (677 passed).

PR #639, opened 2026-07-02
fix(action): trigger publish-action.yml on the tag changesets actually creates

alfhen:
## What

Fixes the self-hostable Lien Review Action's publish path so it can actually
ship once a human does the remaining one-time setup. Scope: code + docs only
(`.github/workflows/publish-action.yml`, `packages/action/README.md`) — no
external resources created, per task scope.

## Why (review finding)

`publish-action.yml` has **never run**: its trigger was `on: push: tags: v*`,
but this repo releases via `changesets/action` (`.github/workflows/release.yml`),
which tags pushes as `<pkg>@<version>` (e.g. `@liendev/lien@0.51.0`), never a
bare `v*` tag. Confirmed against `git tag -l`: every `v*` tag predates the
changesets migration (last one is `v0.19.5`); all tags since are
`@liendev/{lien,parser,core}@x.y.z`. Compounding this, the GHCR image was
never published and the `getlien/lien-review` dist repo doesn't exist, so
`uses: getlien/lien-review@v1` (as documented in the README and dogfooded via
`action.yml`'s `docker://ghcr.io/getlien/lien-review:v1`) is unusable by
anyone outside this repo.

## Approach

Went with **option (a)**: trigger on the tag changesets actually creates,
rather than (b) a manual promotion step or (c) folding publish into
`release.yml` (out of scope here — task scope is `publish-action.yml` +
`action.yml` + docs only, and another PR, #633, is touching CI workflows in
parallel).

- **Trigger**: `push: tags: '@liendev/lien@*'` instead of `v*`.
  `@liendev/parser`, `@liendev/core`, and `@liendev/lien` are version-linked
  in `.changeset/config.json`, so every monorepo release bumps `@liendev/lien`
  and creates that tag exactly once — this is the tag that reliably fires,
  today, with zero additional infra.
  - Why not a dedicated `@liendev/action` tag? `packages/action` and
    `packages/review` are `private: true` (unpublished), and changesets'
    documented default (`privatePackages.tag: false`) versions private
    packages but never tags them — see
    https://github.com/changesets/changesets/blob/main/docs/config-file-options.md.
    Wiring a dedicated tag means changing `.changeset/config.json`, which
    affects tagging behavior for every private package repo-wide — a bigger,
    shared-infra change I kept out of this PR's scope. Documented as a known
    gap + workaround (see runbook) instead.
- **Version/tag coherence**: the tag is `@liendev/lien@0.51.0` — Docker tags
  can't contain `@`/`/`, so the workflow extracts the bare semver
  (`${GITHUB_REF##*@}` → `0.51.0`) and publishes
  `ghcr.io/getlien/lien-review:{0.51.0,v1,latest,sha-<commit>}`. `:v1` is a
  fixed literal (the Action's own `inputs:`/`outputs:` interface major
  version, bumped manually only on a breaking `action.yml` change — same
  convention as `actions/checkout@v4`), not derived from the CLI's semver.
  This matches what `action.yml` (`docker://ghcr.io/getlien/lien-review:v1`)
  and every `uses: getlien/lien-review@v1` reference in the README/examples/
  site docs already expect — verified via grep, no drift found, no changes
  needed there.
- **Dist-repo sync**: uncommented the previously-stubbed sync step (copies
  `action.yml`/`README.md` to `getlien/lien-review` and moves the floating
  `v1` tag), but guarded it to check `GH_DIST_TOKEN` at the top of the script
  and `exit 0` with a `::notice::` log if it's unset, rather than failing the
  job. The GHCR image publish above still succeeds either way, so this
  workflow is mergeable today.
- **Runbook**: added a "Publishing the Action" section to
  `packages/action/README.md` for maintainers — see "Human steps required"
  below, same content.

## Human steps required before this ships

1. **Create the `getlien/lien-review` repo** (public, empty) in the `getlien`
   GitHub org.
2. **Add a `GH_DIST_TOKEN` secret** to this repo (Settings → Secrets and
   variables → Actions) — a PAT/fine-grained token with `contents:write` on
   `getlien/lien-review`.
3. **First trigger**: once both exist, either let the next changeset release
   land normally (creates `@liendev/lien@x.y.z`, fires this workflow), or run
   it manually via **Actions → Publish Action → Run workflow**
   (`workflow_dispatch` — note this path only pushes the immutable
   `sha-<commit>` image tag, not `:v1`/`:latest`/a version tag, so follow it
   with a real tagged release to move those).

None of these three are things I can do from this PR (no org/repo-creation or
secret-management access from CI).

## Verification

- `node -e "yaml.load(...)"` — confirms `publish-action.yml` parses and the
  new `on.push.tags` is `['@liendev/lien@*']`.
- Traced the trigger against real tag history: `git tag -l "*@*"` shows
  `@liendev/lien@0.45.0` … `@liendev/lien@0.50.0` created on every past
  release cycle (most recently by `ebc91f9 chore: version packages`) — the
  new trigger would have fired on every one of those pushes.
- Bash-traced the tag-extraction logic locally:
  `GITHUB_REF="refs/tags/@liendev/lien@0.51.0"` → `VERSION="0.51.0"` (valid
  Docker tag charset, checked against `^[a-zA-Z0-9_][a-zA-Z0-9._-]{0,127}$`).
- `grep -rn "getlien/lien-review"` across the repo — action.yml, README.md,
  examples/lien-review.yml, root README, site docs, and the workflow itself
  all already agree on `:v1`/`@v1`; no inconsistency found, no changes needed
  beyond the workflow + runbook.
- `npm run format:check && npm run lint && npm run typecheck && npm run build`
  — all green (0 lint errors; pre-existing warnings only, unrelated to this
  diff). No test suite touches workflow YAML or this README, so no
  package-level tests were affected.

## Caveats / out of scope

- An action-only change that doesn't also bump `@liendev/lien`'s version
  won't auto-trigger a publish (documented in the runbook, with
  `workflow_dispatch` as the escape hatch). A dedicated `@liendev/action` tag
  would close this gap but requires a `.changeset/config.json` change
  (`privatePackages.tag`) affecting all private packages — left for a
  follow-up.
- Did not touch `ci.yml`/`e2e.yml` (per scope — #633 is adding CI jobs there
  in parallel).
- No changeset included — neither file touched is a published package
  manifest (`@liendev/action`, `@liendev/review` are both `"private": true`).

---

### Lien Review

➡️ **Stable** - 79 pre-existing issues in touched files (none introduced).
🔗 **Impact**: 5 high-risk file(s) with 125 total dependents
| 🔀 test paths | 9 | — |
| 🧠 mental load | 30 | — |
| ⏱️ time to understand | 38 | — |
| 🐛 estimated bugs | 2 | — |

> [!NOTE]
> **Low Risk**

<sup>Reviewed by [Lien Review](https://lien.dev) for commit bf5c274. Updates automatically on new commits.</sup>


alfhen:
coderabbit review

PR #656, opened 2026-07-03
feat(core,cli): add opt-in SQLite structural backend behind the factory seam

alfhen:
## What

Adds `SqliteBackend` (better-sqlite3 + FTS5) implementing the existing `VectorDBInterface`, selectable via global config `backend: sqlite` or `LIEN_BACKEND=sqlite`. **Default stays `lancedb` — zero behavior change for existing users.**

PR 1 of 3 in the structural-first migration (next: flip the default + lexical search replaces semantic search; then: remove LanceDB + embeddings). Design validated in the structural-store benchmark spike (#645): get_files_context ~1000x faster, install footprint 93MB → 1.8MB.

## How

- **Structural store**: single `chunks` table + `file` index; `imports`/`exports`/`importedSymbols`/`callSites` stored as real JSON columns (no Arrow placeholder sentinels).
- **FTS5**: external-content index (porter tokenizer) over `symbolName`, camelCase-split `symbolTokens`, and `content`, kept in sync by AFTER INSERT/UPDATE/DELETE triggers — incremental indexing and the watcher can never drift the index (the spike's highest-risk trap, closed by construction).
- **search()**: ignores the vector argument and runs BM25 keyword search on the query text; a ratio-banded relevance mapping keeps the `SearchResult` score/relevance contract so handlers behave identically. `'parse import'` finds `parseImportStatement` via the split-identifier column — the spike's main tokenizer gap, closed.
- **filters.ts**: the pure JS filters (language/pattern/symbolType/symbol matching) extracted from query.ts and shared by both backends — no behavior change, existing query tests pass unchanged.
- **Parity test**: inserts the same fixture into a real LanceDB and SqliteBackend and asserts identical result counts, (file,startLine,endLine,symbolName) tuples, and deep-equal metadata — the spike's consistency methodology as a regression gate.
- Index path goes through `getLienHome()` (#653), so tests are sandboxed and leak nothing.

Deliberately omitted (YAGNI, documented in schema.ts): `chunk_imports` child table, trigram symbol index, extra indices — none has a consumer yet.

## Testing

- 38 new tests (backend contract 25, FTS 9, LanceDB parity 4); core suite 564 passed, full monorepo suite green.
- Driven end-to-end: indexed a fixture project with `LIEN_BACKEND=sqlite` via the built CLI; FTS search, scanWithFilter, and querySymbols all return correct results (`'password login'` → `verifyCredentials` via docstring).

## Summary by CodeRabbit

* **New Features**
  * Added an optional SQLite-backed storage/search mode alongside the existing default backend.
  * Users can now select it through configuration or an environment variable.
  * Search and symbol queries now work with SQLite FTS-based results.

* **Bug Fixes**
  * Improved validation for backend selection values.
  * Strengthened record filtering and result handling for blank, missing, or invalid fields.
  * Kept search and filtering behavior aligned across both backends.

---

### Lien Review

➡️ **Stable** - 2 pre-existing issues in touched files (none introduced).
| ⏱️ time to understand | 2 | +0 |

> [!WARNING]
> ⚠️ **Review did not complete**
>
> Lien Review did not finish — it stopped unexpectedly while investigating. Any findings shown are partial; re-run the review to retry.

<sup>Reviewed by [Lien Review](https://lien.dev) for commit a81c2e0. Updates automatically on new commits.</sup>

### DIFF
--- a/.changeset/sqlite-structural-backend.md
+++ b/.changeset/sqlite-structural-backend.md
@@ -1,0 +1,6 @@
+---
+'@liendev/core': minor
+'@liendev/lien': minor
+---
+
+Add an opt-in SQLite structural backend behind the existing vector-DB factory seam. Set `backend: sqlite` in the global config (`~/.lien/config.json`) or `LIEN_BACKEND=sqlite` to store chunks in a better-sqlite3 database with an FTS5 lexical index instead of LanceDB; the SqliteBackend implements the same `VectorDBInterface`, so no handler or indexer changes are needed. The default backend is unchanged (`lancedb`), so this release is purely additive.
--- a/package-lock.json
+++ b/package-lock.json
@@ -3233,6 +3233,16 @@
         "tslib": "^2.8.0"
       }
     },
+    "node_modules/@types/better-sqlite3": {
+      "version": "7.6.13",
+      "resolved": "https://registry.npmjs.org/@types/better-sqlite3/-/better-sqlite3-7.6.13.tgz",
+      "integrity": "sha512-NMv9ASNARoKksWtsq/SHakpYAYnhBrQgGD8zkLYk/jaK8jUGn08CfEdTRgYhMypUQAfzSP8W6gNLe0q19/t4VA==",
+      "dev": true,
+      "license": "MIT",
+      "dependencies": {
+        "@types/node": "*"
+      }
+    },
     "node_modules/@types/chai": {
       "version": "5.2.3",
       "resolved": "https://registry.npmjs.org/@types/chai/-/chai-5.2.3.tgz",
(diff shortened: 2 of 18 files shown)


alfhen:
Review findings addressed in a81c2e0:

- **fts-search.ts bm25 ratio** (Lien Review): fixed — ratio clamped to [0,1]; degenerate zero/positive ranks can no longer invert the relevance mapping.
- **row-mapping.ts importedSymbols validation** (Lien Review): fixed — values are validated as string arrays; malformed rows degrade to `undefined`, consistent with the sibling parsers (`parseJsonArray`/`parseCallSites`). Regression test added that corrupts a row on disk.
- **symbolType '' → undefined** (CodeRabbit): behavior kept intentionally — the LanceDB read path (query.ts `buildSearchResultMetadata`) does a bare cast and passes `''` through, so adding `|| undefined` here would *create* the parity divergence the finding warns about. The overpromising docstring is fixed instead.
- **test teardown file handles** (CodeRabbit): fixed — added `SqliteBackend.close()` and all suites close before removing their store dirs.
- **hasData broad catch** (CodeRabbit): kept as-is — return-false-on-throw is the exact LanceDB `hasData` contract this backend replicates; changing it is PR-C scope when the interface is reshaped.

PR #657, opened 2026-07-03
feat(core,cli)!: default to the SQLite backend and lexical FTS5 search

alfhen:
## What

PR 2 of 3 in the structural-first migration (after #656; next PR deletes LanceDB + embeddings code).

- **SQLite is now the default — and only selectable — backend.** `backend: lancedb` (config or `LIEN_BACKEND`) is retired exactly like Qdrant was in #624: stripped on load, warned once, falls back to sqlite. LanceDB code stays in-tree but unreachable; deletion is the next PR.
- **`semantic_search`, `find_similar`, and `get_files_context` related-chunks now run FTS5 BM25 keyword search** on the query text. Embeddings are never computed — indexing always uses `NullEmbeddings`, the worker thread never spawns, no model download ever happens.
- **One-time reindex is automatic**: on first `lien serve` after upgrading, `hasData()` is false for the sqlite store, so the existing auto-index path rebuilds — seconds, not minutes, since there's no embedding pass.
- **Every agent-facing guidance surface rewritten** for keyword/identifier queries (server instructions, tool descriptions, schema examples, `lien init`'s Explore-agent template, the plugin's explore hook — which also probed the now-dead `code_chunks.lance` dir to decide if a repo was indexed — README, CLAUDE.md, dogfood skill).
- **Three CLI commands were silently reading the wrong backend** (`complexity`, `annotate`, `index --force` constructed `new VectorDB` directly); they now go through `createVectorDB`. Caught by driving the built CLI against a fixture — `lien complexity` returned "Index not found" on a fresh index before the fix.

**Open question for @alf**: `semantic_search` keeps its (now slightly misleading) name to avoid breaking every existing user's CLAUDE.md. Happy to rename to `search_code` in a follow-up if you prefer honesty over compat.

## Testing

- Full monorepo suite green (parser 934 / core 561 / review 488 / cli 717 / action 28); zero leaked `~/.lien/indices` dirs.
- New `lexical-search-roundtrip` integration test: real MCP round trip over an embeddings-free index, asserting populated FTS5 results.
- `real-projects` e2e updated to read through the factory: Zod and Chi runs pass locally 14/14 (previously 77 assertion failures when reading sqlite indexes through LanceDB).
- Manually drove the built CLI on a fixture: `index`, `status` (now reports `Backend: sqlite`, lexical search), `complexity --format json`, `annotate` — all correct.

---

### Lien Review

✅ **Improved!** Complexity reduced by 72. 3 pre-existing issues remain in touched files.
| ⏱️ time to understand | 3 | -72 |

> [!WARNING]
> **3 issues found** · Medium Risk
>
> This PR retires LanceDB and makes SQLite/FTS5 the sole reachable backend, routing CLI commands through createVectorDB and removing embedding-dependent code paths. The migration is mostly sound, but status.ts silently masks config errors and the LienConfig schema still advertises LanceDB as the only/default backend, creating a mismatch between types/comments and runtime behavior.
>
> - SQLite is now the only selectable backend; lancedb config/env values warn once and fall back to sqlite
> - semantic_search, find_similar, and get_files_context related-chunks now use lexical FTS5 search with a vestigial empty query vector
> - Three CLI commands (complexity, annotate, index --force) now use createVectorDB instead of constructing VectorDB directly

<sup>Reviewed by [Lien Review](https://lien.dev) for commit 61b8cad. Updates automatically on new commits.</sup>

## Summary by CodeRabbit

* **Breaking Changes**
  * Search now uses SQLite-backed lexical full-text search only; prior LanceDB configurations fall back to SQLite with a warning.
  * Embedding-based behavior is gone, so queries work best with exact keywords, identifiers, and code terms.

* **New Features**
  * Status output now shows the active search backend and lexical search mode.
  * Related file context and similarity lookups now return results using the same keyword-based matching.

* **Bug Fixes**
  * Retired backend settings no longer break config loading or CLI commands.


alf:
I don’t have an opinion, please use the full name of @alfhen in the future.

PR #691, opened 2026-07-07
feat(delta): --base ref comparison + CI gate for new complexity crossings

alfhen:
## What

Adds `lien delta --base <ref>` and wires a CI job that runs it on every pull request, so the sixth commit gate (`lien delta`) finally has a CI backstop — the other five gates (format, lint, typecheck, build, test) already run in CI.

## Why

`lien delta` only ever compared the working tree to `HEAD`. That's the right default for an agent running the gate mid-edit, but it has a blind spot in CI: by the time a PR is checked out, its own commits are already `HEAD` and the working tree is clean, so a crossing introduced by an *earlier* commit in the same PR produces no diff at all against `HEAD` — the gate sees nothing to flag, even though the crossing is sitting right there in the PR's history. Until this PR, `lien delta` ran purely on an agent's honor system with no automated enforcement.

## How

- `--base <ref>`: compares the current state against any ref instead of `HEAD`. File discovery becomes `git diff --name-status --find-renames <ref>` (was `... HEAD`), and "before" content is read via `git show <ref>:path` (was `git show HEAD:path`). Untracked files are still additions regardless of baseline. Omitting `--base` is byte-for-byte identical to the original HEAD-vs-working-tree behavior.
- Composes with every existing flag: `--file` (single-file fast path vs `<ref>`), `--format json`, `--soft`, `--threshold`.
- A `--base` ref that doesn't resolve to a commit is a clear operational error, exit 2 (same convention as an unreadable file or missing git binary).
- CI: new `delta` job on `pull_request` events only, `fetch-depth: 0` checkout (needed so `origin/<base-branch>` resolves locally), builds parser+core+cli, then runs `node packages/cli/dist/index.js delta --base "origin/$GITHUB_BASE_REF"`. Fails the job on exit 1 (a new crossing).
- Docs: added a `--base <ref>` usage section plus a CI-wiring section to `docs/architecture/lien-delta.md`.
- Changeset: minor bump for `@liendev/lien` (parser was untouched — the delta-computation primitive already took ref-agnostic before/after content strings, so no parser changes were needed).

Note: `CLAUDE.md`'s "Before EVERY Commit" section still describes `lien delta` as having no CI backstop — that wording is owned by a parallel in-flight PR (worktree-agent-setup-overhaul) touching the same section, so it's left alone here to avoid a merge conflict. It becomes stale once both land.

## Verified

Local repro (also covered by the new integration tests in `delta-cmd.test.ts`):

```
$ node packages/cli/dist/index.js delta            # vs HEAD: working tree == HEAD, nothing to see
lien delta — no complexity-affecting changes vs HEAD (22 ms)
exit: 0

$ node packages/cli/dist/index.js delta --base <base-sha>   # same repo, vs the PR's base
lien delta — complexity vs c2097333ba51...

  a.ts
    ✗ crossed   target    cognitive 1 → 15 (limit 15)

  ✗ 1 new crossing · 1 file · 40 ms
exit: 1
```

This is exactly the CI-relevant gap `--base` closes: a crossing already committed into a PR's history is invisible to plain `lien delta` but caught against the PR's base branch.

All six gates green locally: `format:check`, `lint` (0 errors, pre-existing warnings only), `typecheck`, `build`, `test` (all packages, 1740+ tests), `lien delta` (exit 0, no new crossings introduced by this change itself).

---

### Lien Review

✅ **Improved!** Complexity reduced by 28. 2 pre-existing issues remain in touched files.
🔗 **Impact**: 1 high-risk file(s) with 2 total dependents
| ⏱️ time to understand | 2 | -13 |

> [!NOTE]
> **Low Risk**
>
> This PR adds `lien delta --base <ref>` and a CI backstop job with careful validation and preserved default behavior. Empty/whitespace `--base` is rejected as a usage error, non-existent refs produce a clear exit-2 message, and the HEAD-vs-working-tree default path is unchanged. The git plumbing correctly reads before-content from the supplied ref while keeping the working tree as the after side, untracked files remain additions, and the single-file fast path composes with `--base`. New integration tests cover the CI-relevant case where a crossing is invisible against HEAD but caught against the PR base.
>
> - Added optional `baseRef` parameter to `collectFileChanges` and `collectFileChange` with explicit ref validation
> - Added `--base <ref>` CLI option and empty-ref validation in `usageFlagError`
> - Updated text report header to name the actual comparison ref
> - Added CI `delta` job running `lien delta --base origin/$GITHUB_BASE_REF` on pull requests with `fetch-depth: 0`

<sup>Reviewed by [Lien Review](https://lien.dev) for commit af5d0b8. Updates automatically on new commits.</sup>

### DIFF
--- a/.changeset/delta-base-ci.md
+++ b/.changeset/delta-base-ci.md
@@ -1,0 +1,9 @@
+---
+'@liendev/lien': minor
+---
+
+`lien delta --base <ref>` — compare against any ref, not just `HEAD`, and give the sixth commit gate a CI backstop.
+
+Until now `lien delta` only ever compared the working tree to `HEAD`, so a crossing introduced by an earlier commit in a PR (already sitting at `HEAD`, with a clean working tree) was invisible to the gate — the other five commit gates are enforced in CI, but this one ran purely on an agent's honor system. `--base <ref>` compares the current state against any ref instead: `git diff --name-status` scoped to that ref, `before` content read via `git show <ref>:path`, same file filtering and edge-case handling (added/deleted/renamed/unborn) as the default mode. Composes with `--file`, `--format json`, `--soft`, and `--threshold`; omitting `--base` is byte-for-byte unchanged.
+
+CI now runs `lien delta --base "origin/$GITHUB_BASE_REF"` on every pull request, so a complexity crossing introduced anywhere in a PR's commits — not just its latest one — fails the build instead of merging silently.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -71,6 +71,44 @@ jobs:
       - name: Build Action image (Dockerfile smoke test)
         run: docker build -f packages/action/Dockerfile -t lien-review:ci .
 
+  delta:
+    name: Complexity delta vs base branch (lien delta --base)
+    runs-on: ubuntu-latest
+    if: github.event_name == 'pull_request'
+    timeout-minutes: 10
+
+    # `lien delta` is the sixth pre-commit gate (see CLAUDE.md), but until now
+    # it only ran on an agent's honor system -- nothing stopped a PR whose
+    # commits raised a function's complexity over threshold from merging, as
+    # long as no individual commit happened to run the gate locally. This job
+    # gives the gate a CI backstop: it compares the checked-out PR state
+    # against the PR's target branch (not just the PR's own last commit, which
+    # a crossing introduced earlier in the same PR could hide from) and fails
+    # the build on any NEW threshold crossing.
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with:
+          # Need the target branch's history locally (as `origin/<base>`) to
+          # diff against -- a shallow, single-ref checkout would not have it.
+          fetch-depth: 0
+          persist-credentials: false
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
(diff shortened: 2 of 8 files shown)


alfhen:
Adversarial review of the implementation (as follow-up to the fleet build):

**Confirmed sound:**
- Default-mode preservation is proven, not asserted — the `--base <HEAD sha>` equivalence test pins byte-identical JSON output, and `headBaseline()` reproduces the original code path exactly (including the unborn-HEAD fallback, correctly unreachable under an explicit `--base`).
- The CI-case test (`catches a crossing introduced relative to base but invisible vs HEAD`) directly demonstrates the gap this closes.
- Two-dot `git diff <base>` semantics are correct in the CI job specifically because `actions/checkout` on `pull_request` checks out the PR **merge commit** — the working tree already reflects merge-with-base, so diffing against `origin/<base>` yields exactly the PR's effective change.
- The job self-hosted on this very PR (the `delta` check ran and passed against its own diff).

**One refinement worth considering (non-blocking):** `origin/$GITHUB_BASE_REF` is re-fetched at job start, so if the base branch advances between the merge-commit computation and the fetch, the diff can attribute base-side changes to the PR (or mask them). Pinning to `github.event.pull_request.base.sha` instead of the branch name removes that race:
```yaml
run: node packages/cli/dist/index.js delta --base "${{ github.event.pull_request.base.sha }}"
```
Rare in practice; fine to land as-is and refine later.

PR #735, opened 2026-07-11
fix(review): fold doc-pass state into the rendered summary; severity-aware dedupe

alfhen:
Triage follow-up on #733 — Lien Review's own findings on the final push landed after the merge (my miss: I triaged inline comments but not the review body). All three confirmed real and fixed:

1. **Dead-code incomplete notice**: `present()`/`formatCheckSummary()` consume only the FIRST summary finding, so the separately-appended doc-pass notice never rendered — an unfinished doc pass still read as "no doc issues". Doc-pass state now folds into the main result before the notices (`mergeDocPassIntoResult`), with `incompleteFromDocPass` making the notice say the doc pass specifically ("code findings are unaffected") instead of implying the whole review is partial.
2. **Misleading risk headline**: doc-truth `error` findings now lift a low/absent risk level to medium and append the contradiction count to the overview.
3. **Severity-blind dedupe**: a doc-pass ERROR overlapping a main-pass WARNING within ±2 lines was silently dropped; dedupe now only drops the doc finding when the main finding is at least as severe.

7 new unit tests. No prompt changes — code-level rendering/merge logic only, zero-LLM-testable, no calibration required.

## Harness evidence
No rule prompt or signal text changed; the calibrate-gated surface is untouched (see #733's certified table — this PR only fixes how already-produced results render/merge).

---

### Lien Review

✅ **Improved!** Complexity reduced by 1. 3 pre-existing issues remain in touched files.
| ⏱️ time to understand | 3 | -1 |

> [!WARNING]
> **1 issue found** · Low Risk
>
> The PR folds doc-pass incomplete state and doc-truth errors into the main result before rendering, fixing the previously unrendered doc-pass notice and misleading low-risk headline. It also makes dedupe severity-aware. The only real issue is that the severity ranking is binary, which can still drop a doc warning when a main info finding overlaps it.
>
> - mergeDocPassIntoResult folds doc-pass incomplete state and doc-truth errors into the main result before summary/incomplete notices are appended
> - mergeDocTruthFindings now uses severity-aware dedupe so a doc error is not dropped by a nearby main warning
> - severityRank is incomplete: it treats warning and info as equal, allowing a main info finding to suppress a doc warning

<sup>Reviewed by [Lien Review](https://lien.dev) for commit d8794ea. Updates automatically on new commits.</sup>

### DIFF
--- a/packages/review/src/plugins/agent/doc-truth-pass.ts
+++ b/packages/review/src/plugins/agent/doc-truth-pass.ts
@@ -226,21 +226,68 @@ function sameLocation(a: AgentFinding, b: AgentFinding): boolean {
   return a.filepath === b.filepath && Math.abs(a.line - b.line) <= DEDUPE_LINE_PROXIMITY;
 }
 
+/** error outranks warning — used so dedupe never drops the sharper finding. */
+function severityRank(f: AgentFinding): number {
+  return f.severity === 'error' ? 1 : 0;
+}
+
 /**
  * Fold the doc pass's findings into the main pass's. Every doc-pass finding gets
  * `ruleId: 'doc-truth'` (the pass has a single rule, so attribution is
  * unambiguous regardless of what the model emitted). A doc-pass finding is
- * dropped when the main pass already reported one at the same file within
- * DEDUPE_LINE_PROXIMITY lines — the main-pass version is kept. Returns a new
- * array; inputs are not mutated.
+ * dropped only when the main pass already reported one at the same location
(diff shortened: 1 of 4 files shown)


alfhen:
Triage of the review finding (severityRank collapsing non-error severities): **dismissed as false positive**. `AgentFinding.severity` is the two-value union `'error' | 'warning'` at the type level (types.ts:62) AND at the runtime validation gate — `isValidFinding` (agent-client-shared.ts:214) rejects any other value before a finding enters the pipeline, so a main-pass 'info' finding cannot exist. The binary rank is exhaustive over the actual domain; adding a third tier would be dead code guarding a phantom severity.

PR #764, opened 2026-07-15
test(review): capture PR #752 broad-catch miss as error-swallowing characterization fixture

alfhen:
## Provenance

Captures getlien/lien PR #752 ("fix(review): stop silently dropping inline comments when batch review fails") at commit `295cc7e0f5bfe2a263d804f39905a261298b45dc` — an intermediate head, **not** the PR's current tip (the branch has since gained follow-up fix commits).

**The bug:** `postPRReview`'s catch block (`packages/review/src/github-api.ts`, ~lines 268-286 at this SHA) catches ANY error from `octokit.pulls.createReview` and — as long as `comments.length > 0` — unconditionally falls back to `postBodyThenRetryCommentsIndividually`, with no check that the failure was actually the 422 anchor-validation case the fallback was designed for. An auth/401, rate-limit/429, 5xx, or plain network error all take the identical path: every per-comment retry fails the same way the batch did, so the function returns `{ posted: 0, dropped: <all> }` instead of throwing — an infrastructure outage silently repackaged as a degraded "success".

On 2026-07-15, Lien Review reviewed the PR at exactly this SHA and produced 4 unrelated findings (start_line strip nit, doc-comment contradiction, `postReviewComment` discard, `index.ts` re-export breaking change) — missing this one. CodeRabbit, reviewing the same SHA, caught it: *"Update the catch handling in postPRReview so postBodyThenRetryCommentsIndividually is invoked only for GitHub validation failures... rethrow authentication, rate-limit, server, timeout, and other errors."*

## Active-rule verification

`build-prompts.ts` against this fixture: active = `['structural-analysis', 'edge-case-sweep', 'incomplete-handling', 'error-swallowing', 'boundary-change', 'stale-duplicate', 'doc-truth']`; skipped = `['concurrency-race', 'untrusted-input-validation']`. No trigger gap — `error-swallowing` and its `ruleCandidates` fallback (`incomplete-handling`) are both active.

## Capture-pr.ts fix (found while building this fixture)

`gh pr view` always returns the PR's **current** description. Lien Review's own `
### Lien Review

> [!NOTE]
> **Low Risk**
>
> This PR adds a characterization test fixture for a known error-swallowing miss (PR #752) and a small sanitization helper in the fixture-capture script. The new `stripLienStatsBadge()` function correctly truncates a captured PR body at the first Lien stats marker when `--sha` pins a historical commit, preventing later review badges from leaking ground truth into the fixture prompt. No production code paths are changed, and the helper's behavior matches its documentation for realistic inputs.
>
> - Added `stripLienStatsBadge()` to `capture-pr.ts` to strip dynamic lien-stats badges from captured PR bodies when using `--sha`.
> - Added a new `error-swallowing` characterization fixture documenting the undiscriminated-catch bug in PR #752.
> - Updated harness README to record the newly documented miss.

<sup>Reviewed by [Lien Review](https://lien.dev) for commit b4e24c1. Updates automatically on new commits.</sup>

## Summary by CodeRabbit

- **Tests**
  - Added a new `error-swallowing` fixture with assertions covering catch-block discrimination failures and the resulting lost-vote impact.
  - Enhanced fixture capture so PR descriptions are sanitized by stripping dynamic lien review badge content when generating historical fixtures.

- **Documentation**
  - Updated documented-miss fixture records to include the newly tracked `error-swallowing` scenario.

### DIFF
--- a/packages/review/test/harness/README.md
+++ b/packages/review/test/harness/README.md
@@ -162,7 +162,7 @@ prompt/pipeline from reaching it.
 
 | Rule | PR | Fixture | Why |
 |---|---|---|---|
-| _(none currently)_ | | | The one entry this table used to carry — `doc-truth/pr658-search-code-rename` — was **promoted to the canary corpus on 2026-07-04** (see the table above): a `--calibrate 10` run scored 10/10 on Finding A and, after re-verifying via `build-prompts.ts` that the guidance-surface passthrough (#665) now carries the fixture's second real finding (`plugins/claude/hooks/augment-explore-task.sh:64`) into the prompt, 9/10 votes from that same calibration run independently reproduced it too — so it's now asserted (Tier 2) alongside Finding A rather than only documented. Full evidence chain and per-run breakdown live in the fixture's `.assertions.ts` header comment. |
+| `error-swallowing` | #752 @ `295cc7e` | `error-swallowing/pr752-undiscriminated-catch-salvage` | `postPRReview`'s catch block salvages EVERY `createReview` error (auth, rate-limit, 5xx, network) the same way it salvages a 422 anchor-validation failure, silently degrading an infra outage into `{posted:0, dropped:<all>}` instead of throwing. CodeRabbit caught it same-SHA on 2026-07-15; Lien Review's live review of the same commit missed it (4 unrelated findings instead). 3-vote baseline on the sanitized fixture: 0/3 — reproduces the live miss. See the fixture's header for the capture-bug (stale `lien-stats` badge leak) found and fixed while building it. |
 
 Run the full multi-model sweep:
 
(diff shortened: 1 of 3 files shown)


alfhen:
Both nitpicks addressed in b4e24c1:

- **`meta.body ?? ''` double eval (lines 464-468):** Fixed — hoisted to a local `originalBody` variable, reused for both the strip call and the comparison.
- **Parameterized `<!-- lien:{sectionId} -->` marker (lines 207-210):** Fixed — this was actually a real gap, not cosmetic. `sectionMarkers()` in `github-api.ts` emits `<!-- lien:{sectionId} -->` as a real, already-used alternative to the legacy `<!-- lien-stats -->` literal, so the old `indexOf` check could miss stripping a parameterized badge and let spoiler content back into a captured fixture. Widened to a regex that matches either form.

Ran the full gate chain (format, lint, typecheck, build, test, `lien delta`) — all green.

PR #765, opened 2026-07-15
fix(review,action): fail the check when the review provider fails every request

alfhen:
## The bug (live repro)

PR #764 (2026-07-15, before OpenRouter credits were topped up): OpenRouter returned `402 insufficient credits` on **every** provider request. The agent-review plugin analyzed zero code, and the `lien-stats` badge said so plainly — yet the action's own check concluded **green**. An enabled, paid-for review that couldn't run at all silently passed.

## Root cause

Not where it first looked. PR #738/#739 ("a starved review fails loudly instead of passing as clean") already made the **review-core layer** correctly classify a total provider outage:

- `OpenAIAgentClient.run()` (`packages/review/src/plugins/agent/openai-client.ts`): a 402 is a non-retryable 4xx, throws immediately with `completedTurns === 0` → `neverRan: true`.
- `appendNeverRanNotice` (`packages/review/src/plugins/agent/index.ts`) escalates that into an **error-severity** `summary` finding (`metadata.neverRan: true`).
- `determineConclusion` (`packages/review/src/engine.ts`) maps any error-severity finding to conclusion `'failure'`.

All of that is already covered by existing tests (`plugins-agent-budget.test.ts`) and was working correctly on `main` before this PR.

The actual remaining gap is one layer further down, in `packages/action/src/index.ts`:

```ts
function exitCodeFor(failOn, conclusion, errorCount, warningCount): number {
  if (failOn === 'never') return 0;   // <-- unconditional, no matter what caused `failure`
  ...
}
```

`fail-on` defaults to `'never'` (advisory-by-default is deliberate, documented policy — see `packages/action/README.md`). But `exitCodeFor` never distinguished "an ordinary advisory finding" from "the review never ran at all" — under the default, **both** collapsed to exit 0. The action's job is the *only* status check (no separate Checks-API run), so exit code 0 == green PR check, full stop. That's why the 402 still passed as green four days after #738/#739 merged: those PRs fixed the `conclusion` computation, not the exit-code decision that actually gates the check.

## Design decision

`packages/action/README.md`'s "Fail-loudly guarantee" section documented the *old* behavior as intentional: "Under the default `fail-on: never` the job itself still exits `0`." I'm overriding that for this one case, and updating the docs to match:

**A total provider failure is an operational failure, not an advisory finding — it now fails the check regardless of `fail-on`, including the default `never`.** The general "advisory by default" philosophy is right for opinionated code-quality findings (you shouldn't be forced to gate CI on an LLM's opinion), but it doesn't hold when the tool did nothing at all. A user who enabled and is paying for a review deserves a red check when it silently didn't happen — otherwise the greenness is a lie.

A **partial** failure (some turns completed before hitting a budget/turn limit) is unaffected — that's a genuine advisory `warning`-severity finding and still obeys `fail-on` exactly as before (verified via existing + new tests). And when the agent isn't enabled at all (no API key), behavior is unchanged — complexity-only review, no never-ran signal is ever produced.

## The matrix

| Case | `ReviewCoreResult.providerFailure` | conclusion | exit code under `fail-on: never` |
|---|---|---|---|
| Total provider failure (every request 402/401/5xx) | `true` | `failure` | **1** (new — was 0) |
| Partial failure (some turns ran, then bailed) | `false` | `neutral` | 0 (unchanged) |
| Agent not enabled (no key) | `false` — no plugin runs | `success` | 0 (unchanged) |
| Ordinary bug/summary findings | `false` | varies | 0 (unchanged — still advisory) |

## Changes

Two commits. First lands the fix at the layer that actually gates the check; second is a follow-up refactor (from a parallel design-review pass) that elevates the signal to a typed field instead of the action package re-deriving it from finding metadata shape:

- `packages/review/src/plugins/agent/index.ts`: new exported `hasProviderFailure(findings)`, colocated with `appendNeverRanNotice` — the single source of truth for the never-ran contract.
- `packages/review/src/review-pr.ts`: `ReviewCoreResult` gains `providerFailure: boolean`, computed once in `analyzeAndPresent()` via `hasProviderFailure`. The two early-return paths (`emptyResult` — no analyzable files, or the complexity pipeline failing before the engine runs) always set it `false` — those failure modes are unrelated to the LLM provider.
- `packages/action/src/index.ts`: `exitCodeFor` takes `providerFailure` and forces exit 1 before consulting `fail-on`; `finishRun` reads `result.providerFailure` directly and logs a clear, actionable message (names the provider error class from the finding's own message, plus generic remediation: 402→add credits, 401/403→check the API key secret, 5xx→re-run).
- `packages/action/test/finish-run.test.ts`: covers total-failure-forces-exit-1 under every `fail-on` mode, a findings-based `failure` conclusion (no provider failure) staying advisory under `never`, and partial-failure staying advisory.
- `packages/review/test/plugins-agent-budget.test.ts`: new `hasProviderFailure` suite (true/false/partial/doc-pass-only/empty cases) next to the existing `appendIncompleteNotice` tests.
- `packages/action/README.md` + `packages/site/docs/guide/lien-review.md`: updated the "Fail-loudly guarantee" section and the `fail-on` table row / "Blocking a PR" paragraph (previously unqualified "never fails CI" claims) to describe the new behavior.

## Changeset

None — both touched packages (`@liendev/action`, `@liendev/review`) are `private: true` (unpublished); nothing here touches a published package (`parser`/`core`/`cli`).

## Harness evidence

This PR touches `packages/review/src/plugins/agent/index.ts`, which trips the harness-attestation gate. The only change there is `hasProviderFailure(findings)` — a pure function reading a boolean off finding metadata. It touches zero prompt/rule/trigger content (`system-prompt.ts`, `rules.ts`) and cannot affect what the agent does or finds, so a fresh `--calibrate 10` run would be pointless spend re-verifying output quality that literally cannot have changed. Evidence instead follows the deterministic zero-LLM proof pattern from #743/#750:

**Method:** for all 23 fixtures on disk under `packages/review/test/harness/fixtures/`, render `system prompt + initial message` via `build-prompts.ts` twice — once with `packages/review/src` checked out from `origin/main`, once from this branch's `HEAD` — and diff byte-for-byte.

```bash
git checkout origin/main -- packages/review/src
# render all 23 fixtures to A/ via: npx tsx test/harness/build-prompts.ts <fixture> > A/<name>.json
git checkout HEAD -- packages/review/src   # confirms `git status` clean before AND after
# render all 23 fixtures to B/ the same way
cmp -s A/<name>.json B/<name>.json   # for every fixture
```

**Result: 23/23 byte-identical.** Zero diffs. Every existing `--calibrate 10` certification for every rule (including the canaries) remains valid unchanged, because the exact bytes an agent run would see are provably unmodified by this PR.

## Gates

`npm run format:check && npm run lint && npm run typecheck && npm run build && npm test` — all green across every package (parser, core, review, cli, action — 2000+ tests, zero failures; lint 0 errors, only pre-existing warnings elsewhere). `lien delta` exits 0 (informational "worsened" time-to-understand/cognitive notes on `exitCodeFor`/`finishRun`/`analyzeAndPresent`, all still far under threshold — not a crossing). Re-verified after the README wording fixes below.

## Test plan

- [x] Unit test: total provider-failure finding forces exit 1 under `fail-on: never/error/any`
- [x] Unit test: a findings-based `failure` conclusion (no provider failure) stays exit 0 under `fail-on: never` (unchanged advisory semantics)
- [x] Unit test: `hasProviderFailure` pure classifier (true/false/partial/doc-pass-only/empty cases)
- [x] Deterministic byte-diff harness evidence: 23/23 fixtures byte-identical (see above)
- [x] CI green (build-and-test, review, CodeRabbit, docs build, release smoke test, complexity delta)
- [x] Lien Review triage: two rounds of findings, all fixed - stale PR-description claim about hasProviderFailure's location (2nd commit), then four remaining unqualified "never fails CI/blocks a PR" spots across README.md and the site guide (2 caught per round, 3rd & 4th commits). Final pass on commit fbb106c: clean, 0 findings.

---

### Lien Review

➡️ **Stable** - 4 pre-existing issues in touched files (none introduced).
🔗 **Impact**: 1 high-risk file(s) with 11 total dependents
| ⏱️ time to understand | 4 | +0 |

> [!NOTE]
> **Low Risk**
>
> This PR fixes the silent-green-check bug for total LLM-provider failures by adding a typed `providerFailure` signal in `@liendev/review` and forcing the GitHub Action to exit 1 regardless of `fail-on`. The change is narrow, well-scoped, and covered by new unit tests. The logic correctly distinguishes total provider failures (every request failed, `neverRan: true`) from partial runs, ordinary findings, and unrelated pipeline failures, and the documentation accurately reflects the new behavior.
>
> - Added `hasProviderFailure()` classifier and `ReviewCoreResult.providerFailure` field to surface total provider failures as an operational signal.
> - Updated `exitCodeFor()` in `packages/action/src/index.ts` to return 1 before consulting `fail-on` when `providerFailure` is true.
> - Added regression tests for total-failure, partial-failure, and advisory-boundary cases; updated README/site docs to describe the new fail-loudly guarantee.

<sup>Reviewed by [Lien Review](https://lien.dev) for commit fbb106c. Updates automatically on new commits.</sup>

### DIFF
--- a/packages/action/README.md
+++ b/packages/action/README.md
@@ -95,7 +95,7 @@ exact cost of every run is printed in the job's step summary (the
 | `threshold`           | no       | `15`                            | Complexity threshold above which violations are reported.                                                                                 |
 | `review-types`        | no       | `complexity,bugs,summary`       | Comma-separated review types to enable. `complexity` toggles the complexity check. `bugs`, `architectural`, and `summary` all come from the single agent reviewer, so they switch it on/off as a group (and only when an API key is set) — they can't be toggled independently. |
 | `block-on-new-errors` | no       | `false`                         | Post `REQUEST_CHANGES` (instead of `COMMENT`) when the PR introduces new error-level complexity violations.                               |
-| `fail-on`             | no       | `never`                         | Whether the review fails the check (so a Required check can block the PR). Default `never` — **advisory**, never fails CI. Opt into gating with `error` (a failure conclusion / new error-level findings) or `any` (any error or warning finding). |
+| `fail-on`             | no       | `never`                         | Whether the review fails the check (so a Required check can block the PR). Default `never` — **advisory** for review findings. Opt into gating with `error` (a failure conclusion / new error-level findings) or `any` (any error or warning finding). Exception: a total LLM-provider failure always fails the check regardless of this setting — see [Fail-loudly guarantee](#fail-loudly-guarantee). |
 
 > The action posts **no check run of its own** — the workflow job is the single
 > status check. Findings surface as workflow annotations (inline on the diff),
@@ -145,13 +145,15 @@ only covers that one model.
 
 ## Blocking a PR on the review
 
-By default the review is **advisory** (`fail-on: never`) — it never fails CI, so
-adding the action can't break anyone's pipeline. To gate merges on it, opt in by
-setting `fail-on` and marking the workflow's job as a **Required status check**
-in your branch protection rules. With `fail-on: error` the action exits non-zero
-only when the review's overall conclusion is a failure (driven by
-`block-on-new-errors`); `fail-on: any` is stricter (any error- or warning-level
-finding fails the check).
+By default the review is **advisory** (`fail-on: never`) for its own findings —
+it never fails CI on those, so adding the action can't break anyone's
+pipeline. To gate merges on findings, opt in by setting `fail-on` and marking
+the workflow's job as a **Required status check** in your branch protection
+rules. With `fail-on: error` the action exits non-zero only when the review's
+overall conclusion is a failure (driven by `block-on-new-errors`); `fail-on:
+any` is stricter (any error- or warning-level finding fails the check). A
+total LLM-provider failure is a separate case that always fails the check,
+even under `fail-on: never` — see below.
 
 ## Fail-loudly guarantee
 
--- a/packages/action/README.md
+++ b/packages/action/README.md
@@ -156,14 +156,19 @@ finding fails the check).
 ## Fail-loudly guarantee
 
 If the agent review's main pass never runs at all — every request to the LLM
-provider failed (exhausted key, a terminal provider error, etc.) — Lien marks
-the result with an **error-severity finding** and a **`failure`** conclusion
-naming the cause, instead of a clean-looking review. That finding is what
-`fail-on: error`/`any` act on, so a starved review fails the check under
-either gating mode. Under the default `fail-on: never` the job itself still
-exits `0` (advisory stays advisory), but the step summary, PR description, and
-`conclusion` output make the failure impossible to mistake for "no issues
-found."
+provider failed terminally (insufficient credits, an invalid/expired key, a
+provider outage, etc.) — Lien marks the result with an **error-severity
+finding** and a **`failure`** conclusion naming the cause, instead of a
+clean-looking review.
+
+This is treated as an **operational failure, not an advisory finding**: a
+review that never ran isn't something `fail-on` gates on, because there's
+nothing to be advisory *about* — no code was analyzed. **The check fails
+regardless of `fail-on`, including the advisory default `never`.** A partial
+run (some turns completed before it bailed on a budget/turn limit) is
+different — that's a genuine advisory finding and still obeys `fail-on` as
+before. Either way, the step summary, PR description, and `conclusion` output
+make the failure impossible to mistake for "no issues found."
 
 ## Fork PRs
 
--- a/packages/action/src/index.ts
+++ b/packages/action/src/index.ts
@@ -23,7 +23,7 @@ import {
 
 import { readInputs, type FailOn } from './inputs.js';
 import { loadContext } from './context.js';
-import { writeStepSummary, writeOutputs, countErrors, hasProviderFailure } from './summary.js';
+import { writeStepSummary, writeOutputs, countErrors } from './summary.js';
 import { actionLogger, group, endGroup, annotate } from './logger.js';
 
 /**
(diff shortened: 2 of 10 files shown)


alfhen:
**Follow-up commit** (per a parallel design-review pass): threaded the never-ran detection through `ReviewCoreResult` as a first-class typed field instead of having the action package re-derive it from finding `metadata` shape.

- `packages/review/src/plugins/agent/index.ts`: new exported `hasProviderFailure(findings)`, colocated with `appendNeverRanNotice` — the single source of truth for the `neverRan` contract.
- `packages/review/src/review-pr.ts`: `ReviewCoreResult` gains `providerFailure: boolean`, computed once in `analyzeAndPresent()`. The two early-return paths (`emptyResult` — no analyzable files, or the complexity-analysis pipeline failing before the engine runs) always set it `false`, since those failure modes are unrelated to the LLM provider.
- `packages/action/src/index.ts` / `summary.ts`: `finishRun` now reads `result.providerFailure` directly; the action-side `hasProviderFailure` classifier is removed (no longer needed — the action package no longer needs to know the internal `metadata.neverRan` shape).
- Test coverage relocated/added accordingly: `packages/review/test/plugins-agent-budget.test.ts` gets a `hasProviderFailure` suite (true/false/partial/doc-pass-only/empty cases) next to the existing `appendIncompleteNotice` tests; `packages/action/test/finish-run.test.ts`'s total-failure tests now set `providerFailure` directly on the `ReviewCoreResult` fixture (mirroring what `review-pr.ts` would actually produce) instead of relying on `finishRun` to re-derive it from findings.

Behavior is unchanged from the first commit — this is purely an encapsulation improvement. Full gate chain re-verified green (format/lint/typecheck/build/test across all 5 packages, `lien delta` exit 0).

alfhen:
CI status: everything is green except **"Require harness evidence or bypass label."** It fired because this PR touches a file under \`packages/review/src/plugins/agent/**\` (\`index.ts\`) — but the only change there is a new exported pure function, \`hasProviderFailure(findings)\`, which reads a boolean off finding metadata. It touches zero prompt/rule content (\`system-prompt.ts\`, \`rules.ts\`) and cannot change what the LLM agent does or finds.

Per CLAUDE.md this gate exists for actual rule/prompt changes, with a documented escape hatch (\`skip-harness-gate\` label) for non-behavioral changes like this one. I'm not applying that label myself — it's a mandatory CI gate and self-authorizing a bypass isn't my call. Flagging for a human decision: apply \`skip-harness-gate\`, or if you'd rather the change to \`plugins/agent/index.ts\` live somewhere outside that directory to sidestep the path-based trigger entirely, let me know and I'll relocate \`hasProviderFailure\`.

PR #773, opened 2026-07-15
feat(core,cli): structural ranking boosts for search_code

alfhen:
## Summary

`search_code` currently ranks purely by BM25 — every file is equal regardless of how central it is to the codebase. Lien already knows the dependency graph; this uses it as a ranking signal.

- **Ranking boost** (`packages/core/src/vectordb/sqlite/fts-search.ts`): within the already-fetched BM25 candidate window (`limit + 20` overfetch), each result's bm25-derived relevance ratio is blended with a structural-importance signal before the final `.slice(0, limit)`:

  ```
  final = ratio * (1 + α · log(1 + dependentCount))
  α = 0.15
  ```

  `ratio` is fts-search's existing (0, 1] bm25 ratio (1 = best lexical match in the result set). `dependentCount` is how many other indexed files import this chunk's file. `log1p` keeps the boost sublinear — going from 0→1 dependents matters far more than 100→200. The result is never less than `ratio`, so the boost only ever **promotes within the existing candidate window** — it can't pull in a result BM25 never considered relevant, and `Array.prototype.sort` is stable so exact ties keep their original BM25 order. At a realistic max (~200 dependents, a hub file), the multiplier caps at ≈1.8x.

  **Escape hatch**: `LIEN_STRUCTURAL_RANKING=off` disables the re-sort and restores pure BM25 ordering (unit-tested).

- **dependentCount data source** (`packages/core/src/vectordb/sqlite/dependent-counts.ts`): `dependentCount` is NOT the authoritative `get_dependents` count — computing that per query (import-index rebuild + fuzzy path/re-export matching) is O(files × unique imports) and explicitly out of scope for "no new expensive per-query computation." Instead this resolves only relative imports (`./foo`, `../bar`) against the importer's directory and does exact extension-stripped path matching — no re-export chains, no Python/PHP heuristics. Bare package specifiers simply don't match anything, which **undercounts, never overcounts** — the safe failure direction for a signal that only ever boosts, never suppresses. This is O(total imports), computed once per SQLite connection and cached in a `WeakMap<Database, Map>` keyed on the connection object — `SqliteBackend.reconnect()` (which the MCP server always calls after any index write's version bump) creates a fresh `Database` object, so the cache self-invalidates on every index update with no extra bookkeeping.

- **Richer metadata**: `search_code` results now carry `metadata.dependentCount` inline (added to `search_code`'s allowlist in `metadata-shaper.ts`) so an agent can triage a result's blast radius without a follow-up `get_dependents` call. This is unconditional — populated regardless of the ranking flag above, since it's informational, not a ranking behavior change.

- **Skipped**: test-file deprioritization for "non-test-intent" queries. Detecting a *result's* file is a test (`isTestFile(path)`) is trivially path-based; detecting the *query's* intent is not — it would need a keyword-sniffing classifier, which is neither trivial nor deterministic. Per the task's own escape hatch ("ONLY if trivially detectable, else skip"), this is left out. In practice `dependentCount` already gives test files a soft, deterministic disadvantage for free: test files are rarely imported by other files, so they usually carry `dependentCount: 0` (see the dogfood table below, "metadata allowlist shaping" query).

## Dogfood comparison (this repo's own index, ranking ON vs OFF)

Built via `node packages/cli/dist/index.js index --force`, queried directly through `createVectorDB` with `LIEN_STRUCTURAL_RANKING` toggled. 8 queries, top-5 each.

**5/8 queries: no change.** The top BM25 hit dominated so strongly that the boost never reordered anything: *chunk overlap config*, *delta gate exit code*, *review budget starvation*, *bm25 keyword search*, *complexity threshold violation*.

**3/8 queries reordered — always at positions #3-5, never #1 or #2:**

| Query | OFF (pure BM25) | ON (boosted) | Verdict |
|---|---|---|---|
| worktree overlay rebuild | #4 `worktree-aware-indexing.md:445` ("...Verified" appendix, deps=0) | #4 `overlay-index.ts:20` (`OverlayBuildResult`, deps=6) | Improvement — surfaces the actual rebuild implementation over a docs appendix |
| metadata allowlist shaping | #5 `metadata-shaper.test.ts:37` (test fixture, deps=0) | #5 `metadata-shaper.ts:1` (production file, deps=1) | Improvement — the soft test-deprioritization effect described above, in action |
| dependency analyzer risk level | #3 `mcp-tools.md:338` (docs: "get_dependents > Risk Levels", deps=0) | #3 `test-association.md:554`; `mcp-tools.md:338` drops out of top-5, replaced by `complexity-analyzer.ts:434` (`enrichWithDependencies`, deps=2) and an ADR doc | Debatable — loses a strongly on-topic docs hit from the visible window. Mitigated: `dependency-analyzer.ts:532` (the actual `analyzeDependencies` function, deps=3) is unchanged at #2 in both ON and OFF, so the implementation is still the top code hit either way |

**Honest take**: the boost worked as designed — it never touched the #1 result in any of the 8 queries, and where it did reorder, it was consistently within the bottom half of the top-5. Two of three changes look like clear wins (production code over its own test fixture; an implementation file over a docs appendix). The third is a real tradeoff (a strong docs hit falls out of the top-5), but the underlying code result an agent actually needs was never at risk. Shipping at α=0.15; if this proves too aggressive in practice, the fix is a one-line α reduction, not a redesign.

## Files

- `packages/core/src/vectordb/sqlite/dependent-counts.ts` (new) + test
- `packages/core/src/vectordb/sqlite/fts-search.ts` — boost math + re-sort + dependentCount attachment
- `packages/core/src/vectordb/types.ts` — widened `SearchResult.metadata` with optional `dependentCount`
- `packages/cli/src/mcp/utils/metadata-shaper.ts` — `dependentCount` added to `ToolResultMetadata` + `search_code`'s allowlist (allowlist keys now typed against `SearchResult['metadata']` instead of raw `ChunkMetadata`, since `dependentCount` is a search-path-only addition, not part of the parser's chunk schema)
- `packages/cli/src/mcp/tools.ts` — documents the ranking behavior + `dependentCount` field on `search_code`

## Validation

- Unit tests (no LLM spend): blend math (never demotes, monotonic, sublinear, custom α, bounded at realistic max), tied-BM25 ordering resolved by dependentCount, flag-off preserves pure BM25 order, dependentCount metadata attach/omit/zero-is-not-empty.
- Full gate chain: format/lint/typecheck/build/test all green; `lien delta` exit 0 (no new threshold crossings — see CI checks below).

---

### Lien Review

> [!NOTE]
> **Low Risk**
>
> This PR adds a structural-importance boost to FTS search results and exposes `metadata.dependentCount` on `search_code` responses. The implementation is well-isolated: the boost is computed within the existing BM25 overfetch window, capped at 2×, and gated by `LIEN_STRUCTURAL_RANKING=off`. The per-connection `WeakMap` cache self-invalidates on reconnect. The score/relevance metadata intentionally reflects pure BM25 while ordering uses the boosted value, and this tradeoff is documented explicitly in both code comments and the tool description. No logic errors, breaking changes, or silent failure modes were found.
>
> - New `dependent-counts.ts` computes cheap, conservative per-file import counts from relative imports only
> - `fts-search.ts` re-sorts the BM25 candidate window by `ratio * min(2, 1 + 0.15 * log1p(dependentCount))`
> - `metadata-shaper.ts` adds `dependentCount` to `search_code` output and strips it for other tools
> - `types.ts` widens `SearchResult.metadata` with the optional `dependentCount` field

<sup>Reviewed by [Lien Review](https://lien.dev) for commit e344b0e. Updates automatically on new commits.</sup>


alfhen:
Addressed in 9008507: the "never crosses a bm25 band" claim was wrong (confirmed by your ratio=0.6/dependentCount=200 example) — no nonzero multiplicative boost can guarantee that near a continuous band boundary. Fixed by (1) rewriting the doc comments / changeset to state the real, honest guarantee instead of an unachievable one, and (2) adding a hard `MAX_STRUCTURAL_BOOST_MULTIPLIER = 2` ceiling to bound worst-case growth for pathological dependentCounts (normal behavior at the documented ~200-dependent realistic max is unaffected, since 1.8x < 2x). Added two tests: one pinning the ceiling at an extreme dependentCount, one pinning the now-documented band-crossing behavior itself so it's a known, tested tradeoff rather than a silent surprise. Re-ran the dogfood comparison post-fix — identical output, since this repo's max dependentCount (8) is far from where the cap matters.

alfhen:
Addressed in e344b0e: this is intentional, not a bug — recomputing score/relevance from the boosted value would let a highly-connected-but-weakly-matching file inflate its own relevance band, which would make "relevance" meaningless as a pure lexical-match indicator (and `dependentCount` is already exposed in metadata, with the exact formula documented, so the boosted value is fully reconstructable by a caller who wants it). Documented this explicitly in three places (scoreRow, keywordSearch, and the search_code tool description) and added a test pinning the exact scenario you raised: a lower-scoring-but-popular file ranks above a higher-scoring one while each still reports its own pure bm25 score/relevance.

PR #917, opened 2026-07-28
docs(dogfood): protocol + workflow for foreign-repo dogfood of MCP tools and nudges

alfhen:
## Why

Lien's own `CLAUDE.md` has a **"Lien MCP Tools — MANDATORY Usage"** section telling any agent working here to call `get_files_context` before editing and `get_dependents` before touching an exported symbol. The nudge plugin's whole premise is producing those behaviors in agents that were *never told* to do them.

So dogfooding the nudges inside this repo measures the instructions, not the product. This is the pre-registration for a session that runs **outside** it.

## What this adds

- **`docs/development/dogfood-oss-corpus-protocol.md`** — the frozen protocol: five contamination controls (all gates), a nine-repo / nine-language corpus, six frozen metrics, phase gates, abort criteria, cost. It extends the contamination plumbing already proven by [`nudge-ab-v2`](docs/development/nudge-ab-v2-protocol.md) — `assertNoAncestorClaudeMd`, the per-process `--settings` override that disables the ambient `lien@lien` plugin — from synthetic fixtures to real code.
- **`scripts/experiments/dogfood-oss/workflow.js`** — the orchestrator, in three sequential stages (`provision` → `surface` → `behavioral`) so a red gate stops the spend and a human reads results between stages.

Nothing is executed by this PR. It is the plan plus the instrument.

## Scope

| Surface | Coverage |
|---|---|
| MCP tools | all 6, one agent each: doc-promises-vs-actual in both directions, plus a per-tool edge matrix |
| Nudge hooks | all 10, zero-LLM: output channels, 12 env kill switches, habituation guard, fail-open, `session_id` traversal rejection, `stop_hook_active` (a blocking Stop hook without it is an infinite loop), latency vs the 5 s timeout |
| Behavioral | 9 repos × 2 task shapes = 18 headless sessions on ordinary prompts |
| Truth audit | every nudge that fired, refuted by two independent skeptics under different lenses |
| First-run UX | the advertised "No per-project setup required" claim, tested honestly |

## Corpus screening already found something

Screening candidates against control C1 on 2026-07-28 **rejected three of nine** first-choice repos for shipping their own agent instructions:

| Rejected | Language | Reason |
|---|---|---|
| `cli/cli` | Go | ships `AGENTS.md` |
| `guzzle/guzzle` | PHP | ships `AGENTS.md` **and** `CLAUDE.md` |
| `onevcat/Kingfisher` | Swift | ships `AGENTS.md` **and** `CLAUDE.md` |

(`rubocop/rubocop`, screened for Ruby, ships both too.)

Two consequences. Roughly a third of popular OSS repos now carry agent-instruction files, so *"clone a random repo"* is **not** a contamination control on its own — C1 must be enforced mechanically, and the quarantine gate is load-bearing rather than ceremonial. And the three rejects were Go, PHP and Swift: three of the four languages whose test-association layer was measured broken in the 2026-07-25 sweep. Dropping them would have gutted the most valuable coverage in the protocol, so each was **replaced in-language**.

## Method commitments worth reviewing

- **Truthfulness (M2) is the headline metric, not response rate.** Three of the nudges are built on test associations. If the #907/#909/#911/#913/#914 fixes didn't hold on code nobody tuned them against, those nudges don't go quiet — they assert something *false*, which is the worst failure mode this product has. Any refuted claim is filed HIGH regardless of how helpful the true ones were.
- **Single-arm, and it stays that way.** Response rates are observation, never lift. The only publishable Lien nudge lift claims remain the paired-arm A/B v2 results (blast 10/10 vs 3/10, verify 10/10 vs 0/10). A paired arm over this corpus is a legitimate follow-up at ~2× the Phase 3 cost, explicitly out of scope, and may not be back-fitted.
- **Task-decoupled by construction**, with headroom asserted at zero LLM cost before any session runs — the v1-null lesson.
- **Silent plumbing looks exactly like a behavioral null**, so Phase 2 must prove a nudge was emitted before Phase 3 is allowed to conclude an agent ignored it.

## Cost

Phases 0/1/2/5 are deterministic or Claude-subagent work. Phase 3's 18 headless sessions bill against the Claude subscription, not per-token API credit. **No `OPENROUTER_API_KEY` spend at all** — this is not a repeat of the harness-calibration pattern. All dispatched agents run on **Sonnet** per standing model policy; Phase 4's adversarial pass is escalated to Opus per-run via `args.auditModel` only if a refutation looks unconvincing.

## Dogfooding evidence

Per the repo's own mandate, the checks that actually apply to a docs + `scripts/experiments/` change:

- `node .github/scripts/docs-truth.mjs` → **`87 files checked, no violations`** (the new doc included)
- `node --check scripts/experiments/dogfood-oss/workflow.js` → clean; 9 corpus entries, 8 `model: MODEL` + 1 `model: AUDIT_MODEL`
- All four cross-references in the doc resolve on disk
- Corpus C1 screening run live against the GitHub API — that's what produced the three rejections above
- Every factual claim in the protocol verified against `origin/main` source, not memory: hook output channels (`additionalContext` / `updatedInput` / `{"decision":"block"}`), `stop_hook_active` guard, the 12 env switches, `plugin.json`'s "No per-project setup required", and `lien` genuinely not being on PATH

Gates 1–5 (`format:check`, `lint`, `typecheck`, `build`, `test`) glob `packages/**` only, so this change is outside all of them — same as the existing `nudge-ab-v2` kit. Running them would re-verify main's state, not this diff, so I've said what I checked instead of implying a green chain.

## Incidental finding (not fixed here)

`CLAUDE.md` states gate 6 `lien delta` "is **not** [CI-backstopped] — there is no CI step for it, so it's agent-honor-only." That is now stale: `.github/workflows/ci.yml` has a `delta:` job whose own comment says it exists to give the gate a CI backstop. Left alone to keep this PR to one concern.

## Summary by CodeRabbit

* **New Features**
  * Added a structured evaluation workflow for testing MCP tools and plugin behavior across a fixed repository corpus.
  * Added staged validation covering tool responses, hook behavior, live trials, adversarial audits, cold-start checks, and telemetry.
  * Added machine-readable findings, trial results, audit verdicts, trust reporting, and ranked issue summaries.

* **Documentation**
  * Added a comprehensive protocol defining corpus selection, contamination controls, evaluation tasks, metrics, abort criteria, and reporting guidelines.

---

### Lien Review

> [!WARNING]
> **1 issue found** · Low Risk
>
> This PR adds a frozen dogfood protocol document and four experiment orchestration scripts for exercising Lien's MCP tools and nudge plugin against foreign OSS repos. The scripts are careful about contamination controls, error handling, and measurement fidelity. The one concrete defect is a platform-dependent corpus-root fallback in workflow.js that diverges from provision.mjs on systems where `os.tmpdir()` is not `/tmp`, which would cause the orchestrator to fail to locate the provisioned corpus.
>
> - New dogfood protocol doc and instrument scripts under scripts/experiments/dogfood-oss/
> - mcp-call.mjs provides a stdio MCP client that treats tool-level errors as successful measurements
> - provision.mjs clones, C1-scans, indexes, and records M6 coverage for the nine-repo corpus
> - trial.mjs runs one headless Claude session per (repo, task) and records nudge events
> - workflow.js drives the three sequential stages

<sup>Reviewed by [Lien Review](https://lien.dev) for commit 2f51e67. Updates automatically on new commits.</sup>

---


alfhen:
Both CodeRabbit findings were real. Fixed in e5c0c94f.

**1. `.filter(Boolean)` before `.slice(TOOLS.length)` (line 409) — confirmed, fixed as suggested.**

Reproduced before fixing, with tool agent #3 resolving null:

```
OLD tools: ["t1","t2","t4","t5","t6","hA"]  hooks: ["hB"]
NEW tools: ["t1","t2","t4","t5","t6"]       hooks: ["hA","hB"]
```

Hook report `hA` was being relabeled as a tool report *and* the hook partition silently lost an entry. Now partitioned at the fixed boundary first, then each side filtered independently. Worth catching — a protocol whose whole point is distrusting silent partial coverage should not itself mislabel which agent produced which report.

**2. `LIEN_STRUCTURAL_RANKING` (line 354) — real inconsistency, fixed in the opposite direction to the suggestion.**

The finding is correct that doc and script disagreed: the protocol's Phase 2 checklist listed this switch among the hook kill switches, and the script's Part A/B prompts omitted it.

I fixed the doc rather than the script. As the finding itself notes, the switch gates `search_code`'s ranking and is not one of the ten hook switches — so adding an assertion for it to the hook-plumbing phase would put a tool-ranking check in a phase that has no tools in scope. The script was already right: `search_code`'s Phase 1 edge matrix covers `LIEN_STRUCTURAL_RANKING=off vs on ordering`. The doc now names eleven hook switches and explicitly points at Phase 1 for the twelfth, so the count reconciles without moving the assertion somewhere it doesn't belong.

Verification: `node --check` clean, `docs-truth` 88 files / no violations, and the partition fix proven in both directions rather than assumed.

alfhen:
## Dogfood executed — results

The protocol in this PR has now been run. Summary below; full record in `.wip/dogfood-oss-report.md`, instruments in `scripts/experiments/dogfood-oss/`.

**Corpus:** 9 repos / 9 languages, cloned outside the Lien tree, C1-asserted at pinned SHAs, all indexed (9/9, ≤2.2s each).

### One-sentence result

**The nudge delivery mechanism is well-built; what it delivers is sometimes wrong.** All ten hooks passed every plumbing check. Every HIGH finding is in the data layer.

### Hook trust ledger — 10/10 pass

| Property | Result |
|---|---|
| Output channel matches documented channel | 10/10 |
| Fails open (6–8 adverse scenarios each) | 10/10 |
| Env kill switches honored | 10/10 (11 switches) |
| Habituation guard / TTL / session dedup | verified where applicable |
| `session_id` path-traversal rejected | 6/6 that use it |
| Latency vs 5s timeout | 151–678 ms (worst = 13% of budget) |
| `recap-stop.sh` infinite-loop guard | **dual guard verified** |

The pre-run worry that a blocking Stop hook could loop is resolved.

### The headline: dependency attribution is wrong in both directions

Same subsystem, opposite errors, both with confident risk labels:

| | Query | Returns | Ground truth |
|---|---|---|---|
| False **zero** | gin `symbol:"StringToBytes"` | **0, `"low"`** | 11 grep-verified call sites |
| False **zero** | symfony `symbol:"__construct"` | **0, `"low"`** | 3 `new Cursor(...)` sites |
| False **graph** | `src/Command/Command.php` (**path does not exist**) | **93, `"critical"`** | inherits real `Command/Command.php` graph via basename collision |
| Correct | same files, no `symbol` / class name / non-colliding fake path | correct | — |

`get_dependents({filepath, symbol})` is marked **REQUIRED** in `instructions.ts` before any signature change, and `tools.ts` recommends `symbol` for "precise usage tracking" — so the mandated call is the one returning "no callers, low risk" on code with eleven callers. `riskLevel: "low"` is not a missing warning; it is an affirmative green light. It propagates: `api-delta-write.sh` uses the same symbol-level path.

The basename-collision case looks accidentally right (it resolves to the real file) but means the tool cannot distinguish paths — in any repo with several `Command.php` files it will confidently attribute the wrong file's callers. It also collides with #927: that fix adds a "not indexed" note which will now sit beside `93 dependents, risk: critical` on the same call.

### Shipped during the run

| PR | What | Verification |
|---|---|---|
| **#926** | PHP + Java test associations | I toggled it in/out on identical corpora: PHP 0→48→0%, Java 8→20→8%. Causal, not coincidence. |
| **#927** | Unindexed paths distinguished from empty results (3 handlers) | verbatim before/after on real symfony/console |
| in flight | false-zero symbol attribution | — |

⚠️ #926 and #927 both touch `get-files-context.ts` — whichever merges second needs a rebase.

### Other HIGH findings

- `list_functions` silently degrades an **invalid or catastrophic-backtracking regex to no filter at all** — byte-identical response to omitting `pattern`. The ReDoS guard works; failing open *silently* is the defect.
- `list_functions` stamps every result `score: 0, relevance: "not_relevant"` in symbols mode (both fields undocumented for this tool). An agent filtering on `relevance` discards every correct answer.
- `get_complexity` and `get_files_context` use **different default thresholds (30 vs 15)** for the identical metric on the identical function; `get_complexity`'s own schema says 15 while it uses 30.
- `get_complexity`'s `violations[]` is silently capped at `top` while `summary.violationCount` is the uncapped total (286 vs 10 observed).
- Batch `get_files_context` truncation guts the **first** files in the batch with no per-file indication.
- Response metadata materially under-documented across all six tools (+ an undocumented `indexInfo` on every response).

### A finding I retracted

I filed Swift test-associations as 100% broken. **Wrong** — my sampler matched `/Test coverage:/i` and `annotate` emits `"Test coverage inferred from symbol usage…"` with no colon, so a working feature scored zero. Re-measured: 20%; #923 works.

It was caught because the fix agent assigned to it **refused to fix a non-bug and returned evidence instead**. Worth preserving as a norm — an agent that produces a change to match a wrong premise converts a measurement error into a shipped regression.

### Coverage gaps (stated, not hidden)

- **Behavioral trials are the thinnest part.** 5 tasks sited, 1 completed at time of writing. Alamofire T1 was never sited because `get_dependents` returns 0 for every sampled file there — the author recorded it in `tasks.json`'s `omitted[]` rather than fabricate a fixture.
- Phase 4 ran ad hoc rather than as a stage: every HIGH claim was independently re-verified against grep ground truth before recording, and one was retracted.
- Cold npx latency unmeasured (clearing the npx cache is known-fragile on this machine).
- retrofit/Java and console/PHP not reached by the discovery-tool sweep.

### Protocol amendments earned the hard way

Three corrections to the plan in this PR, each found by running it:

1. **PostToolUse `additionalContext` never appears in a headless transcript** — only `SessionStart` events do. M1 must come from `nudge-events.jsonl`, with claim text reconstructed at the pinned SHA. Measuring from the transcript would have produced a confident, entirely false "no nudge fired" for every trial.
2. **A corpus under `~/.claude/` cannot be edited** — Claude Code treats it as sensitive. One trial completed 38 turns of investigation and then had its `Edit` denied.
3. **Match every output shape the consumer surface emits**, or you manufacture findings (see the retraction).

alfhen:
## Phase 3 behavioral trial — hono T1 (n=1)

Single-arm observation. **No lift claim** (protocol §4).

**Setup.** honojs/hono @ `224d2f5c`, fresh headless `claude --model sonnet`, plugin wired from the shipped `hooks.json`, `lien` shimmed to this build, MCP server enabled against the clone. C1/C2/C4 asserted clean. 13.8 min.

**The agent did the task properly** — 7 files, 91 insertions, including the target `src/utils/url.ts` and new test cases.

| Metric | Result |
|---|---|
| `annotate` nudge fired | **yes, ×14** |
| `blast` nudge fired | no — **correctly** |
| Acted on (M3) | **no — zero Lien tool calls** |

### The blast nudge was right to stay silent

The agent solved the task without touching the signature:

```
OLD: export const checkOptionalParameter = (path: string): string[] | null => {
NEW: export const checkOptionalParameter = (path: string): string[] | null => {
```

Byte-identical; it rewrote the body. So T1 never exercised the blast nudge.

That's a **task-design limitation, not a product finding.** The headroom precondition proved the hook *fires* on a signature change to that symbol — it could not prove the agent would *make* one. Worth fixing in the task design before anyone concludes anything about blast behavior in-session.

### The M3 result is the interesting one

The agent:

1. received server instructions marking `get_files_context` **REQUIRED before Edit/Write on any file**;
2. found Lien's MCP tools were **deferred**, not in its tool list;
3. ran `ToolSearch({query: "select:mcp__lien__get_files_context"})` — it was actively trying to comply;
4. loaded the tool successfully;
5. **never called it**, or any other Lien tool;
6. edited 7 files, having received 14 Read-time impact annotations along the way.

Zero `mcp__lien__*` invocations across 68 tool calls (23 Bash, 23 Read, 12 Grep, 8 Edit, 1 Write, 1 ToolSearch).

The deferred-tool interaction deserves attention on its own: the plugin's premise is that an agent gets these tools ambiently, and here they sat behind a discovery step. The agent paid that cost and still didn't convert.

### How much this supports

n=1, single arm, one language, one task shape. Enough to say *this* agent didn't use the tools despite the mandate and 14 annotations. **Not** enough for a general conversion rate, and it says nothing about the counterfactual — which is exactly what the paired-arm A/B design (deliberately out of scope here) exists to answer.

alfhen:
## Phase 3 trial 2 — hono T2 (test-verify). Mixed, and the mix is instructive.

3.3 min, C1/C2/C4 clean. The agent edited `jws.ts`, added 50 lines to `jwt.test.ts`, and ran `npx vitest run src/utils/jwt/jwt.test.ts` twice.

**Three things worked, and deserve equal billing with the failures:**

- `test-run-note.sh` observed and recorded the real test runs (4 `test_run` signals) — the acted-on side of the funnel works against real commands.
- `recap-stop.sh` correctly stayed **silent** at Stop: the ledger re-raises only *unresolved* risk, and the tests had been run. That restraint is what makes a Stop-time nudge tolerable rather than nagging.
- So `did-you-run-tests: shown 0, acted-on 0` is an honest row, not a gap.

**One thing went badly wrong → #929.** The association list for `jws.ts` names seven tests and **omits `jwt.test.ts`**, which directly imports it (`import { signing } from './jws'`) and references its symbols 39 times. The reminder therefore says:

> `Lien: you changed src/utils/jwt/jws.ts — associated tests: runtime-tests/node/index.test.ts, src/middleware/jwk/index.test.ts (+5 more). Run them before completing.`

An agent that obeys runs seven tests, misses the covering one, sees green, and concludes it verified the change.

**The agent ignored the nudge, found the right test itself, and was correct.** Following the nudge would have made its behavior *worse*. This is TypeScript — the healthiest language in the corpus at 88% — so it isn't a PHP/Java/Swift edge case.

Same caveat as T1: n=1, single arm, not a rate.

alfhen:
## Phase 3 trial 3 — serilog T2 (C#). Full chain fired; content was fabricated.

2.7 min, C1/C2/C4 clean. The agent edited `PropertyToken.cs`, `Alignment.cs` and `MessageTemplateParserTests.cs`, built, and ran `dotnet test` several times.

**The only trial where the Stop recap fired**, so the whole chain — annotate → edit → test-run observation → Stop recap — ran end to end in one session: `annotate` shown ×2, `test-verify` shown ×1, `test_run` signals ×4. The plumbing worked.

**The content did not → #930.** Verbatim, after editing `Alignment.cs`:

> `• 3 files import this — src/Serilog/GlobalUsings.cs, test/Serilog.Tests/GlobalUsings.cs, test/Serilog.PerformanceTests/GlobalUsings.cs; risk: medium (3 callers, 1 untested).`

Real users of `Alignment`: `MessageTemplateTextFormatter.cs`, `MessageTemplateRenderer.cs`, `Padding.cs`, `PropertyToken.cs`, `MessageTemplateParser.cs`.

**5 real, 3 reported, zero overlap.** Every reported dependent is wrong and every real one is missing; `risk: medium (3 callers)` is derived entirely from boilerplate.

Mechanism: `global using Serilog.Parsing;` is resolved as importing every file in that namespace. The same convention that manufactures the false edges hides the real ones — with a global using in place, per-file imports don't appear, so type-level usage is never counted. Affects every .NET 6+ repo.

**Instrument limitation found:** the recap's claim text can't be reconstructed — `test-verify` events carry no file (the recap is session-scoped), so my `annotate`-replay returns empty. With hook output also absent from transcripts, the recap is currently the one nudge whose wording cannot be recovered post hoc. A future run auditing recap truthfulness must capture `lien recap --session <id>` live.

alfhen:
**Correction to an earlier note.** I twice flagged that #926 and #927 would conflict on `get-files-context.ts` and that the second to merge would need a rebase. Having checked the actual hunks, that was too pessimistic:

```
#926: @@ -11,10   (imports)      @@ -302,6 / -316,7 / -331,12   (findTestAssociations)
#927: @@ -1,6     (imports)      @@ -662,6 / -696,7            (handleGetFilesContext)
```

Different functions, non-overlapping ranges. GitHub currently reports both `MERGEABLE / CLEAN`.

The only adjacency is the two import blocks near the top of the file, which git normally merges without help. So: merge in either order, and just re-check the second one's status afterwards rather than assuming a rebase is required.

Both files being touched by two PRs was the right thing to notice; concluding "conflict" from that alone was not.

alfhen:
## gin T3 — the redesigned trial. The harm pathway is now *observed*, not inferred.

T1's shape couldn't trigger the nudge it targeted: on hono the agent rewrote the function body and left the exported signature byte-identical, so nothing fired. A bare rename makes the change unavoidable. The prompt deliberately does **not** mention updating usages — instructing that would manufacture the behavior under measurement:

> "The helper StringToBytes in internal/bytesconv reads backwards relative to the naming used elsewhere in that package. Rename it to FromString."

gin @ `34dac209`, 1.5 min, C1/C2/C4 clean. **blast ×3, annotate ×5, test-verify ×1** — the full chain fired.

### What happened, in order

1. Nudge received:
   `⚠ lien: exported symbol removed — StringToBytes (0 dependents, risk low). Callers will break — check get_dependents.`
   *(self-contradictory: warns callers will break while asserting there are none)*
2. **The agent obeyed it** — called `get_dependents({filepath:"internal/bytesconv/bytesconv.go", symbol:"StringToBytes"})`.
3. Lien answered: `dependentCount: 0`, `riskLevel: "low"`, `totalUsageCount: 0`, `riskReasoning: []`. Ground truth: **11 call sites across 4 files.**
4. The agent then ran `grep -rn "StringToBytes" --include=*.go .`
5. And said: *"Now I'll rename `StringToBytes` to `FromString` in all these locations"* — **"these locations" came from grep, not Lien.**
6. It correctly updated `auth.go`, `binding/form_mapping.go`, `render/json.go`, `render/text.go` + tests, and ran `go build`.

### What this establishes

- A real agent, in a real session, **did** make the call the docs mandate.
- That call **did** return an affirmative false all-clear.
- The correct outcome came **entirely from grep**.
- **Lien contributed nothing to the right answer, and its one contribution was wrong.**

The harm didn't land — but only because the agent distrusted the tool and cross-checked. That's the uncomfortable part: Lien's pitch is that it beats grep for precisely this question. Here grep beat Lien, and an agent that had *followed* Lien, as `instructions.ts` instructs, would have renamed a function with eleven callers and shipped a broken build.

It also sharpens the earlier trials. "Agents ignore the nudges" was the wrong reading. **This agent obeyed — and obedience is what exposed the bug.**

### Secondary: blast fires on test-symbol renames

Two of three blast events were for `TestStringToBytes` in `bytesconv_test.go`. Nothing can depend on a Go test function, so that's pure noise — and noise is how a nudge becomes wallpaper. Minor beside the false zero, but worth folding into the next pass on the blast nudge.

### Fix

#931 addresses exactly this, verified on this same symbol: `0/low` → `4 dependents, totalUsageCount 11, risk high`, and the nudge text goes from `(0 dependents, 0 untested, risk low)` to `(4 dependents, 1 untested, risk high)`. Re-running this trial post-merge is the natural verification.

alfhen:
## Inline review triage — all four addressed

**1 & 2 (workflow.js:389, :409)** — CodeRabbit's original findings, both fixed earlier in `e1fd49bc`. The slice-misalignment one was real and I reproduced it before fixing (a dead tool agent slid a hook report into the tool partition). The `LIEN_STRUCTURAL_RANKING` one I fixed in the opposite direction to the suggestion — the switch gates `search_code` ranking, not a hook, so the *doc* was wrong to file it under hook plumbing; Phase 1's edge matrix already covers it.

**3 (`.changeset/swift-symbol-usage-association.md`) — not from this PR.** It arrived via the `origin/main` merge (`git diff --name-only origin/main...HEAD | grep changeset` → 0). It's #923's file and already merged.

That said, the finding looks correct and worth someone picking up: the changeset says *"Three gates keep this precise"* while the implementation has four (`isUnambiguousIdentifierShape`, `isMultiSegmentIdentifier`, unique non-self non-extension definition, and ≥1 `isTypeShapedIdentifier` driver per edge). Cosmetic, but changesets become public release notes, so the count being wrong is the kind of thing that outlives the PR. Not fixing it here to keep this diff to one concern.

**4 (workflow.js:61) — real, mine, and worse than reported.** Fixed in the commit above.

The reviewer flagged that the hardcoded `/Users/alfhenderson/.claude/jobs/460173c9/tmp/corpus` fallback is machine-specific and contradicts both the protocol's recommended neutral root and the committed manifest. All true. What the review couldn't know is that the path was *also already stale*: that location had to be abandoned mid-run because Claude Code treats everything under `~/.claude/` as a sensitive file and **denies edits there** — it killed a behavioural trial after a full 38-turn investigation.

So the fallback wasn't just unportable, it pointed at a known-unusable directory.

Checking it exposed the same flaw one level down: `provision.mjs` still **defaulted** to `$CLAUDE_JOB_DIR/tmp/corpus`, i.e. the same trap. Fixed the root cause rather than the symptom — both files now resolve `LIEN_DOGFOOD_CORPUS` with a neutral `/tmp` fallback, so they can't disagree about where the corpus lives and neither steers a future run back into it. The reason is recorded in a comment at both sites so it doesn't get "cleaned up" later.

Gates after the fix: `lien delta --base` exit 0, `docs-truth` 89 files / no violations.

Issue #902, opened 2026-07-25
Go's dominant same-package test convention (foo.go / foo_test.go, no import) leaves 65% of tested source files with zero test-association signal (cli/cli)


alfhen:
## Summary

Idiomatic Go unit tests are co-located in the same directory and the same package as the code they test (`package foo` in both `foo.go` and `foo_test.go`) specifically so the test can reach unexported identifiers — there is no `import` statement connecting them at all, because none is needed or even legal (you can't import your own package). Lien's test-association engine (`findTestAssociationsFromChunks` / `matchesFile`) works entirely off `chunk.metadata.imports`, so it is structurally blind to this convention: a same-package test file leaves no import-based trace pointing back at the file(s) it tests.

This exact mechanism was already spotted and named in #867's Go repro ("*same-package-blind-spot cases aside — see the separate, structural 'Go has no import statement for same-package tests' observation, which is not fixable the same way and is not part of this issue*") but was never broken out into its own tracked issue. Filing it now, quantified against a second, larger, differently-shaped Go repo (cli/cli vs. #867's gin), because the scale is bigger than a footnote suggests.

## Repro + quantification (github/cli, i.e. `cli/cli`)

```
git clone --depth 1 https://github.com/cli/cli.git
cd cli
lien init --editor claude-code --legacy && lien index -f
```

Prevalence of the convention itself: **355 of 356** (`99.7%`) `*_test.go` files in the repo declare the *same* `package` name as their non-test siblings in the same directory (the one exception is a genuine external-package integration test).

Impact on Lien's own test-association output (via the batch `findTestAssociationsFromChunks` path over the full `scanAll()` result): of the **457** non-test `.go` files that have at least one same-directory `*_test.go` sibling, **297 (65.0%)** get **zero** test associations from Lien — despite every one of them having real, same-package unit test coverage sitting right next to it on disk. Concrete examples (each has a real, substantial sibling test file covering it): `pkg/cmd/label/list.go` (-> `list_test.go`, package `label`), `pkg/cmd/run/run.go`, `pkg/cmd/codespace/list.go`, `context/remote.go`, `cmd/gen-docs/main.go`.

[...]

Issue #894, opened 2026-07-25
lien annotate resolves a different project root than lien index in nested-git / monorepo-subdirectory setups


alfhen:
## Symptom

`lien annotate` and `lien index`, run from the SAME working directory, can resolve DIFFERENT project roots when that directory is a git repository nested inside another git repository's tree (or a repo-less subdirectory of a monorepo checkout). `annotate` walks up to the nearest `.git` to find the project root; if the cwd's subtree has no `.git` of its own, it lands on the outer repository's root — which has no index — while `lien index` (run earlier with an explicit path or different resolution) indexed the inner project. Result: `annotate` silently reports against an unindexed root instead of the index the user just built.

## How it was found

During the #864 dogfood (PR #893): running `lien annotate` inside `assertj-core/` (a subdirectory of the assertj monorepo checkout, no own `.git`) picked the outer root while the index had been built for `assertj-core`. Verified not to be a bug in the #864 change itself — direct `findTestAssociationsFromChunks` calls against the correctly-scoped index returned 1360+ correct associations.

## Shape of a fix (suggestion, unvalidated)

Root resolution should be consistent across commands — e.g. both resolve via the nearest initialized Lien project (config/index presence) before falling back to the nearest `.git`, or `annotate` should warn loudly when the resolved root has no index rather than proceeding silently.

## Severity

Edge case (git-repo-nested-in-git-tree / monorepo subdirectory), low priority — filed so the diagnosis isn't lost.

PR #923, opened 2026-07-28
feat(parser,cli): Swift symbol-usage test-association signal (#869 spike)

### LINKED ISSUE #869: Whole-module-import languages (Swift/XCTest) have no per-file test-association signal — structural gap, not a matching bug
## Summary

For languages where test files import their subject as a whole module rather than a specific per-symbol path (Swift's `import Alamofire` / `@testable import Alamofire`), `chunk.metadata.imports` for every test file in the module is the same single string — the module name — regardless of which specific source file it actually tests. There is no per-file signal left to match against, so test-association is structurally unable to distinguish "this test covers `Session.swift`" from "this test covers `HTTPHeaders.swift`" from raw import data alone. This is a design gap rather than a simple matching bug (see #868 for the separate, fixable false-positive matching bug this same investigation surfaced), and is filed for awareness/design discussion, following the precedent of #863/#864 being filed as judgment calls rather than mechanical fixes.

## Repro (Alamofire/Alamofire)

Real import blocks from 5 different Tests/*.swift files, each testing a different Source/ file:
```
Tests/SessionTests.swift:       @testable import Alamofire, import Foundation, import XCTest
Tests/RequestTests.swift:       import Alamofire, import Foundation, import Testing, import XCTest
Tests/HTTPHeadersTests.swift:   import Alamofire, import XCTest
Tests/AFError+AlamofireTests.swift: import Alamofire, import Foundation
Tests/RequestInterceptorTests.swift: @testable import Alamofire, import Foundation, import Testing
```
Every one imports the bare module `Alamofire` — never anything more specific. `SwiftImportExtractor` correctly extracts exactly this (verified via sqlite `imports` column); there's nothing wrong with the extraction. But `matchesFile('Alamofire', 'Source/Core/Session')` cannot succeed under any of the 5 existing strategies — there's no shared path component between a bare module name and a specific file path at all, by construction. Confirmed via `lien annotate`: every substantive file with real, heavy XCTest coverage (`Session.swift`, `Request.swift`, `HTTPHeaders.swift`, `AFError.swift`, `RequestInterceptor.swift`) reports `• No test coverage.`

`isTestFile()`'s Swift-specific handling correctly classifies which files *are* tests (that part works) — the gap is entirely in the second step, matching a specific test file to the specific source file(s) it exercises.

## Why this is a design question, not a one-line fix

Unlike #867 (manifest-aware root mapping) or #868 (false-positive boundary matching), there's no missing piece of information Lien could read off disk to solve this — Swift's `import` statement genuinely does not encode per-file granularity. Two directions worth considering, both with tradeoffs:
1. **Content-based heuristic**: pair `FooTests.swift` <-> `Foo.swift` by name proximity (strip `Tests` suffix, look for a same-named file under the paired source root). Works for the common 1:1 naming convention (`SessionTests.swift` <-> `Session.swift`) but doesn't help integration/suite-style test files that legitimately cover many source files, and risks false confidence of its own if the naming convention isn't followed.
2. **Leave as an honest gap**: rather than a misleading `No test coverage.` (which reads as "this is untested," when the true state is "test coverage is real but unknown to Lien"), consider distinguishing "no coverage found" from "coverage undeterminable for this language's import style" so the annotation doesn't actively mislead.

## Impact

This is a real, structural feature gap for a whole family of languages/frameworks that import by module rather than by path (Swift/XCTest confirmed; likely also affects any language/testing-framework combination with the same whole-module-import convention). Concretely: on Alamofire, this produced not just silence but two confirmed *misleading* signals as a side effect (tracked as part of #868's evidence) — a 43-line stub file (`Source/Alamofire.swift`) became a false "hub" matching all 41 test files' bare module import, and `Combine.swift` appeared to have correct, precise test coverage but only because of a coincidental name collision with Apple's unrelated `Combine` framework import. A Swift developer trusting `annotate`'s test-coverage line today would be actively misled toward false confidence on exactly the files that most need the safety net (the ones with the most real test coverage), not just left without a signal.

alfhen:
## Correction (adversarial re-verification, after initial "ship" claim)

**The original version of this PR claimed 100% precision on all three calibration repos. That claim was wrong and is retracted.** An adversarial reviewer reproduced the exact 46-edge Alamofire set and found a real false positive; re-verifying every edge in all three repos by opening the actual call sites (not just re-confirming declaration uniqueness) found **10 confirmed false positives total**, enough to fail the ship gate outright on `swift-composable-architecture` (41/49 = 83.7%, well under the ~95% bar).

**Root cause of the bad original evidence:** the mechanism verified that a driving symbol's *definition* was unique in-project, but never verified that a test's actual *call site* resolves to that definition. Swift lets a bare method name collide with something the indexer never sees at all:

- **Stdlib protocol witness** — Alamofire's `Tests/TestHelpers.swift:505` calls `decoder.singleValueContainer()` inside `extension HTTPHeaders: Decodable`. That's the Swift stdlib `Decoder` protocol's own requirement, satisfied by every conforming type in existence — not Alamofire's differently-typed `URLEncodedFormEncoder.singleValueContainer()`, which merely happened to be the only *project-indexed* declaration of that bare name.
- **Stdlib type's own extension overload** — `swift-composable-architecture`'s `Effect.swift` adds a `name:`-taking `TaskGroup.addTask` shim; every real test call site (`group.addTask { ... }`, no `name:` argument) resolves to the plain stdlib `TaskGroup.addTask(priority:operation:)` instead.
- **External package's free function** — the same repo's bare `withDependencies { ... }` calls (4 test files) resolve to the separate `swift-dependencies` package's own top-level function — a `Package.swift` dependency, never part of the indexed project — not `TestStore.withDependencies`.
- A second Alamofire edge (`asURL`, `Tests/TestHelpers.swift`) was borderline/incidentally-mixed rather than cleanly correct and is counted as a defect here for the same reason: the mechanism can't distinguish it from the confirmed-bad cases without receiver-type information it doesn't have.

**Fix shipped:** a new `isTypeShapedIdentifier` gate — an edge is kept only when at least one of its driving symbols is leading-uppercase (allowing Swift's `_`-prefixed SPI convention, e.g. `_CancelID`) and multi-segment. This is a structural, method-name-agnostic shape check, not a hardcoded list of risky method names. Verified to eliminate all 10 confirmed false positives.

**Honest precision, before and after the fix (every edge in all three repos, independently re-verified by opening real call sites):**

| Repo | Pre-fix edges | Pre-fix honest precision | Post-fix edges | Post-fix precision |
|---|---|---|---|---|
| Alamofire/Alamofire | 46 | 44/46 = **95.7%** (2 confirmed FPs) | 26 | **26/26 = 100%** |
| vapor/vapor | 57 | 57/57 = **100%** (0 FPs found) | 25 | **25/25 = 100%** |
| pointfreeco/swift-composable-architecture | 49 | 41/49 = **83.7%** (8 confirmed FPs) | 29 | **29/29 = 100%** |

**The cost of the fix, stated plainly:** roughly half of all previously-good edges are lost project-wide (62 of 142 confirmed-good pre-fix edges, 44%). Most notably, **every edge to Alamofire's own `Request.swift` — the file that originally motivated this whole investigation — is now lost.** `Request` is single-segment (fails `isMultiSegmentIdentifier`, same shape as `Session`/`Get`/`Run`) and can never itself be a type-shaped driver, and none of its distinctively-named methods (`prepareForRetry`, `cURLDescription`, `onURLRequestCreation`, ...) have a type-shaped symbol alongside them in the tests that call them. `vapor` loses 32 confirmed-good edges (its `withApp` test-harness hub, `guardMiddleware`, `hexEncodedString`, etc. — all independently verified correct) for **zero** precision benefit, since vapor had no defects to fix in the first place.

The ship gate (per the original #869 design comment) is a precision-only bar, and the fix brings all three repos to a verified, re-checked 100%. But the practical value of the shipped mechanism is now meaningfully smaller than originally presented — this is flagged prominently, not glossed over, so it can be weighed on its merits rather than on the retracted numbers below.

---

## Summary

Implements the measure-gated Swift symbol-usage association spike from #869's design comment: a deterministic, zero-LLM signal that recovers real (if lower-confidence) test-association coverage for Swift/XCTest files, where whole-module imports (`import Alamofire`) carry zero per-file signal today.

- New `packages/parser/src/swift-symbol-usage-signals.ts` (mirrors `stale-literal-signals.ts`'s template): associates test `T` with source `S` iff `T`'s `callSites` reference a symbol `X` that (a) passes **both** the shipped `isUnambiguousIdentifierShape` **and** a new, stricter `isMultiSegmentIdentifier` (≥2 camel/underscore segments — required because the shipped gate alone passes every single Capitalized word like `Get`/`Run`/`Session` trivially), (b) is defined in **exactly one** non-test Swift file project-wide, (c) that file isn't `T` itself, and — added post-correction — (d) the edge has **at least one type-shaped (leading-uppercase, multi-segment) driving symbol** (see the Correction above for why).
- Definition-side exclusion of `extension <ForeignType>` declarations — a type extended but never actually *declared* in-project (Foundation/stdlib types like `HTTPURLResponse`) — unless the type also has a real, non-extension declaration somewhere in the project. No hardcoded denylist: "every declaration of X in this corpus is an extension" already *is* the foreign-type test.
- Surfaced as a **distinct fourth label tier** in `lien annotate` — "inferred from symbol usage" — never merged into the confident import-based `tests` set, mirroring #902's Go same-directory tier-2 discipline exactly. Deliberately kept out of `get_files_context`, `@liendev/review`'s gap detection, and `verify-tests`'s ledger/scope-matching (see "Scope call" below).

## Ship-gate: calibration on 3 real Swift repos of different shapes

See the Correction section above for the full, honest before/after precision tables and methodology (independent re-verification of every edge by opening the actual call sites — exhaustive, since every repo's edge count is under 60 both before and after the fix). All three repos clear the ~95% bar **after** the `isTypeShapedIdentifier` fix; `swift-composable-architecture` did not clear it before.

### Named targets (Alamofire, the repo #869's original investigation used)

- **`Source/Core/HTTPHeaders.swift`** ← `Tests/HTTPHeadersTests.swift`, via `HTTPHeaders`, `HTTPHeader` — the clean 1:1 case, recovered structurally, **still recovers after the fix** (both driving symbols are type-shaped).
- **`Source/Core/Request.swift`** — the file that originally motivated #869's investigation. Recovered pre-fix via `cURLDescription`/`onURLRequestCreation`/`onURLSessionTaskCreation`/`eventMonitor`/`prepareForRetry`/`cacheResponse`; **no longer recovers post-fix** — see the Correction above for why (no type-shaped driver is structurally possible for this file). Reverts to the honest "not determinable" label, same as before this feature shipped.
- **`Source/Core/AFError.swift`** — genuinely has real test coverage (`Tests/AFError+AlamofireTests.swift` exists), but shows the honest "not determinable" label both before and after this PR, and before/after the fix. Known limitation: `AFError`'s failure-reason types are nested inside `extension AFError { enum … }`, so their declaration name is captured as the dotted compound `AFError.MultipartEncodingFailureReason` while a callsite only ever records the bare last segment — the two never match. Conservative miss (recall), not a precision defect.

### `lien annotate` before/after (verbatim, captured against the final, corrected code)

**HTTPHeaders.swift** (Alamofire, the clean 1:1 case — still recovers):
```
# before (main branch build, no symbol-usage signal at all)
Lien impact for Source/Core/HTTPHeaders.swift:
  • Test coverage not determinable from imports (whole-module import).

# after (this branch, post-correction)
Lien impact for Source/Core/HTTPHeaders.swift:
  • Test coverage inferred from symbol usage (not import-verified): Tests/HTTPHeadersTests.swift.
```

**Request.swift** (Alamofire — recovered by the original spike, lost by the correction):
```
# before this PR, AND after the isTypeShapedIdentifier fix (identical — the honest label)
Lien impact for Source/Core/Request.swift:
  • Test coverage not determinable from imports (whole-module import).
  • Max cyclomatic complexity: 15 (1 function over warn threshold).
```

**URLEncodedFormEncoder.swift** (Alamofire — the confirmed `singleValueContainer` false positive, now cleaned up):
```
# after the fix — only the genuine ParameterEncoderTests.swift edge remains;
# the spurious TestHelpers.swift/singleValueContainer edge is gone
Lien impact for Source/Features/URLEncodedFormEncoder.swift:
  • Test coverage inferred from symbol usage (not import-verified): Tests/ParameterEncoderTests.swift.
```

**AFError.swift** (Alamofire, genuinely-untested-by-this-heuristic example — real tests exist, honest label unaffected by anything in this PR):
```
Lien impact for Source/Core/AFError.swift:
  • Test coverage not determinable from imports (whole-module import).
```

**`lien annotate --tests-only` (the real post-edit hook path)**, HTTPHeaders.swift:
```
Lien: you changed Source/Core/HTTPHeaders.swift — no import-verified test match, but symbol usage suggests: Tests/HTTPHeadersTests.swift (inferred, not import-verified). Consider running them before completing.
```

### Known, accepted characteristics (not precision defects — distinct from the false positives above)

- **"Hub" files**: a shared test-harness helper genuinely called by nearly every test file (vapor's `withApp`, TCA's `TestStore`) produces a high-fan-out association. Every one of *these* edges is correct (the test really does call the uniquely-defined, type-independent helper) but low-specificity — reviewer-attack-vector #3 from the design comment, an accepted characteristic. (Several of vapor's `withApp` edges were lost to the `isTypeShapedIdentifier` fix regardless, since `withApp` alone has no type-shaped co-driver.)
- **Protocol-method collisions the design comment specifically flagged as an attack vector** (`singleValueContainer`, `encodeNil`, etc. — reviewer-attack-vector #2) DID materialize as real false positives, contrary to the original PR's claim. This is now corrected above.

## Scope call: `get_files_context`

Per the design's explicit permission to skip if not "trivial": **not wired into `get_files_context`**. That handler (`packages/cli/src/mcp/handlers/get-files-context.ts`) maintains its own fully separate re-implementation of test-association matching against a different chunk shape (`SearchResult`, not `CodeChunk`), and two of its neighboring functions already sit at/over complexity threshold (`computeComplexityHeadroom` cognitive 25/15, `buildMetadata`-adjacent risk in a 4-importer file). Duplicating a second signal computation into that hot path for a spike whose primary deliverable is the `lien annotate` surface isn't trivial, and is deferred to a follow-up if this proves out further. Wired fully into `lien annotate` (both the full annotation and the `--tests-only` post-edit reminder).

## Gates

- `npm run format:check` — pass
- `npm run lint` — 0 errors (38 pre-existing warnings, untouched)
- `npm run typecheck` — pass
- `npm run build` — pass
- `npm test` — pass (parser 1387, core+review 378, cli 1253, action 41 — all green)
- `lien delta` — exit 0, no NEW over-threshold crossings

Changeset: `@liendev/parser` + `@liendev/lien` patch (updated to describe the final, corrected shipped behavior — this PR has not been released, so the changeset describes the end state directly rather than layering a second entry).

## Rebase note

Branched from `origin/main` at the time origin/main included #919 (`resolveProjectRoot` via nearest completed index) and #922 (path-matching boundary fix), both landing after this work started. Rebased on top and manually resolved the `FileTestAssociations`/`scanTestAssociations`/`runTestsOnly` three-way conflict with #894's `indexMissing`/`formatNoIndexWarning` work (both features' fields now coexist cleanly) — verified with a full rebuild + test run + `lien delta` post-resolution.

Closes #869.

---

### Lien Review

> [!WARNING]
>
> This PR adds Swift symbol-usage fallback detection for source changes that lack direct test coverage. It introduces identifier-shape heuristics to recognize type-like symbols, but the implementation mishandles leading-underscore SPI/implementation-detail prefixes, causing `_CancelID`-style identifiers to be rejected despite tests and docs claiming they are accepted. The bug reduces recall for Swift internal types and would fail the newly added test.
>
> - Adds Swift extension and symbol-usage signal detection
> - Adds identifier-shape helpers (`isTypeShapedIdentifier`, `isMultiSegmentIdentifier`)
> - Mishandles `_`-prefixed SPI identifiers in `splitIdentifierSegments`

<sup>Reviewed by [Lien Review](https://lien.dev) for commit cdbe964. Updates automatically on new commits.</sup>

### DIFF
--- a/.changeset/swift-symbol-usage-association.md
+++ b/.changeset/swift-symbol-usage-association.md
@@ -1,0 +1,35 @@
+---
+"@liendev/parser": patch
+"@liendev/lien": patch
+---
+
+#869: a measure-gated spike recovering non-import test-association signal
+for Swift/XCTest, where whole-module imports (`import Alamofire`) carry zero
+per-file information (see the existing "not determinable" honesty label).
+
+New deterministic, zero-LLM signal (`packages/parser/src/swift-symbol-usage-signals.ts`,
+mirroring `stale-literal-signals.ts`'s template): a test chunk's own
+`callSites` versus which single source file uniquely defines the referenced
+symbol. Two gates keep this precise:
+
+- A new, stricter `isMultiSegmentIdentifier` helper (>= 2 camelCase/
+  underscore segments) — the shipped `isUnambiguousIdentifierShape` (docRefs'
+  gate) passes every single Capitalized word trivially (`Get`, `Run`,
+  `Session`, `Client`), so it's insufficient as a collision-resistance gate
+  on its own. Both gates apply together; `isUnambiguousIdentifierShape`
+  itself is untouched.
+- `extension <ForeignType>` declarations are excluded from the definition
+  side unless the type also has a real, non-extension declaration
+  in-project — the one false-positive shape measured on Alamofire (a file
+  merely extending a Foundation type like `HTTPURLResponse` otherwise looks
+  like it "defines" that type to every test that references it).
+
+Calibrated on Alamofire/Alamofire plus two additional real Swift repos of
+different shapes; see the #869 PR for the full precision tables.
+
+Surfaced as a DISTINCT third label tier in `lien annotate` — "inferred from
+symbol usage", never merged into the confident import-based association —
+mirroring #902's Go same-directory tier-2 discipline. Deliberately kept out
+of `get_files_context`, `@liendev/review`'s gap detection, and
+`verify-tests`'s ledger/scope-matching, same conservative call as that
+precedent.
--- a/.changeset/swift-symbol-usage-association.md
+++ b/.changeset/swift-symbol-usage-association.md
@@ -10,7 +10,7 @@ per-file information (see the existing "not determinable" honesty label).
 New deterministic, zero-LLM signal (`packages/parser/src/swift-symbol-usage-signals.ts`,
 mirroring `stale-literal-signals.ts`'s template): a test chunk's own
 `callSites` versus which single source file uniquely defines the referenced
-symbol. Two gates keep this precise:
+symbol. Three gates keep this precise:
 
 - A new, stricter `isMultiSegmentIdentifier` helper (>= 2 camelCase/
   underscore segments) — the shipped `isUnambiguousIdentifierShape` (docRefs'
@@ -20,12 +20,25 @@ symbol. Two gates keep this precise:
   itself is untouched.
 - `extension <ForeignType>` declarations are excluded from the definition
   side unless the type also has a real, non-extension declaration
-  in-project — the one false-positive shape measured on Alamofire (a file
+  in-project — one false-positive shape measured on Alamofire (a file
   merely extending a Foundation type like `HTTPURLResponse` otherwise looks
   like it "defines" that type to every test that references it).
+- `isTypeShapedIdentifier`: an edge needs at least one leading-uppercase,
+  multi-segment driving symbol, or it's demoted. Added after adversarial
+  re-verification (opening actual call sites, not just re-confirming
+  declaration uniqueness) found real false positives where a bare method
+  name collided with something the indexer can't see at all — a stdlib
+  protocol witness (`Decoder.singleValueContainer()`), a stdlib type's own
+  extension overload (`TaskGroup.addTask(name:...)`), or an external
+  package's free function (`swift-dependencies`' `withDependencies`). This
+  gate is necessary (one calibration repo failed precision without it) but
+  costly: roughly half of all previously-good edges are lost project-wide,
+  including every edge to Alamofire's `Request.swift` — see the #869 PR for
+  the full before/after precision tables.
 
 Calibrated on Alamofire/Alamofire plus two additional real Swift repos of
-different shapes; see the #869 PR for the full precision tables.
+different shapes (vapor/vapor, pointfreeco/swift-composable-architecture);
+see the #869 PR for the full precision tables.
 
 Surfaced as a DISTINCT third label tier in `lien annotate` — "inferred from
 symbol usage", never merged into the confident import-based association —
--- a/docs/architecture/test-association.md
+++ b/docs/architecture/test-association.md
@@ -86,6 +86,26 @@ Unlike the whole-module/enclosing-namespace gaps above, this one needs no heuris
 
 Both tiers are structurally bounded to the same directory (Go's own compiler guarantee) — the worst either can produce is "attributed to the wrong file in the same package," never the cross-directory textual collisions the bare-identifier guard above protects against.
 
+## Swift symbol-usage association: a measure-gated non-import signal
+
+Swift's whole-module-import gap (above) is a genuine structural blind spot — but the SQLite chunk store already carries a DIFFERENT signal that doesn't depend on imports at all: a test chunk's own `callSites` (the AST-extracted symbols it references) versus which single source file uniquely DEFINES that symbol. Measured against a real Alamofire/Alamofire clone ([#869](https://github.com/getlien/lien/issues/869)), this recovers real, spot-checked coverage for exactly the files the whole-module gap above leaves dark — including the named cases that motivated the investigation (`Request.swift`, `HTTPHeaders.swift`).
+
+`packages/parser/src/swift-symbol-usage-signals.ts` implements the rule: test `T` associates with source `S` iff `T`'s `callSites` reference a symbol `X` such that:
+
+1. `X` passes **both** the shipped `isUnambiguousIdentifierShape` (docRefs' gate) **and** this module's own, stricter `isMultiSegmentIdentifier` — at least 2 camelCase/PascalCase/underscore segments. The shipped gate alone is insufficient here: its case-transition check is satisfied by the first two characters of ANY Capitalized word (`Ge` in `Get`, `Se` in `Session`), so it passes `Get`/`Run`/`Map`/`Session`/`Client` just as readily as `TypeMap`/`HTTPHeaders` — fine for its original prose-vs-code job, useless as a collision-resistance gate. `isMultiSegmentIdentifier` is a separate, additional helper; the shipped gate itself is untouched.
+2. `X` is defined in **exactly one** non-test Swift file project-wide.
+3. That defining file is not `T` itself.
+
+The definition side excludes `extension <ForeignType> { ... }` declarations from counting as a definition, unless the type also has a real (non-extension) declaration somewhere in the project. This is the one false-positive shape measured on Alamofire: `extractSwiftClasses` treats `extension HTTPURLResponse { ... }` as *defining* `HTTPURLResponse`, so every test that happens to reference that Foundation type would otherwise falsely hub onto the extending file. Because a Foundation/stdlib type is never actually declared inside the project, "every declaration of X in this corpus is an extension" already IS the foreign-type test — no hardcoded list of framework type names required.
+
+This is a strictly ADDITIVE, lower-confidence THIRD tier, never merged into the confident import-based association above (or Go's tier 2) — mirroring that tier's discipline exactly:
+
+```text
+Test coverage inferred from symbol usage (not import-verified): Tests/HTTPHeadersTests.swift.
+```
+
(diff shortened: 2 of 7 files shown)


alfhen:
Triage of the out-of-diff review finding (splitIdentifierSegments / _CancelID): **dismissed as a refuted false positive**, verified empirically at head cdbe9647 by the adversarial re-review. The engine's premise misreads the implementation order: the camelCase-boundary regex runs *before* the underscore split, so `_CancelID` becomes `_Cancel ID` → split on `[_\s]+` → `['Cancel','ID']` (2 segments) — `isTypeShapedIdentifier('_CancelID')` and `('_EffectPublisher')` both return true when the real functions are run. Live proof: `lien annotate` on TCA's Cancellation.swift and Publisher.swift keeps both SPI-driven edges; if the finding were correct they would be demoted. The builder's underscore tests pass because the code genuinely behaves as documented. No code change needed.

PR #926, opened 2026-07-28
fix(parser,cli): PHP root-mapped PSR-4 + capitalized Tests/, Java same-package test-association

### LINKED ISSUE #925: test-association: PHP PSR-4 root mapping, capitalized Tests/, and Java same-package convention all miss mainstream repos
Found via the foreign-repo dogfood sweep (docs/development/dogfood-oss-corpus-protocol.md), measured via \`lien annotate\` (the same path the nudge hooks use) against real repos.

Symptoms (25-file sample per repo):

| Repo | Language | Measured coverage | Ground truth |
|---|---|---|---|
| symfony/console | PHP | 0% (0/25) | 100 \`*Test.php\` files under \`Tests/\` |
| square/retrofit | Java | 8% (2/25) | 101 \`*Test.java\` files, multi-module Gradle |

Root causes:

1. \`php-psr4.ts\`'s \`collectPsr4Entries\` skips a PSR-4 entry whose directory is the empty string (\`""\`) — a valid Composer convention meaning "map this namespace prefix directly onto the project root" (symfony/console's own \`composer.json\`: \`"Symfony\\\\Component\\\\Console\\\\": ""\`). With no \`src/\` subdirectory, the whole PSR-4 map ends up empty for this repo.
2. \`isTestFile\`'s generic directory-segment check (\`test|tests|spec|specs|__tests__\`) is case-sensitive. Symfony's convention capitalizes its test directory (\`Tests/\`), so every PHP test file in the corpus is excluded from test-chunk scanning before import-matching even runs.
3. Java (and Go) permit same-package access with no \`import\` statement at all. Go's #902 fix recovers this via "same directory" (compiler-enforced). Java has no such compiler guarantee, but the universal Maven/Gradle Standard Directory Layout (\`src/<sourceSet>/java/<package/path>/...\`) makes "same package-relative path after stripping the source-root prefix" an equally reliable, deterministic signal — unexploited today, so retrofit's cross-module same-package test files (\`retrofit-adapters/guava/src/test/java/...\` testing \`retrofit-adapters/guava/src/main/java/...\`) show zero coverage.

Fix branch: \`worktree-dogfood-oss-plan\` (fixes 1 and 2 in \`packages/parser/src/php-psr4.ts\` / \`packages/parser/src/utils/path-matching.ts\`; fix 3 adds a same-package-test-convention tier for Java, mirroring \`go-same-directory-tests.ts\`).

alfhen:
## Summary

Fixes #925 — found by the foreign-repo dogfood sweep, which measured Lien's
test-association layer (the thing `test-reminder.sh`/`test-run-note.sh`/
`recap-stop.sh` are built on) at near-zero coverage on real PHP and Java
repos despite large, conventionally-named test suites.

Two unrelated root causes, shipped as two commits:

1. **PHP** (`php-psr4.ts`, `path-matching.ts`): `collectPsr4Entries` treated
   an empty-string PSR-4 directory as invalid and skipped it — but that's a
   *valid* Composer convention meaning "map this namespace prefix directly
   onto the project root" (symfony/console's own composer.json:
   `"Symfony\\Component\\Console\\": ""`). Separately, `isTestFile`'s
   directory-segment check was case-sensitive, so Symfony's own capitalized
   `Tests/` convention excluded every PHP test file from test-chunk scanning
   before import-matching ever ran. Both had to be fixed together to recover
   real coverage on this repo.

2. **Java** (new `java-same-package-tests.ts`): Java, like Go, grants
   same-package access with no `import` statement at all — Go's #902 fix
   recovers this via "same directory" (compiler-enforced), but Java has no
   such guarantee, and a real multi-module Gradle/Maven build routinely puts
   a test class in a *different* module's source root that happens to
   declare the identical package. Recovered instead via the universal
   Maven/Gradle Standard Directory Layout (`src/<sourceSet>/java/<package/
   path>/...`), mirroring Go's two-tier discipline (tier 1 basename pairing
   folded in unhedged, tier 2 package-level fallback distinctly labeled,
   `lien annotate`-only).

Swift's honest "not determinable" behavior (whole-module imports) is
correct, unchanged, working-as-designed — the previously-shipped #923
symbol-usage fallback already recovers real signal on 20% of sampled files;
the earlier "0%" reading conflated that with the dogfood harness's own
coverage-counting regex, which doesn't match the inferred tier's wording.
No code change needed there.

## Dogfood evidence (required — verbatim `lien annotate` re-measurement against real clones, 25-file sample, same method as the original finding)

| Repo | Language | Before | After (confident "Test coverage:") | After (incl. honest tier-2/inferred) |
|---|---|---|---|---|
| symfony/console | PHP | 0% (0/25) | **48% (12/25)** | 48% |
| square/retrofit | Java | 8% (2/25) | 20% (5/25) | **92% (23/25)** |
| Alamofire/Alamofire | Swift | 0% (0/25) | 0% (honest gap, unchanged, by design) | 20% (5/25, inferred tier — pre-existing #923, not this PR) |
| honojs/hono | TypeScript | 88% | 88% (no regression) | 88% |
| gin-gonic/gin | Go | 64% | 64% (no regression) | 88% |
| tokio-rs/tokio | Rust | 60% | 60% (no regression) | 60% |
| serilog/serilog | C# | 52% | 52% (no regression) | 52% |

Repro:
```
cd packages/cli && node dist/index.js index --force   # in each corpus clone
node dist/index.js annotate Command/Command.php       # symfony/console, was "No test coverage."
# → now: "Test coverage: Tests/ConsoleEventsTest.php, Tests/ConsoleBundleTest.php (+62 more)."
```

## Test plan
- [x] Failing unit tests written first for each root cause, then fixed (`php-psr4.test.ts`, `path-matching.test.ts`, `java-same-package-tests.test.ts`)
- [x] `npm run format:check` / `npm run lint` (0 errors) / `npm run typecheck` / `npm run build` / `npm test` (all packages, 1253+27+1411+378+1701+41 tests, all green) / `node packages/cli/dist/index.js delta` (exit 0, no crossed thresholds)
- [x] Re-measured all 7 corpus repos above with the fixed build (table above)

## Summary by CodeRabbit

- **New Features**
  - Added Java test association support for same-package tests, including basename matching and package-level fallback detection.
  - Expanded test association discovery across annotation, context, and reminder workflows.
  - Added support for Java’s standard Maven/Gradle source layouts.
  - Improved recognition of test directories regardless of capitalization.

- **Bug Fixes**
  - Corrected PSR-4 import resolution when Composer mappings target the workspace root.

---

### Lien Review

| 🔀 test paths | 1 | +0 |
| 🧠 mental load | 1 | +0 |

> [!NOTE]
> **Low Risk**
>
> This PR fixes two real test-association blind spots: PHP PSR-4 root-mapped namespaces (empty-string source dirs) and case-sensitive `Tests/` directory detection, plus adds Java same-package test association via Maven/Gradle Standard Directory Layout. The changes are narrow, well-documented, and mirror the existing Go same-directory discipline. All three fixes are backed by unit tests, and the dogfood table shows measured coverage recovery without regressions in other languages.
>
> - PHP `collectPsr4Entries` now treats empty-string PSR-4 directories as valid root mappings instead of skipping them
> - `isTestFile`'s directory-segment check is now case-insensitive, catching capitalized `Tests/` conventions (e.g., Symfony) while still requiring an exact segment
> - New Java same-package test-association module (`java-same-package-tests.ts`) adds tier-1 basename pairing and tier-2 package-level fallback, wired into `test-associations.ts`, `get-files-context.ts`, and `annotate-cmd.ts`

<sup>Reviewed by [Lien Review](https://lien.dev) for commit 9d83501. Updates automatically on new commits.</sup>

### DIFF
--- a/packages/cli/src/cli/annotate-cmd.ts
+++ b/packages/cli/src/cli/annotate-cmd.ts
@@ -9,13 +9,18 @@ import {
   hasWholeModuleImports,
   hasEnclosingNamespaceAccess,
   hasSameDirectoryTestConvention,
+  hasSamePackageTestConvention,
   isTestFile,
   normalizePath,
   buildGoTestDirIndex,
   findGoPackageLevelTests,
+  toJavaTestCandidate,
+  buildJavaTestDirIndex,
+  findJavaPackageLevelTests,
   type BlastRadiusRisk,
   type CodeChunk,
   type GoTestCandidate,
+  type JavaTestCandidate,
 } from '@liendev/parser';
 import { findDependents, type DependentInfo } from '../mcp/handlers/dependency-analyzer.js';
 import {
@@ -178,36 +183,29 @@ function adaptChunkImports(chunks: DependencyAnalysisChunk[]): CodeChunk[] {
 // signature honest without leaking the SearchResult import here.
 type DependencyAnalysisChunk = Awaited<ReturnType<typeof findDependents>>['allChunks'][number];
 
-/**
- * #902 tier 2, last resort: when `tests` (the core tier-1/import-based
- * result) is empty and `filepath`'s language sets
- * `sameDirectoryTestConvention` (Go), every test file in the same directory
- * — real, same-package signal, but coarser than tier 1, so it's kept
- * strictly separate from `tests` rather than merged into it. This is why
- * `scanTestAssociations`'s `packageLevelTests` field is additive: callers
- * that record or verify against `.tests` (`lookupTestAssociations` /
- * `verify-tests`'s ledger and scope-matching) see zero behavior change —
- * only `lien annotate`'s own printed text (`formatTests`/
- * `formatTestReminder`) consults this.
- */
-function computeGoPackageLevelFallback(
-  tests: string[],
-  filepath: string,
-  allChunks: CodeChunk[],
-  rootDir: string,
-): string[] {
-  if (tests.length > 0) return [];
-  const language = detectLanguage(filepath);
-  if (!language || !hasSameDirectoryTestConvention(language)) return [];
-
+/** Build the shared normalize-with-cache helper `computeGoPackageLevelTests`/`computeJavaPackageLevelTests` both need. */
+function makeCachedNormalizer(rootDir: string): (p: string) => string {
   const cache = new Map<string, string>();
-  const normalize = (p: string): string => {
+  return (p: string): string => {
     if (cache.has(p)) return cache.get(p)!;
     const normalized = normalizePath(p, rootDir);
     cache.set(p, normalized);
     return normalized;
   };
+}
 
+/**
+ * #902 tier 2: every same-directory-test-convention (Go) test file sharing
+ * `filepath`'s directory — real, same-package signal, but coarser than tier
+ * 1. Only called once `computePackageLevelFallback` has confirmed `tests` is
+ * empty and `filepath`'s language is Go.
+ */
+function computeGoPackageLevelTests(
+  filepath: string,
+  allChunks: CodeChunk[],
+  rootDir: string,
+): string[] {
+  const normalize = makeCachedNormalizer(rootDir);
   const candidates: GoTestCandidate[] = allChunks
     .filter(chunk => isTestFile(chunk.metadata.file))
     .filter(chunk => {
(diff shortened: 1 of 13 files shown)


alfhen:
## Coordination note: PR #931 duplicates the PSR-4 fix here

#931 (symbol-level `get_dependents` attribution) independently found and
fixed the exact same PSR-4 empty-string bug as this PR's first commit —
`collectPsr4Entries` treating a composer.json `autoload.psr-4` dir of `""`
as absent, and `normalizeSourceDir` turning `""` into `"/"` instead of `""`.
Found from the dependents side rather than test-association, but it's the
same root cause in the same file (`packages/parser/src/php-psr4.ts`),
semantically identical apart from comments.

**This PR should merge first** (older, already reviewed, and PSR-4
resolution fits its test-association theme better than #931's dependents
scope). Once this merges, #931 will rebase onto `main` and drop its
duplicate commit (`84c56b6b`) — it should go empty or conflict cleanly since
`main` will already have this version of the fix.

No action needed here other than being aware of the ordering — flagging so
whoever merges either PR sees it from both sides.

alfhen:
## This fix is worth more than the title suggests — a third symptom, independently verified

While verifying #931 I traced a separate consumer-visible bug to the **same** case-sensitive `isTestFile` directory check this PR fixes. Recording it here so the value isn't undersold at review, and so there's a concrete post-merge check.

On `symfony/console` at the pinned dogfood SHA, `get_dependents({filepath:"Cursor.php", symbol:"Cursor"})` currently returns:

```
dependentCount            6
productionDependentCount  6
testDependentCount        0
   Tests/CursorTest.php                     | isTestFile: False   <-- test
   Helper/QuestionHelper.php                | isTestFile: False
   Helper/ProgressBar.php                   | isTestFile: False
   Command/InvokableCommand.php             | isTestFile: False
   ArgumentResolver/ArgumentResolver.php    | isTestFile: False
   Tests/Command/InvokableCommandTest.php   | isTestFile: False   <-- test
```

Two of six are plainly tests, both classified as production.

### Why it matters beyond a wrong number

`productionDependentCount` is what makes a change look load-bearing. "6 production dependents" reads as materially scarier than "4 production, 2 test", and that number reaches the model through the blast nudge's risk wording. The `untested` counts in `riskReasoning` derive from the same classification, and so does the recap's "edited files whose associated tests were never observed running" logic — so one case-sensitive check quietly skews three separate surfaces.

### So this PR fixes at least three symptoms of one root cause

1. PHP test-association coverage 0% → 48% (the headline).
2. Capitalized `Tests/` directories excluded from test-chunk scanning.
3. Test files miscounted as production dependents, inflating perceived risk (above).

### Suggested post-merge verification

Re-run the query above and confirm `productionDependentCount: 4 / testDependentCount: 2`. It's a one-command check and it exercises a path the PR's own tests may not cover.

Corpus is at `/tmp/lien-dogfood-460173c9/console`; client is `scripts/experiments/dogfood-oss/mcp-call.mjs` on #917's branch.

(Also note the coordination comment already posted here: #931 duplicates this PR's PSR-4 commit and will drop it on rebase — **this PR should merge first**.)

PR #927, opened 2026-07-28
fix(cli): distinguish unindexed paths from empty results in MCP handlers

alfhen:
## Summary

**Headline defect: `get_dependents` on an unindexed path returns `riskLevel: "low"` — the literal "safe to edit" signal.** A path the index has never heard of (typo'd directory prefix, wrong case, a file outside the indexed project) comes back with `dependentCount: 0`, `riskLevel: "low"`, `riskReasoning: []`. That's not merely a missing warning — it's the exact affirmative signal the tool exists to produce when a file genuinely has no dependents, emitted instead for a file that was never found at all. For the one tool whose entire job is to catch "is this safe to change?" before an edit, a wrong risk verdict is a worse defect than an empty context payload.

`get_files_context` and `get_complexity` have the identical root cause and I confirmed both genuinely have it (this is not a speculative sweep): none of the three checks whether a caller-supplied filepath is actually in the index before answering. All three query chunk/dependency data directly, and an unindexed path produces exactly the same shape as a real file that legitimately has zero chunks/dependents/violations — a confident, empty (or falsely-reassuring) success with no way to tell the two cases apart.

`search_code`, `find_similar`, and `list_functions` were checked and correctly excluded: they take a query/pattern rather than a filepath, a 0-results answer to "does anything match this keyword" is a legitimate and different outcome, and all three already document their own 0-results notes for that reason. The defect is specific to the three filepath-taking handlers above.

## Fix

- New `packages/cli/src/mcp/utils/unindexed-paths.ts`: `findUnindexedPaths()` checks requested filepaths against `ManifestManager.getIndexedFiles()` — the indexer's own ledger of what's indexed, independent of chunk count. Reusing the existing manifest rather than adding a new scan is the right call here: it's what distinguishes "not indexed" from "indexed but genuinely empty" without a fresh `scanAll`. `formatUnindexedPathsNote()` renders an unmissable `⚠ Lien:`-prefixed note (same tone as the existing `complexityHeadroomWarning`).
- Wired into all three handlers via the existing `note` response field — the same diagnostic-note convention `search_code`/`find_similar`/`list_functions` already use, not a new shape.
- A mixed batch (some good paths, some bad) reports exactly the unindexed entries by name — one bad path can't hide behind a good one, and the design deliberately never throws: a mixed batch still returns every valid entry's real data, with the invalid one clearly marked instead of failing the whole call.
- **Fails open** on any manifest-read error (missing `dbPath`, unreadable file, etc.) — mirrors `ManifestManager.load()`'s own fail-open policy. This diagnostic addition can never crash a handler or false-flag a path just because the manifest was momentarily unreadable; a reviewer would otherwise reasonably ask what happens there.
- Updated the three tool descriptions in `packages/cli/src/mcp/tools.ts` to document the new `note` behavior.

## How this was found

Found via a foreign-repo dogfood of Lien's own MCP tools, not a fixture or a fuzzer: I hit this by guessing a plausible `src/` prefix against a real clone of `symfony/console`, which doesn't use that layout (the actual path is `Command/Command.php`, no `src/`). That's an easy, ordinary mistake for an agent working an unfamiliar repo — exactly Lien's target usage — which is what makes this reachable in practice rather than only in adversarial testing.

## Dogfood evidence (real symfony/console clone, real MCP surface)

Reproduced via a minimal stdio MCP client issuing real `callTool` requests against the actual built CLI (`node dist/index.js serve`) — not calling the handler functions directly.

**`get_dependents` — the risk-verdict bug, before:**
```json
{
  "filepath": "src/Command/Command.php",
  "dependentCount": 0,
  "riskLevel": "low",
  "riskReasoning": [],
  "dependents": []
}
```
No error, no warning — reads as "nothing depends on this, safe to change."

**`get_dependents`, after** — same call, now also carries:
```json
"note": "⚠ Lien: not found in the index: \"src/Command/Command.php\". This is NOT the same as \"indexed with no results\" — do not treat it as a low-risk or dependency-free file. Check for a typo (missing/extra directory prefix, wrong case) before editing; try search_code or list_functions to find the real path, or run \"lien index\" if the file was added recently."
```

**`get_files_context` — before:**
```json
{ "file": "src/Command/Command.php", "chunks": [], "testAssociations": [] }
```
**After:** same `chunks`/`testAssociations`, plus the identical `note` shown above.

**Good path unaffected** (`Command/Command.php`, the real path): 6 real chunks returned for `get_files_context`, `get_dependents` reports its real dependent count — only the pre-existing truncation note present, no unindexed-path note in either case.

**Mixed batch** (`get_files_context` with `["Command/Command.php", "does/not/exist.php"]`):
```
note: ⚠ Lien: not found in the index: "does/not/exist.php". ...
files["Command/Command.php"].chunks.length > 0
files["does/not/exist.php"].chunks == []
```
The bad entry is named explicitly; the good entry's real data is untouched — confirms the no-throw, mixed-batch design works as intended.

**`get_complexity`** with `files: ["src/Command/Command.php"]` — before: `filesAnalyzed: 0`, no note. After: same, plus the note. In a mixed batch (`["Command/Command.php", "src/does/not/exist.php"]`), `filesAnalyzed: 1` with the note naming only the bad entry. A whole-codebase sweep (no `files` param) is untouched — no manifest check runs, no note.

**Secondary finding (separate, pre-existing, out of scope for this PR):** while gathering this evidence, `get_dependents` on the bad path at one point returned the *same* `dependentCount`/`riskLevel` as the real path (93 dependents) rather than 0. This traces to `matchesFile`'s import-path fuzzy matching being permissive enough that a PHP namespace-resolved import specifier coincidentally collided with the guessed path — a real, separate bug in path-matching heuristics, unrelated to this fix (this PR's `note` still correctly flags the path as absent from the manifest regardless of what the dependent-count heuristic returns). Flagging for a follow-up, not fixing here to stay in scope.

## Gates

All six pre-commit gates green in a clean worktree branched from `origin/main` (isolated from unrelated concurrent work in progress elsewhere in the repo):
- `npm run format:check` ✅
- `npm run lint` ✅ (0 errors; 38 pre-existing warnings unrelated to this change)
- `npm run typecheck` ✅
- `npm run build` ✅
- `npm test` ✅ (parser 27, core 1387, review 378, action 41, cli 1270 — all packages, exit 0)
- `lien delta` ✅ (exit 0 — 4 pre-existing functions worsened slightly but none crossed threshold; new code (`findUnindexedPaths`, `formatUnindexedPathsNote`) at cognitive complexity 1)

## Test plan

- [x] Unit tests for `findUnindexedPaths`/`formatUnindexedPathsNote`: manifest hit/miss, mixed batch, workspace-root canonicalization, fail-open on a throwing manifest read, fail-open on a missing `dbPath`
- [x] `get_dependents`: unindexed path note, no note when indexed
- [x] `get_files_context`: unindexed single path, indexed-but-empty path (no false positive), mixed batch
- [x] `get_complexity`: unindexed single file, mixed batch, no note for a whole-codebase sweep (no `files` param)
- [x] Full existing suite re-verified green (no regressions from the new manifest check)

---

### Lien Review

| 🔀 test paths | 1 | +0 |
| 🧠 mental load | 1 | +0 |

> [!NOTE]
> **Low Risk**
>
> This PR adds a diagnostic layer that distinguishes unindexed filepaths from genuinely empty results in three filepath-taking MCP handlers (`get_files_context`, `get_dependents`, `get_complexity`). The core logic in `unindexed-paths.ts` checks requested paths against `ManifestManager.getIndexedFiles()`, fails open on any manifest error, and renders a prominent `note` field. The change is localized, well-tested, and the response-shape additions are optional fields documented in the tool descriptions. No callers are broken: the new `note` parameter is optional in `buildSingleFileResponse`/`buildMultiFileResponse`, and sibling handlers that take queries rather than filepaths correctly do not need the check.
>
> - New `findUnindexedPaths`/`formatUnindexedPathsNote` utilities wired into the three filepath-based MCP handlers
> - Optional `note` field added to handler responses without changing existing field semantics
> - Intentional fail-open behavior on manifest read errors to avoid false-flagging indexed paths

<sup>Reviewed by [Lien Review](https://lien.dev) for commit 9f06c3e. Updates automatically on new commits.</sup>

### DIFF
--- a/packages/cli/src/mcp/handlers/get-complexity.test.ts
+++ b/packages/cli/src/mcp/handlers/get-complexity.test.ts
@@ -27,6 +27,16 @@ vi.mock('@liendev/core', async () => {
   };
 });
 
+vi.mock('../utils/unindexed-paths.js', async importOriginal => {
+  const original = await importOriginal();
+  return {
+    ...(original as Record<string, unknown>),
+    findUnindexedPaths: vi.fn().mockResolvedValue([]),
+  };
+});
+
+import { findUnindexedPaths } from '../utils/unindexed-paths.js';
+
 describe('handleGetComplexity', () => {
   const mockVectorDB = {
     scanWithFilter: vi.fn(),
@@ -63,6 +73,7 @@ describe('handleGetComplexity', () => {
     if (mockAnalyzeFn) {
       mockAnalyzeFn.mockClear();
     }
+    vi.mocked(findUnindexedPaths).mockResolvedValue([]);
   });
 
   const getMockAnalyze = () => (globalThis as any).__mockAnalyze;
@@ -548,6 +559,69 @@ describe('handleGetComplexity', () => {
         indexVersion: 1234567890,
         indexDate: '2025-12-19',
       });
+    });
+  });
+
+  describe('unindexed path note (files param)', () => {
+    const emptyReport: ComplexityReport = {
+      summary: {
+        filesAnalyzed: 0,
+        avgComplexity: 0,
+        maxComplexity: 0,
+        totalViolations: 0,
+        bySeverity: { error: 0, warning: 0 },
+      },
+      files: {},
+    };
+
+    it('adds an unmissable note when a requested file has no manifest entry', async () => {
+      vi.mocked(findUnindexedPaths).mockResolvedValue(['src/does/not/exist.ts']);
+      getMockAnalyze().mockResolvedValue(emptyReport);
+
+      const result = await handleGetComplexity({ files: ['src/does/not/exist.ts'] }, mockCtx);
+
+      const parsed = JSON.parse(result.content![0].text);
+      expect(parsed.note).toContain('⚠ Lien:');
+      expect(parsed.note).toContain('"src/does/not/exist.ts"');
+    });
+
+    it('in a mixed batch, names only the unindexed entry — a good path never masks a bad one', async () => {
+      vi.mocked(findUnindexedPaths).mockResolvedValue(['does/not/exist.ts']);
+      getMockAnalyze().mockResolvedValue(emptyReport);
+
+      const result = await handleGetComplexity(
+        { files: ['src/file1.ts', 'does/not/exist.ts'] },
+        mockCtx,
+      );
+
+      expect(findUnindexedPaths).toHaveBeenCalledWith(
+        mockVectorDB,
+        ['src/file1.ts', 'does/not/exist.ts'],
+        expect.any(String),
+      );
+      const parsed = JSON.parse(result.content![0].text);
+      expect(parsed.note).toContain('"does/not/exist.ts"');
+      expect(parsed.note).not.toContain('"src/file1.ts"');
+    });
+
+    it('adds no note when every requested file is indexed', async () => {
+      vi.mocked(findUnindexedPaths).mockResolvedValue([]);
+      getMockAnalyze().mockResolvedValue(emptyReport);
+
+      const result = await handleGetComplexity({ files: ['src/file1.ts'] }, mockCtx);
+
+      const parsed = JSON.parse(result.content![0].text);
+      expect(parsed).not.toHaveProperty('note');
+    });
+
+    it('skips the unindexed-path check entirely for a whole-codebase sweep (no files param)', async () => {
+      getMockAnalyze().mockResolvedValue(emptyReport);
+
+      const result = await handleGetComplexity({ top: 10 }, mockCtx);
+
+      expect(findUnindexedPaths).not.toHaveBeenCalled();
+      const parsed = JSON.parse(result.content![0].text);
+      expect(parsed).not.toHaveProperty('note');
     });
   });
 });
--- a/packages/cli/src/mcp/handlers/get-complexity.ts
+++ b/packages/cli/src/mcp/handlers/get-complexity.ts
@@ -2,6 +2,7 @@ import collect from 'collect.js';
 import { wrapToolHandler } from '../utils/tool-wrapper.js';
 import { GetComplexitySchema } from '../schemas/index.js';
 import type { GetComplexityInput } from '../schemas/index.js';
+import { findUnindexedPaths, formatUnindexedPathsNote } from '../utils/unindexed-paths.js';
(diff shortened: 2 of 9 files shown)


alfhen:
## Review-state caveat before merging — CodeRabbit never actually reviewed this

Triaging the dogfood PRs against the repo's "CI-green ≠ review-clean" rule turned up a subtler variant worth naming: **a green check that is green because the reviewer gave up.**

```
$ gh pr checks 927 | grep -i coderabbit
CodeRabbit   pass   0   Review rate limited
```

The check reports `pass`, but the reason is **"Review rate limited"** — CodeRabbit skipped this PR rather than finding nothing. Compare the sibling dogfood PRs, where it genuinely ran:

| PR | CodeRabbit status | Meaning |
|---|---|---|
| #926 | `Review completed` → "No actionable comments were generated" | genuinely clean |
| #931 | `Review completed`, 0 inline | genuinely clean |
| **#927** | **`Review rate limited`** | **never reviewed** |

Lien Review *did* run here and returned Stable / Low Risk / none-introduced, and there are 0 inline comments — so there is no known problem. But "no findings" and "no review" are different states, and the check surface renders them identically.

Requesting a re-review below. Worth a human confirming it lands before merge rather than reading the green tick as coverage.

@coderabbitai review

alfhen:
**Merging with the CodeRabbit gap on the record.**

CodeRabbit's re-review never landed — its status is still `Review rate limited`, which is an external service constraint rather than a signal about this code, and it shows no sign of clearing.

What *was* verified before merging:

- All CI checks SUCCESS: build-and-test, lint-docs, harness-evidence, review, `lien delta --base`, release smoke test, docs build.
- Lien Review: **Stable — none introduced**, Low Risk.
- Inline review comments: **0**.
- Author's dogfood evidence: verbatim before/after through the real MCP surface against a cloned `symfony/console`, covering the bad path, the good path, and a mixed batch, across all three handlers.
- Full six-gate chain green in a clean worktree branched from origin/main.

So the repo's documented pre-merge gate (triage `lien-stats` + inline comments) is satisfied. Merging on that basis.

Recording it explicitly because "no findings" and "no review" are different states and the green tick renders them identically — if something surfaces here later, this is where to start.

Issue #887, opened 2026-07-25
matcher: two-segment bare requires still fan out (rack-protection shape, one segment outside #868's guard)


alfhen:
From the association-wave acceptance re-sweep (2026-07-25): sinatra's bundled rack-protection sub-gem does `require 'rack/protection'` — a TWO-segment bare specifier, outside #868's single-segment guard scope — and still fans out to all 20 files under rack-protection/lib/rack/protection/. Same failure shape as #868, one segment-count outside what that wave deliberately fixed (the plan's own scope trap forbade rewriting matchesFile wholesale). Not a regression; flagged in the re-sweep as expected residual. Fix belongs in the #886 consolidation design rather than another point guard — filing so the shape is tracked rather than living only in a gitignored memo. Related: #868, #883, #886.

alfhen:
Closed into #886 per the owner — the two-segment shape is now an explicit acceptance case of the consolidation design rather than a standalone point fix.

Issue #884, opened 2026-07-25
matcher: module-name-basename file inside the 1-segment leniency window still false-hubs whole-module-import repos (Alamofire.swift, #868 residual)


alfhen:
The one named acceptance target that survived the association fix wave (acceptance re-sweep 2026-07-25, .wip/resweep-acceptance-2026-07-25.md). Source/Alamofire.swift (a 43-line file whose basename equals the module name) still falsely claims all ~38 Alamofire test files: every test does a bare `import Alamofire` (whole-module), and matching bare import 'Alamofire' against target 'Source/Alamofire' passes #883's strategy-2 guard because it has exactly one leading segment — the same leniency window that legitimately allows auth → src/auth.rs. Confirmed three independent ways (raw primitive, direct matchesFile call, live lien annotate on a fresh index); provably identical before and after #868/#883, i.e. a survivor, not a regression.

Scoped fix directions from the re-sweep (either keeps the general guard untouched):
1. For languages with the existing wholeModuleImports flag (added in #881), suppress the bare-import-equals-file-basename match when the import is the module's own name — extending the #869 honesty treatment to this one coincidental-basename case.
2. Alternatively: a bare import matching a basename-equal file in a wholeModuleImports language only wins if no other evidence contradicts (heavier; probably YAGNI).

Direction 1 is night-sized and uses plumbing that already exists. Related: #868 (closed — the guard works everywhere else, quantified in the re-sweep), #869 (the structural whole-module gap this borders), #883 (the guard implementation).

PR #934, opened 2026-07-29
fix(parser,cli): stop Python bare-module matching from fabricating cross-language test-coverage hubs

### LINKED ISSUE #929: test associations omit a direct-importing test file, and the reminder then names 7 wrong ones
Found by the foreign-repo dogfood (#917), in **TypeScript** — the healthiest language in the corpus (hono, 88% association coverage). Not a PHP/Java/Swift edge case.

## The omission

`src/utils/jwt/jws.ts` in `honojs/hono`. Its associations, per `lien annotate`:

```
Lien impact for src/utils/jwt/jws.ts:
  • 14 files import this — src/utils/jwt/types.ts, src/utils/jwt/jwt.ts, … +10 more; risk: medium
  • Test coverage: runtime-tests/node/index.test.ts, src/middleware/jwk/index.test.ts (+5 more).
```

Seven tests listed. **`src/utils/jwt/jwt.test.ts` is not among them** (`annotate … | grep -c jwt.test.ts` → `0`).

Ground truth — it is a *direct importer* that exercises the file heavily:

```
$ grep -nE "^import|from '" src/utils/jwt/jwt.test.ts
5:import type { HonoJsonWebKey } from './jws'
6:import { signing } from './jws'

$ grep -cE "signing|verifying|jws" src/utils/jwt/jwt.test.ts
39
```

So the association layer omits a same-directory test that imports the target directly, while surfacing seven more distant ones.

## Why it matters more than a missing entry

`test-reminder.sh` renders that list as an instruction:

```
Lien: you changed src/utils/jwt/jws.ts — associated tests: runtime-tests/node/index.test.ts,
src/middleware/jwk/index.test.ts (+5 more). Run them before completing.
```

An agent that obeys runs seven tests and **misses the one that covers its change**. This is worse than silence: it is a confident, specific, actionable instruction pointing away from the relevant test, and it produces a green result that reads as verification.

## Observed behavior in a real trial

In a headless trial on this exact file (plugin installed, no repo `CLAUDE.md`, prompt never mentioning tests), the agent:

- edited `jws.ts`,
- found `jwt.test.ts` **on its own**,
- added 50 lines of cases to it,
- ran `npx vitest run src/utils/jwt/jwt.test.ts` twice.

It ignored the reminder and was right. Following the nudge would have made its behavior *worse* — which inverts the product's value proposition for this case.

Worth noting the rest of the chain behaved correctly: `test-run-note.sh` recorded 4 `test_run` signals, and `recap-stop.sh` correctly stayed silent at Stop because the tests had in fact been run. The session-ledger logic is fine; the input data is wrong.

## Suspected shape

Whatever selects "test coverage" for a file is not simply "test files that import it" — a direct importer in the same directory ranks below seven others, or is filtered out before ranking. Worth checking whether a cap or a relevance ordering is dropping it, since 7 were shown and the list was truncated at "+5 more".

## Related

- #928 — `get_dependents` fabricating graphs via bare-module matching (same subsystem family, different mechanism).
- #926 — PHP/Java association fixes (different root causes; this one survives them and is language-independent as far as we can tell).

alfhen:
## Summary

Fixes #929 — a foreign-repo dogfood finding: `src/utils/jwt/jws.ts` in
honojs/hono showed 7 "Test coverage" entries and omitted
`src/utils/jwt/jwt.test.ts`, a test that directly imports the file and
exercises it 39 times.

## Root cause

`matchesFile`'s Strategy 5 (`matchesPythonModule`) applies Python's "a bare
package import covers every file nested under it" semantic **unconditionally,
regardless of the importer's actual language** (`matchesFile` is otherwise
deliberately language-agnostic). A resolved bare specifier from any other
language can coincidentally look exactly like a Python identifier:

- **TypeScript (hono):** every test that does `import { Hono } from '../..'`
  resolves that specifier down to the bare string `src` — which then
  satisfies `matchesParentPythonPackage('src', 'src/utils/jwt/jws')` for
  **every single file under `src/`**, fabricating "this test covers
  everything" for tests with zero real relationship to the target.
- **Go (gin):** an ordinary whole-package import like
  `"github.com/gin-gonic/gin/binding"` resolves to the bare `binding` after
  module-prefix stripping, and the same unbounded `startsWith` matches every
  file inside `binding/`.

Traced with targeted debug instrumentation against the real hono index —
`jwt.test.ts`'s own direct `./jws` import genuinely matched (it was never
filtered out), it just sorted **last** among 7 matches because the other 6
were bogus `src`-barrel hits, and `lien annotate`'s display cap
(`MAX_TESTS_LISTED = 2`) never got far enough to show it.

## Fix

- `matchesFile` gains an `allowPythonModuleMatching` parameter (default
  `true` — every direct caller of `matchesFile`, including the existing
  Python test fixtures, is unaffected).
- `importMatchesTarget` — the single choke point behind `get_dependents`,
  `get_files_context`, and `lien annotate`'s test-association matching —
  now derives it from the importer's actual language via a new
  `hasPythonModuleSemantics`, mirroring the existing `hasSingleFileImportSemantics`
  (#887) guard exactly.
- `collectImportMatchedTests` (parser) / `collectImportMatchedTestFiles`
  (cli) now additionally rank an **exact literal** import match ahead of any
  fuzzier match, so a genuine direct importer can never again be crowded out
  by scan order alone, independent of the Strategy-5 fix above.

## Dogfood evidence (verbatim, against the existing indexed hono corpus — not re-cloned/re-indexed)

Before:
```
$ node <cli> annotate src/utils/jwt/jws.ts
  • Test coverage: runtime-tests/node/index.test.ts, src/middleware/jwk/index.test.ts (+5 more).
$ node <cli> annotate src/utils/jwt/jws.ts | grep -c "jwt.test.ts"
0
```

After (this fix):
```
$ node <cli> annotate src/utils/jwt/jws.ts
Lien impact for src/utils/jwt/jws.ts:
  • 14 files import this — src/utils/jwt/types.ts, src/utils/jwt/jwt.ts, src/middleware/jwt/jwt.ts, src/middleware/jwk/jwk.ts, +10 more; risk: high (14 callers, max complexity 36, untested high-complexity dependent).
  • Test coverage: src/middleware/jwk/index.test.ts, src/utils/jwt/jwt.test.ts.
  • Max cyclomatic complexity: 15 (1 function over warn threshold).
$ node <cli> annotate src/utils/jwt/jws.ts | grep -c "jwt.test.ts"
1
```

Both remaining entries are genuine direct importers (both literally import
`.../utils/jwt/jws` in their own `imports` list) — the 5 bogus matches are
gone, and `jwt.test.ts` is no longer truncated out.

## Corpus re-measurement (25-file sampler, same method as #926/#929's own baselines)

| Repo | Lang | Baseline (confident) | Pre-fix (this run) | Post-fix (this run) |
|---|---|---|---|---|
| honojs/hono | TS | 22/25 | 21/25 | 13/25 |
| gin-gonic/gin | Go | 16/25 | 16/25 | 8/25 |
| tokio-rs/tokio | Rust | 15/25 | 15/25 | 15/25 |
| serilog/serilog | C# | 13/25 | 1/25 | 0/25 |
| pallets/flask (Python sanity check, not in original scope) | Python | — | 9/25 | 9/25 (**0 flips**) |

**The hono/gin drops are the fix working as intended, not a regression** — I
diffed every flipped file's raw annotate output before/after and manually
verified each one:
- Every hono flip's pre-fix "match" was the **identical** bogus signature
  (`runtime-tests/node/index.test.ts, src/helper/css/index.test.tsx (+3
  more)` — the universal `src`-barrel false hub), for files with no real
  per-file test at all.
- Every gin flip (e.g. `binding/header.go`, `render/html.go`) previously
  "matched" root-package files (`mode_test.go`, `deprecated_test.go`) via
  the bare-`binding`/bare-`render` false hub; post-fix they correctly fall
  through to Go's own honest tier-2 "package-level fallback" message citing
  their **real** same-directory siblings — strictly more correct, just not
  credited by this sampler's narrow `/Test coverage:/i` regex (which,
  per #926's own comment, doesn't count the honest inferred/package-level
  tiers as "confident" either).
- tokio: zero flips — Rust's own Strategy-2 leniency already covered its
  legitimate bare-import cases without ever needing Strategy 5.
- flask: zero flips at all (verified explicitly) — confirms genuine Python
  matching is fully preserved.
- serilog: pre-fix **on this specific indexed corpus** measured 1/25, not
  the cited 13/25 baseline — this mismatch exists independently of this fix
  (reproduced with the unmodified code) and looks like a corpus-state/
  measurement discrepancy unrelated to #929; flagging rather than
  investigating further since re-indexing was out of scope for this task.

### Side finding (not fixed here, out of scope for #929)

A few hono files (e.g. `src/middleware/jsx-renderer/index.ts`) had their
*only* prior "confident" match come from the bogus `src`-barrel hub, masking
a **separate, pre-existing** gap: a bare self-referencing import specifier
(`import { x } from '.'`) is stored literally as `"."` and never resolved to
a real path by the extractor, so it never matches anything even though the
test genuinely imports its own sibling `index.ts`. Worth its own follow-up
issue; not touched here.

## Test plan

- [x] New unit tests: `hasPythonModuleSemantics` + `#929 Python-bare-module
      guard` regression pins in `path-matching.test.ts`, mirroring #884/#887's
      existing test shape (matchesFile alone still permissive by default;
      importMatchesTarget suppresses the cross-language false hub; genuine
      Python matching preserved).
- [x] New unit tests: "direct-importer ranking (#929)" in both
      `test-associations.test.ts` and `get-files-context.test.ts` — an exact
      direct importer scanned *after* a fuzzy match still sorts first in the
      result.
- [x] `npm run format:check` / `npm run lint` (0 errors) / `npm run
      typecheck` / `npm run build` (all packages) / `npm test` (all
      packages: parser 1419, cli 1254, core 27, review 378, site 1701,
      action 41 — all green) / `node packages/cli/dist/index.js delta`
      (exit 0; 4 "worsened but under threshold" cognitive-complexity deltas
      from the added ranking logic, no new-over-threshold crossings)
- [x] Verbatim dogfood re-measurement against the real, already-indexed
      corpus (table above) — no re-clone, no re-index.

---

### Lien Review

| 🔀 test paths | 1 | +0 |
| 🧠 mental load | 1 | +0 |

> [!NOTE]
> **Low Risk**
>
> This PR fixes a cross-language false-hub bug by gating Python bare-module matching on the importer's actual language and by ranking exact direct-importer matches ahead of fuzzy matches in test-association collection. The core change is sound: `matchesFile` keeps its previous default behavior via the new `allowPythonModuleMatching` parameter, while `importMatchesTarget` correctly derives it from `hasPythonModuleSemantics`. The exact-match partitions in both parser and CLI correctly apply the existing `isUnresolvableWholeModuleImport` guard, so Swift whole-module imports cannot jump into the trusted exact bucket. No correctness issues or breaking changes were found.
>
> - `matchesFile` gains `allowPythonModuleMatching` (default `true`); `importMatchesTarget` derives it from the importer language via `hasPythonModuleSemantics`.
> - `collectImportMatchedTests` and `collectImportMatchedTestFiles` now partition exact literal direct imports before fuzzier matches, with the #884 whole-module guard applied to both buckets.
> - New regression tests cover the TypeScript/Go false-hub suppression, preserved Python matching, and exact-importer ranking behavior.

<sup>Reviewed by [Lien Review](https://lien.dev) for commit 2afb178. Updates automatically on new commits.</sup>

---

Attested: degraded:budget_starved · 651K/802K tokens

### DIFF
--- a/.changeset/fix-python-bare-module-false-hub.md
+++ b/.changeset/fix-python-bare-module-false-hub.md
@@ -1,0 +1,36 @@
+---
+"@liendev/parser": patch
+"@liendev/lien": patch
+---
+
+#929: a direct-importing test file could be omitted from `lien annotate`'s
+"Test coverage" line and from `get_files_context`'s `testAssociations`,
+crowded out by unrelated files that matched only through a false hub.
+
+Root cause: `matchesFile`'s Strategy 5 (`matchesPythonModule`) applies
+Python's "a bare package import covers every file nested under it" semantic
+unconditionally, regardless of the importer's actual language. A resolved
+bare specifier from any other language can coincidentally look exactly like
+a Python identifier -- confirmed on a real TypeScript repo (hono), where a
+test's own package-root barrel import (`import { Hono } from '../..'`,
+resolved to the bare specifier `src`) matched every single file under `src/`,
+and on a real Go repo (gin), where an ordinary whole-package import
+(`"github.com/gin-gonic/gin/binding"`, resolved to the bare `binding` after
+module-prefix stripping) matched every file in that package directory. Both
+shapes fabricated "this test covers everything" for files with no real
+relationship to the target, sometimes displacing the file's own genuine
+direct importer once the result list was truncated for display.
+
+Fix: `matchesFile` gains an `allowPythonModuleMatching` parameter (default
+`true`, preserving this function's own behavior for direct callers);
+`importMatchesTarget` -- the shared choke point behind `get_dependents`,
+`get_files_context`, and `lien annotate`'s test-association matching --
+now derives it from the importer's actual language via the new
+`hasPythonModuleSemantics`, mirroring the existing `hasSingleFileImportSemantics`
+(#887) guard. Genuine Python bare-package matching is unaffected (verified:
+zero result changes across a 25-file Python corpus sample).
+
+Additionally, `collectImportMatchedTests`/`collectImportMatchedTestFiles` now
+rank an exact, literal direct import ahead of any fuzzier match, so a real
+direct importer can no longer sort behind other real matches and be
+truncated out of the displayed list purely due to chunk-scan order.
--- a/.changeset/fix-python-bare-module-false-hub.md
+++ b/.changeset/fix-python-bare-module-false-hub.md
@@ -33,4 +33,8 @@ zero result changes across a 25-file Python corpus sample).
 Additionally, `collectImportMatchedTests`/`collectImportMatchedTestFiles` now
 rank an exact, literal direct import ahead of any fuzzier match, so a real
 direct importer can no longer sort behind other real matches and be
-truncated out of the displayed list purely due to chunk-scan order.
+truncated out of the displayed list purely due to chunk-scan order. That
+exact-match check applies the same #884 whole-module guard as the fuzzy
+path, so a Swift bare `import Module` can't jump the queue into the exact
+bucket just because the target file's basename happens to equal the module
+name.
--- a/packages/cli/src/mcp/handlers/get-files-context.test.ts
+++ b/packages/cli/src/mcp/handlers/get-files-context.test.ts
@@ -352,6 +352,32 @@ describe('handleGetFilesContext - test-association scan (scanAll fast path + cac
     });
   });
 
+  describe('direct-importer ranking (#929)', () => {
+    // Mirrors @liendev/parser's own test-associations.test.ts coverage --
+    // this handler is get_files_context's separate implementation (see
+    // collectImportMatchedTestFiles's doc comment), so it needs the
+    // identical ranking guarantee: a test whose own import resolves to
+    // exactly the target must not sort behind a fuzzier match.
+    it('ranks an exact direct importer ahead of a fuzzy match, even when the fuzzy match is scanned first', async () => {
+      const scanAll = vi
+        .fn()
+        .mockResolvedValue([
+          makeChunk('src/auth.rs'),
+          makeChunk('tests/other_test.rs', ['auth']),
+          makeChunk('tests/auth_direct_test.rs', ['src/auth']),
+        ]);
+      const ctx = makeCtx({ scanAll, indexVersion: 1 });
+
+      const result = await handleGetFilesContext(
+        { filepaths: 'src/auth.rs', includeRelated: false },
+        ctx,
+      );
+
+      const parsed = JSON.parse(result.content![0].text);
+      expect(parsed.testAssociations).toEqual(['tests/auth_direct_test.rs', 'tests/other_test.rs']);
+    });
+  });
+
   it('caches the scan across calls with the same indexVersion (no re-scan)', async () => {
     const scanAll = vi.fn().mockResolvedValue([]);
     const ctx = makeCtx({ scanAll, indexVersion: 42 });
--- a/packages/cli/src/mcp/handlers/get-files-context.test.ts
+++ b/packages/cli/src/mcp/handlers/get-files-context.test.ts
@@ -376,6 +376,31 @@ describe('handleGetFilesContext - test-association scan (scanAll fast path + cac
       const parsed = JSON.parse(result.content![0].text);
       expect(parsed.testAssociations).toEqual(['tests/auth_direct_test.rs', 'tests/other_test.rs']);
     });
+
+    it('does not let a Swift whole-module bare import jump the exact-match queue ahead of a real direct importer', async () => {
+      // Mirrors test-associations.test.ts's identical regression pin: a
+      // top-level Swift file whose basename equals its own module name
+      // means a bare `import Alamofire` (whole-module) normalizes to
+      // literally the same string as the target itself -- the #884
+      // Alamofire shape, reintroduced one layer up via the exact-match
+      // bucket unless the whole-module guard is applied there too.
+      const scanAll = vi
+        .fn()
+        .mockResolvedValue([
+          makeChunk('Alamofire.swift'),
+          makeChunk('OtherTests.swift', ['Alamofire']),
+          makeChunk('AlamofireTests.swift', ['./Alamofire']),
+        ]);
+      const ctx = makeCtx({ scanAll, indexVersion: 1 });
+
+      const result = await handleGetFilesContext(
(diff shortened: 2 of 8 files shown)


alfhen:
@coderabbitai review

alfhen:
**Merging.** Triage record:

- **All 22 checks green**, including the full E2E matrix — Python, TypeScript, JavaScript, PHP, Rust, Go, Java, C#, Ruby, Kotlin, Swift, plus MCP round-trip. That matters more here than on the other dogfood PRs: this change alters matching semantics *every* language shares, so an 11-language pass is the right backstop.
- **Lien Review:** Stable, none introduced.
- **Both inline findings fixed and independently verified.** The `isExactDirectImport` helper in each of `test-associations.ts` and `get-files-context.ts` now reads `!isUnresolvableWholeModuleImport(imp, importerFile) && normalize(imp) === normalizedTarget`. Extracting the shared helper also took the complexity deltas from four "worsened" to zero crossings — the tidier fix was the cheaper one.
- **CodeRabbit: `Review rate limited`** — never ran, same external constraint as #927. Recording it rather than reading the green tick as coverage.

I verified two of the author's claims independently rather than on trust:

1. `annotate src/utils/jwt/jws.ts` (this branch, `index --force`) → `Test coverage: src/utils/jwt/jwt.test.ts, src/middleware/jwk/index.test.ts` — direct importer ranked first.
2. The gin coverage drop is genuine false-positive removal: `binding/header.go` exports only unexported internals (`headerBinding`, `Name`, `Bind`, `mapHeader`), and `mode_test.go` contains **zero** references to any of them. It was never real coverage.

Two notes for the record:

**This PR retroactively corrects the dogfood's headline numbers** (#917, and see #929). The M6 coverage table was substantially measuring false positives in barrel-importing languages — hono ~88%→52%, gin ~64%→32%, while Python showed zero flips. The languages I'd called *healthy* were the most inflated, because the Python bare-package heuristic was firing hardest where it had no business firing at all.

**Credit where it's due:** the author rejected my framing. I hypothesized ranking/truncation; the real cause was a cross-language false-positive hub, and the correct fix was nothing like what I asked for. Its own Swift regression test then caught a wrong assumption it had made about the expected output. That's the sequence I want, not compliance with a wrong brief.

Issue #930, opened 2026-07-28
C# global using resolves namespace-wide, producing dependency graphs with zero overlap with reality


alfhen:
Found by the foreign-repo dogfood (#917) on `serilog/serilog`, during a live behavioral trial. Related to #928 (fabricated graphs) and #929 (association pollution) but a **distinct mechanism**, and the most complete failure of the three: for the file below, the reported dependents and the real dependents have **zero overlap**.

## What the model was told

Verbatim `annotate` output shown to an agent mid-trial after it edited `src/Serilog/Parsing/Alignment.cs`:

```
Lien impact for src/Serilog/Parsing/Alignment.cs:
  • 3 files import this — src/Serilog/GlobalUsings.cs, test/Serilog.Tests/GlobalUsings.cs,
    test/Serilog.PerformanceTests/GlobalUsings.cs; risk: medium (3 callers, 1 untested).
  • Test coverage: test/Serilog.Tests/GlobalUsings.cs, test/Serilog.PerformanceTests/GlobalUsings.cs (+1 more).
```

## Ground truth

The three "importers" are `global using` boilerplate. The **real** users of `Alignment`:

```
$ grep -rln "Alignment" src --include='*.cs' | grep -v GlobalUsings
src/Serilog/Formatting/Display/MessageTemplateTextFormatter.cs
src/Serilog/Rendering/MessageTemplateRenderer.cs
src/Serilog/Rendering/Padding.cs
src/Serilog/Parsing/PropertyToken.cs
src/Serilog/Parsing/MessageTemplateParser.cs
```

**5 real dependents, 3 reported, 0 in common.** The risk verdict `medium (3 callers)` is derived entirely from boilerplate.

## Mechanism

`src/Serilog/GlobalUsings.cs` (25 lines, no code) contains namespace-level usings:

```csharp
global using Serilog.Parsing;
global using Serilog.Policies;
global using Serilog.Rendering;
```

`global using <Namespace>;` appears to be resolved as an import of **every file in that namespace**, so each `GlobalUsings.cs` becomes a dependent of every file in every namespace it lists — while the actual type-level usages, which in C# need no per-file `using` once a global using exists, are not counted at all.

That is why the two lists don't merely differ, they invert: the convention that creates the false edges is the same one that removes the real ones.

## Blast radius

- Affects **every** modern C# repo — `global using` is the default template in .NET 6+.
- Sampled 20 serilog source files: **14 of 14** with coverage listed `GlobalUsings.cs`.
[...]

Issue #928, opened 2026-07-28
get_dependents fabricates dependency graphs via language-blind bare-module matching


alfhen:
Found by the foreign-repo dogfood (#917). Distinct from the unindexed-path fix in #927 — that adds a `note`, it does **not** suppress the wrong data — and distinct from the false-zero symbol-attribution bug being fixed separately. All three live in the same subsystem.

## Summary

`matchesFile`'s bare-module fuzzy-match strategy is language-blind and fabricates dependency graphs for **real, correctly-spelled files**.

## Repro — a Rust benchmark with 80 impossible dependents

`benches/copy.rs` in `tokio-rs/tokio` is a benchmark: a leaf compilation target. Nothing in a Rust crate *can* import a bench.

```
$ node scripts/experiments/dogfood-oss/mcp-call.mjs <corpus>/tokio get_dependents \
    '{"filepath":"benches/copy.rs"}'

  dependentCount: 80 | risk: medium
  riskReasoning: ['14 callers', '8 untested', 'max complexity 6']
  first 5 claimed dependents:
    tokio/src/fs/mod.rs
    tokio/src/io/util/copy_bidirectional.rs
    examples/udp-codec.rs
    benches/time_timeout.rs
    benches/sync_semaphore.rs
```

None of these depend on the bench. The matcher is matching the bare module name `copy` against every `use …copy…` in the crate — `tokio::io::util::copy`, `copy_bidirectional`, and so on.

Note also the **internal inconsistency**: `dependentCount: 80` while `riskReasoning` says `14 callers`. Two numbers for the same question in one response.

## Second shape — basename collision on a nonexistent path

```
src/Command/Command.php   (DOES NOT EXIST)  -> 93 dependents, risk "critical"
Command/Command.php       (real file)       -> 93 dependents, risk "critical"
src/Cursor.php            (DOES NOT EXIST)  ->  6 dependents, risk "high"
totally/made/up/File.php  (no collision)    ->  0 dependents, risk "low"
```

A nonexistent path that collides on **basename** silently inherits the real file's entire graph. This looks accidentally right, but it means the tool cannot distinguish paths: in any repo with several `Command.php` files — routine in PHP and Java — it will confidently attribute the wrong file's callers.

Reproduced across four languages (Rust, Python, Go, PHP).

## Why this matters

[...]

Issue #933, opened 2026-07-29
riskLevel can report "low" while its own components say critical


alfhen:
Found while verifying #931 post-merge, on a build from current main (post-#926/#927/#931). Not caused by any of those PRs — #926 *exposed* it by fixing test-file classification, which stopped one boost from firing and revealed that nothing else picks up the slack.

## The contradiction

`symfony/console`, `get_dependents({filepath:"Cursor.php", symbol:"Cursor"})` on current main:

```
dependentCount            6
productionDependentCount  4
testDependentCount        2
riskLevel                 low          <-- the headline verdict
riskReasoning             ['4 callers', 'max complexity 31']
complexityMetrics.maxComplexity        31
complexityMetrics.complexityRiskBoost  critical   <-- contradicts the verdict
```

A file with **4 production callers** and a **`critical`** complexity risk boost is summarised as **`low` risk**, and its own `riskReasoning` cites 4 callers and complexity 31 while the verdict says low.

## Why it matters

`riskLevel` is the field that reaches a model. It is what `api-delta-write.sh` prints, and the server instructions tell agents to gate on it:

> If `riskLevel` is "high" or "critical", list affected dependents to the user before proceeding.

So an agent following the documented rule will *not* list dependents here, because the summary says low — even though the components underneath it say otherwise. It is a milder cousin of the false-zero bug #931 just fixed: not a wrong count, but a wrong *verdict* over correct counts.

## Suspected mechanism

Before #926, `Tests/CursorTest.php` and `Tests/Command/InvokableCommandTest.php` were misclassified as production (case-sensitive `isTestFile` missing capitalised `Tests/`), so the 4 real production dependents counted as "untested" and the untested-boost pushed the verdict up. #926 correctly fixed the classification — and with the untested-boost no longer firing, `complexityRiskBoost: critical` apparently does not lift the verdict on its own.

So the composition looks like it is dominated by the untested count, with `complexityRiskBoost` computed but not effectively contributing. Contrast a case where it does resolve correctly, same build:

```
gin internal/bytesconv/bytesconv.go, symbol StringToBytes
  dependentCount 4 | riskLevel high
  riskReasoning ['4 callers', '1 untested', 'max complexity 27', 'untested high-complexity dependent']
```

Identical caller count, verdict `high` — because one dependent is untested. The only difference between `high` and `low` here is the untested count, not the blast radius or the complexity.

## Suggested direction

Two candidate invariants, either of which would have caught this:

[...]

PR #936, opened 2026-07-29
fix(cli): honest dependentAttributionIncomplete for C# zero-dependent get_dependents scans

alfhen:
## Summary

Follow-up to #930. The false-edge fix (`global using` no longer resolving
to a file-to-file import) already shipped in #932 — **that part is on
`main`**, not this PR. This PR addresses the concern raised on #932 after
merge: removing the false edges alone leaves `get_dependents` reporting a
confidently-empty `dependentCount: 0` / `riskLevel: "low"` for a C# file
with real callers the import graph structurally can't see (C#'s enclosing-
namespace access means a real caller needs no per-file `using` at all once
a `global using` exists for that namespace). A silent, confident zero is
arguably worse than the fabrication it replaced — it reads as "safe to
change" on a file that may have real callers Lien simply never saw.

**Fix:** `get_dependents` now sets `dependentAttributionIncomplete: true`
plus a `dependentAttributionNote` whenever a file-level query (no `symbol`)
comes back with zero dependents for a language where
`hasEnclosingNamespaceAccess` is set (only C# today). This mirrors the
`symbolAttributionDegraded` pattern already shipped for symbol-level
queries (#928) — same idea, applied to the file-level "zero" case instead
of the symbol-level "unresolvable" case.

```json
{
  "dependentCount": 0,
  "riskLevel": "low",
  "dependents": [],
  "dependentAttributionIncomplete": true,
  "dependentAttributionNote": "No import-based dependents were found for src/Serilog/Parsing/Alignment.cs, but its language lets real callers use its exports with no per-file import naming it at all (C#'s \"global using\" / implicit enclosing-namespace member access). The import graph has no signal for that usage shape, so dependentCount: 0 and riskLevel: \"low\" here mean \"the scan found nothing,\" not \"nothing depends on this file\" — don't treat this as a verified clear."
}
```

The flag only fires for the specific shape the concern was about: a
file-level query, genuinely zero dependents, in an enclosing-namespace-
access language. It does NOT fire when:
- a `symbol` was requested (that's `symbolAttributionDegraded`'s territory, unaffected),
- the file has any real dependents at all (even 1),
- the language doesn't have `hasEnclosingNamespaceAccess` set (every other language, unaffected).

## What this does NOT do

It does not recover the true dependents. That needs type-reference-based
resolution the codebase doesn't have today for the general
`get_dependents`/`annotate` path — see the "Part 2" discussion on #930/#932
for why that's a separate, larger effort (no existing extension point,
unlike the directory-based same-package/same-directory heuristics
`test-associations.ts` already has for Go/Java test-association matching).
This PR makes the failure mode honest instead of silently misleading, which
is the smaller, well-scoped alternative explicitly called out as acceptable
when full type-reference resolution is out of scope.

## Verbatim dogfood (serilog/serilog, real 254-file C# corpus, reindexed with this branch's build)

```
$ node <cli> serve   # via MCP get_dependents
get_dependents({"filepath":"src/Serilog/Parsing/Alignment.cs"})
```
```json
{
  "dependentCount": 0,
  "testDependentCount": 0,
  "riskLevel": "low",
  "riskReasoning": [],
  "dependents": [],
  "dependentAttributionIncomplete": true,
  "dependentAttributionNote": "No import-based dependents were found for src/Serilog/Parsing/Alignment.cs, but its language lets real callers use its exports with no per-file import naming it at all (C#'s \"global using\" / implicit enclosing-namespace member access). The import graph has no signal for that usage shape, so dependentCount: 0 and riskLevel: \"low\" here mean \"the scan found nothing,\" not \"nothing depends on this file\" — don't treat this as a verified clear."
}
```

Regression-checked the same corpus:
- `PropertyToken.cs` (1 real test dependent found): `dependentAttributionIncomplete` correctly **absent** — the flag doesn't fire when real dependents exist.
- A `symbol`-scoped query on a C# file with zero dependents: `dependentAttributionIncomplete` correctly **absent** — that's `symbolAttributionDegraded`'s territory, unaffected.
- A TypeScript file with zero dependents (hono corpus, no `enclosingNamespaceAccess`): flag correctly never fires for non-C# languages (guaranteed by `hasEnclosingNamespaceAccess` only being set for C# — covered by existing tests, not language-specific to this PR).

## Gate chain (all six)

```
npm run format:check   ✅ 0 issues
npm run lint            ✅ 0 errors (38 pre-existing warnings, none in touched files)
npm run typecheck       ✅ 0 errors
npm run build           ✅ all 5 packages
npm test                ✅ 211 files, 0 failures (parser 50, core 36, review 57, cli 62, action 5, parser-native 1)
npm run docs:build      ✅ + read the rendered mcp-tools.html output (renders cleanly)
node packages/cli/dist/index.js delta --base origin/main
                         ✅ exit 0 — 3 advisory "worsened" only (findDependents Halstead-time
                            metric 49m→54m after extracting the new logic into its own
                            `checkDependentAttributionIncomplete` helper to keep it under the
                            60m threshold it briefly crossed pre-extraction; logRiskAssessment
                            9→10, buildDependentsResponse 4→5 — all well under their limits),
                            no new-over-threshold crossing
```

Diff vs `origin/main` is exactly 6 files: the two handler files + their
tests, the docs update, and a new changeset (kept separate from #932's
already-shipped, not-yet-released changeset rather than reopening it).

## Files

- `packages/cli/src/mcp/handlers/dependency-analyzer.ts` — `checkDependentAttributionIncomplete` + wiring into `findDependents`'s return
- `packages/cli/src/mcp/handlers/get-dependents.ts` — response fields + `buildDependentsResponse`/`logRiskAssessment` wiring
- `packages/cli/src/mcp/handlers/dependency-analyzer.test.ts`, `get-dependents.test.ts` — new tests mirroring the `symbolAttributionDegraded` test pattern
- `packages/site/docs/guide/mcp-tools.md` — response-field documentation
- `.changeset/csharp-dependent-attribution-incomplete.md`

---

### Lien Review

> [!NOTE]
> **Low Risk**
>
> This PR adds an honesty flag (`dependentAttributionIncomplete` + `dependentAttributionNote`) to file-level `get_dependents` queries that return zero dependents for C# files, where enclosing-namespace access and `global using` mean real callers may exist without any per-file import edge. The change is narrowly scoped: it only fires when there is no `symbol`, zero dependents, and the language has `hasEnclosingNamespaceAccess` set (currently only C#). The implementation correctly checks all three conditions, surfaces the flag through the MCP response, and avoids duplicate logging. Tests cover the trigger case and the three exclusion cases (symbol query, non-zero dependents, non-C# language). Documentation and guidance claims match the code behavior. No breaking changes or wrong-output paths were found.
>
> - Added `checkDependentAttributionIncomplete` helper that sets `dependentAttributionIncomplete` only for file-level, zero-dependent C# queries
> - Surfaced `dependentAttributionIncomplete` and `dependentAttributionNote` in the `get_dependents` MCP response via `buildDependentsResponse`
> - Updated `logRiskAssessment` to accept the optional warning level parameter and skip generic logging when the flag is set
> - Added tests and updated MCP tool docs/instructions

<sup>Reviewed by [Lien Review](https://lien.dev) for commit bd49c69. Updates automatically on new commits.</sup>

## Summary by CodeRabbit

- **New Features**
  - `get_dependents` now flags cases where import-based analysis may not identify all file dependents.
  - Responses include an explanatory note when zero dependents cannot be treated as a confirmed absence of usage.
  - Risk reporting avoids presenting these cases as definitively low risk.

- **Documentation**
  - Clarified how to interpret `dependentAttributionIncomplete`, zero dependent counts, and low-risk results for file-level queries.

- **Tests**
  - Added coverage for C#, TypeScript, symbol-level, and normal dependent queries.

---

Attested: degraded:budget_starved · 632K/892K tokens

### DIFF
--- a/.changeset/csharp-dependent-attribution-incomplete.md
+++ b/.changeset/csharp-dependent-attribution-incomplete.md
@@ -1,0 +1,23 @@
+---
+"@liendev/lien": patch
+---
+
+Follow-up to #930 (`global using` no longer producing false import edges,
+shipped in #932): removing those false edges alone left `get_dependents`
+reporting a confidently-empty `dependentCount: 0` / `riskLevel: "low"` for a
+C# file with real callers Lien simply has no signal for — a false-all-clear
+that's arguably worse than the fabrication it replaced, since C#'s
+enclosing-namespace access means a real caller needs no per-file `using` at
+all once a `global using` exists for a namespace.
+
+`get_dependents` now sets `dependentAttributionIncomplete: true` plus a
+`dependentAttributionNote` for exactly that shape (a file-level query with
+zero dependents found, in a language where `hasEnclosingNamespaceAccess` is
+set) — mirroring the `symbolAttributionDegraded` pattern already shipped for
+symbol-level queries (#928). `dependentCount: 0` in that case now reads as
+"the import graph found nothing," not a verified "nothing depends on this
+file."
+
+This does not recover the true dependents (that needs type-reference-based
+resolution the codebase doesn't have today — tracked separately); it makes
+the failure mode honest instead of silently misleading.
--- a/packages/cli/src/mcp/handlers/dependency-analyzer.test.ts
+++ b/packages/cli/src/mcp/handlers/dependency-analyzer.test.ts
@@ -414,6 +414,54 @@ describe('findDependents', () => {
     });
   });
 
+  describe('dependentAttributionIncomplete (C# enclosing-namespace-access floor, #930)', () => {
+    it('flags a file-level query with zero dependents for a C# file', async () => {
+      // Once a `global using` exists for a namespace, a real C# caller needs
+      // no per-file `using` at all -- the import graph has no signal for
+      // that usage, so a zero-dependent result here is "nothing found," not
+      // a verified "nothing depends on this file."
+      mockDB.scanAll.mockResolvedValue([createChunk('Alignment.cs', { exports: ['Alignment'] })]);
+
+      const result = await findDependents(mockDB as any, 'Alignment.cs', mockLog);
+
+      expect(result.dependents).toHaveLength(0);
+      expect(result.dependentAttributionIncomplete).toBe(true);
+      expect(mockLog).toHaveBeenCalledWith(
+        expect.stringContaining('enclosing-namespace access'),
+        'warning',
+      );
+    });
+
+    it('does not flag a C# file that has real file-level dependents', async () => {
+      mockDB.scanAll.mockResolvedValue([
+        createChunk('Alignment.cs', { exports: ['Alignment'] }),
+        createChunk('Consumer.cs', { imports: ['Alignment.cs'] }),
+      ]);
+
+      const result = await findDependents(mockDB as any, 'Alignment.cs', mockLog);
+
+      expect(result.dependents).toHaveLength(1);
+      expect(result.dependentAttributionIncomplete).toBeUndefined();
+    });
+
+    it('does not flag a zero-dependent TypeScript file (no enclosing-namespace access)', async () => {
+      mockDB.scanAll.mockResolvedValue([createChunk('src/target.ts', { exports: ['doStuff'] })]);
+
+      const result = await findDependents(mockDB as any, 'src/target.ts', mockLog);
+
+      expect(result.dependents).toHaveLength(0);
+      expect(result.dependentAttributionIncomplete).toBeUndefined();
+    });
+
+    it('does not flag a zero-dependent C# SYMBOL query (that is symbolAttributionDegraded/real-empty territory instead)', async () => {
+      mockDB.scanAll.mockResolvedValue([createChunk('Alignment.cs', { exports: ['Alignment'] })]);
+
+      const result = await findDependents(mockDB as any, 'Alignment.cs', mockLog, 'Alignment');
+
+      expect(result.dependentAttributionIncomplete).toBeUndefined();
+    });
+  });
+
   describe('complexity metrics', () => {
     it('should calculate correct file and overall complexity metrics', async () => {
       mockDB.scanAll.mockResolvedValue([
--- a/packages/cli/src/mcp/handlers/dependency-analyzer.ts
+++ b/packages/cli/src/mcp/handlers/dependency-analyzer.ts
@@ -9,6 +9,8 @@ import {
   isUnresolvableWholeModuleImport,
   importMatchesTarget,
   hasSingleFileImportSemantics,
+  detectLanguage,
+  hasEnclosingNamespaceAccess,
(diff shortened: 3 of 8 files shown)


alfhen:
@coderabbitai review

alfhen:
## Closed the doc gap myself (author had gone idle) — plus the riskLevel call

The mechanism here is right and I verified it independently: with this branch built and `index --force`, `get_dependents {"filepath":"src/Serilog/Parsing/Alignment.cs"}` → `dependentCount 0 | riskLevel low | dependentAttributionIncomplete true`, while `PropertyToken.cs` (1 real dependent) correctly carries no flag.

But the flag wasn't reaching its consumer:

```
tools.ts mentions dependentAttributionIncomplete:        0
instructions.ts mentions dependentAttributionIncomplete: 0
packages/site/docs/guide/mcp-tools.md:                   1
```

Documented for **humans**, invisible to the **model**. An agent received a field it was never told existed, while `instructions.ts` still told it to gate on `riskLevel` — which still reads `"low"` for a file whose callers are admittedly unknowable. That is the same failure shape this dogfood keeps surfacing: the information exists internally and isn't surfaced to the consumer.

Added to both model-facing surfaces, in actionable form:

> When this is true, treat `dependentCount` 0 and `riskLevel` "low" as a **FLOOR, not a finding** — it means Lien could not see callers, not that none exist. Verify with grep before concluding a symbol is unused. `riskLevel` is not adjusted for this and may still read "low"; the flag is the authoritative signal.

### On `riskLevel` — deliberately unchanged, and why

I considered making `riskLevel` non-`low` when attribution is incomplete and decided against it **in this PR**. It is the same family as #933 (`riskLevel: "low"` sitting alongside `complexityRiskBoost: "critical"` on the same response), and the aggregate-vs-components question deserves one considered change rather than two partial ones that could disagree. Documenting the limitation plainly is the honest interim: a reader is now told, at the point of use, that the verdict is unreliable when the flag is set.

If a reviewer prefers `riskLevel` change here instead, that's a reasonable call — but then #933 should be closed by the same change.

### Verification

- Built bundle carries both new strings (`treat dependentCount 0 and riskLevel`, `grep before concluding the symbol is unused`).
- Live server's `serverInstructions` confirmed to include the flag.
- `format:check` clean, `lint` 0 errors, `typecheck` clean, `packages/cli` tests 1284 passed, `lien delta --base origin/main` exit 0.

Also worth noting for the record: the author's correction on attribution was right and better-evidenced than my own. It built a throwaway worktree at `5a21f45c` (with #926 **and** #931 merged), reindexed from scratch, and reproduced the original bug byte-for-byte — proving #926 never touched `csharp.ts`. #932 was the part-1 fix. My earlier "#926 already did part 1" claim was wrong twice over, and it took the author's clean-room reproduction to pin down why.

PR #937, opened 2026-07-29
fix(parser,cli): stop get_dependents from fabricating dependency graphs

### LINKED ISSUE #928: get_dependents fabricates dependency graphs via language-blind bare-module matching

alfhen:
## Summary

Fixes #928: `matchesFile`'s fuzzy-match strategies are language-blind and can fabricate an entire dependency graph for a file — a confident wrong `get_dependents` answer, not a degraded one, since this is the tool the MCP server instructions require callers to check before changing any exported symbol.

Two independent root causes, both found via the foreign-repo dogfood (#917):

1. **Rust `self::`/`super::` collapsed to a directory-less string at extraction time.** `RustImportExtractor`/`convertRustModulePath` stripped `self::`/`super::` down to a bare (or merely `../`-prefixed) specifier with zero knowledge of the importing file's real location. That ambiguous string then had to be resolved by `matchesFile`'s generic bare-identifier leniency — the same leniency that legitimately supports Rust's `crate::auth` -> `src/auth.rs` convention — which has no way to tell "this bare word came from a real crate-root reference" apart from "this bare word came from a same-directory sibling reference that happens to collide with an unrelated top-level directory." On tokio-rs/tokio, `tokio/src/fs/mod.rs`'s `self::copy` and `tokio/src/io/util/copy_bidirectional.rs`'s `super::copy` both fuzzy-matched the completely unrelated `benches/copy.rs` (a leaf benchmark — nothing can import a bench), and the (otherwise correct) re-export/transitive-dependent graph walk then pulled in every real consumer of the real `tokio::fs::copy`/`tokio::io::util::copy` functions, fabricating 80 dependents for the bench.

   **Fix:** teach the Rust extractor to resolve `self::`/`super::` precisely against the importing file's own location, using Rust's actual file-to-module convention (a `mod.rs`/`lib.rs`/`main.rs` represents its own containing directory; any other file represents a module nested one level inside it — so `super::` from a "leaf" file stays in the same directory, it does not ascend). This is threaded through as a new, Rust-only, ungated `rustImporterFile` parameter (mirroring the existing `rustCrateMap` precedent) — every other language's extractor ignores it, and Rust keeps its exact prior behavior when the parameter is omitted. See `resolveRustRelativeModulePath`'s doc comment in `packages/parser/src/ast/languages/rust.ts` for the full four-shape derivation.

2. **No existence check before the fuzzy search.** `get_dependents` never checked whether the requested `filepath` was actually present in the index before running the fuzzy dependent search — it only cared whether some import specifier textually resolved to the target string, never whether the target itself was real. `matchesFile`'s multi-segment boundary strategy (Strategy 2) deliberately has no cap on how many extra leading directories the *target* may carry beyond an already-resolved import specifier (this is required — it's exactly how a real PHP PSR-4 vendor prefix like `Domain/Services/Auth` legitimately matches `web/Domain/Services/Auth`), which means a nonexistent path that shares a real file's trailing path segments is *textually indistinguishable* from a legitimate deep-prefixed match. On symfony/console, a guessed `src/Command/Command.php` (doesn't exist — the repo has no `src/`) returned the identical 93-dependent, `critical`-risk graph as the real `Command/Command.php`, purely because `Command/Command` is a suffix of `src/Command/Command`.

   **Fix:** `findDependents` now checks `allChunksByFile.has(normalizedTarget)` (free — it's the same map already built from the existing scan, keyed by the same normalizer) before running the fuzzy search at all. When the target has no chunks anywhere in the index, the search is skipped entirely: `dependents` comes back empty, a new `targetIndexed: false` field is set on `DependencyAnalysisResult`, and `get_dependents` surfaces an explicit `⚠ Lien:` `note` so a `dependentCount: 0`/`riskLevel: "low"` reads as "unresolved," not "confirmed safe." This is a self-contained, in-memory check (no new I/O, no dependency on the not-yet-merged #927) — when #927 lands, its manifest-based `note` and this one will agree rather than contradict.

Also addresses the reported `dependentCount`/`riskReasoning` inconsistency (`dependentCount: 80` next to `"14 callers"` in the original repro): the reasoning was already computed from `productionDependentCount` by design (a test file calling the target shouldn't weigh into risk the same way a production caller does), just without saying so. Now labeled `"N production callers"` explicitly, so the two numbers are visibly, deliberately different rather than looking like a bug in themselves.

## Acceptance criteria

- [x] tokio `benches/copy.rs` -> not 80 fabricated dependents. **0**, and `targetIndexed: true`/no note (it's a real, indexed file — this is a confirmed-zero answer, not an "unresolved" one).
- [x] `src/Command/Command.php` (nonexistent) does **not** return the real file's 93-dependent graph. Now **0**, `targetIndexed: false`, explicit note.
- [x] `Command/Command.php` (real) still returns its correct graph. Still **93**, `critical`, unchanged.
- [x] `dependentCount` and `riskReasoning`'s caller count agree or the difference is explained. Explained: `riskReasoning` now says `"N production callers"`.
- [x] No regression on file-level lookups that currently work — full test suite green (below), plus real cross-file Rust dependents are now *more* accurate, not just fewer (see dogfood evidence).

## Dogfood evidence (real corpora, real built CLI, real MCP tool calls)

Per CLAUDE.md's mandatory dogfood policy, all four cases were re-run against **copies** of the actual corpora from the original repro (`/tmp/lien-dogfood-460173c9`, not re-cloned, not re-indexed in place — the team's shared corpus indices were left untouched; I copied `tokio/` and `console/` to a scratch location, indexed the copies via `lien serve`'s on-connect auto-index using this branch's own freshly built CLI, and queried through a real MCP `stdio` client (`@modelcontextprotocol/sdk`), not by calling handler functions directly).

**1. tokio `benches/copy.rs` — before:**
```json
{
  "dependentCount": 80,
  "productionDependentCount": 14,
  "testDependentCount": 66,
  "riskLevel": "medium",
  "riskReasoning": ["14 callers", "8 untested", "max complexity 6"],
  "dependents": [
    { "filepath": "tokio/src/fs/mod.rs" },
    { "filepath": "tokio/src/io/util/copy_bidirectional.rs" },
    { "filepath": "examples/udp-codec.rs" },
    { "filepath": "benches/time_timeout.rs" },
    "...76 more, mostly tokio/tests/fs_*.rs and benches/*.rs"
  ]
}
```

**after:**
```json
{
  "dependentCount": 0,
  "testDependentCount": 0,
  "riskLevel": "low",
  "riskReasoning": [],
  "dependents": []
}
```
(no `note` field — the file is real and indexed; this is a confirmed-zero answer)

**Accuracy check — the 79 real callers previously misattributed to the bench are now correctly attributed to the real file that defines the function:**
```
get_dependents({filepath: "tokio/src/fs/copy.rs"})
  -> dependentCount: 79, first: tokio/src/fs/mod.rs, examples/udp-codec.rs, benches/time_timeout.rs, ...
get_dependents({filepath: "tokio/src/io/util/copy.rs"})
  -> dependentCount: 155, first: tokio/src/io/util/copy_bidirectional.rs, examples/udp-codec.rs, ...
get_dependents({filepath: "tokio/src/io/util/copy_bidirectional.rs"})
  -> dependentCount: 154, first: examples/udp-codec.rs, benches/time_timeout.rs, ...
```

**2. PHP `src/Command/Command.php` (nonexistent) — before:**
```json
{ "dependentCount": 93, "riskLevel": "critical",
  "riskReasoning": ["93 callers", "93 untested", "max complexity 28", "untested high-complexity dependent"] }
```
**after:**
```json
{
  "dependentCount": 0,
  "riskLevel": "low",
  "riskReasoning": [],
  "note": "⚠ Lien: \"src/Command/Command.php\" has no chunks anywhere in the index — every count above is a deliberate 0, not a confirmed empty dependency graph. This can mean the path was never indexed, is misspelled (wrong directory prefix, wrong case), or genuinely has no extractable content. Do not treat this as a low-risk or dependency-free file; check for a typo before editing, try search_code or list_functions to find the real path, or run \"lien index\" if the file was added recently."
}
```

**3. PHP `Command/Command.php` (real) — before and after (unchanged, confirming no regression):**
```json
{
  "dependentCount": 93,
  "riskLevel": "critical",
  "riskReasoning": ["93 production callers", "93 untested", "max complexity 28", "untested high-complexity dependent"],
  "dependents": [{ "filepath": "SingleCommandApplication.php" }, { "filepath": "ConsoleBundle.php" }, { "filepath": "Application.php" }, "...90 more"]
}
```
(`"93 callers"` -> `"93 production callers"` is the only change — the label clarification described above; the count and every dependent are identical)

**4. PHP `src/Cursor.php` (nonexistent) — before:** `dependentCount: 6, riskLevel: "high"`. **after:**
```json
{ "dependentCount": 0, "riskLevel": "low", "note": "⚠ Lien: \"src/Cursor.php\" has no chunks anywhere in the index — ..." }
```

**5. PHP `totally/made/up/File.php` (no collision, for completeness) — before:** `dependentCount: 0, riskLevel: "low"`, no distinguishing signal from a real isolated file. **after:** same `dependentCount: 0`, now with the same `note`, honestly flagging it as unresolved rather than looking identical to a confirmed-empty real file.

## Gates

All six pre-commit gates green in this worktree (branched from `origin/main`):
- `npm run format:check` — pass
- `npm run lint` — 0 errors, 38 pre-existing warnings (unrelated to this change, same count as reported on a recent sibling PR)
- `npm run typecheck` — pass across all packages
- `npm run build` — pass
- `npm test` (all packages) — parser 1397, core 378, review 1701, action 41, cli 1259 — all green
- `lien delta` — exit 0. No new-over-threshold crossings. One pre-existing-but-under-threshold function (`findDependents`) worsened from 49m to 50m "time to understand" after extracting the new existence-check into its own named helper (`seedIfTargetIndexed`) specifically to keep it from crossing; several Rust extractor functions show minor "worsened" time-to-understand from the added `importerFile` parameter threading, none crossing their thresholds.

## Constraints followed

- Own worktree branched from `origin/main` (not `worktree-fix-symbol-dependents`, which is reserved for the separate symbol-attribution false-zero bug).
- Did not touch `packages/parser/src/utils/path-matching.ts`'s `matchesFile`/`matchesAtBoundaryPrecise`/PHP-namespace logic at all — I traced the PHP repro down to that file's deliberately-uncapped multi-segment leading-directory leniency (needed for real PSR-4 vendor prefixes) and confirmed a purely textual fix there is impossible without also breaking legitimate matches of the identical shape; the existence check in `dependency-analyzer.ts` is the sound fix instead. This keeps this PR's diff clear of the file PR #926 is concurrently modifying.
- Did not touch `dependency-analyzer.ts`'s symbol-level attribution path (the false-zero bug on `worktree-fix-symbol-dependents`) — my change to that file is additive (a new existence-check gate ahead of the existing fuzzy search) and should rebase cleanly.

---

## Update: rebased onto main (2026-07-29)

Main moved several times since this PR was opened (#931 symbol attribution, #932 C# global using, #934 Python bare-module, #917 dogfood docs, #927 unindexed-path notes, #936 C# dependentAttributionIncomplete). Rebased and resolved real conflicts in `dependency-analyzer.ts`, `get-dependents.ts`/`.test.ts`, and `symbols.ts` (the last from #931's `processImportSymbolsList` API change, which needed the same `importerFile` threading as the method it superseded).

**Re-verified after rebase, with a fresh `lien index --force` on newly-copied corpora and this branch's own freshly rebuilt CLI (not a stale/other-worktree binary):**
- tokio `benches/copy.rs` → still `dependentCount: 0`, no `note` (`targetIndexed: true` — a real, indexed file, confirmed-zero, not unresolved).
- `src/Command/Command.php` (nonexistent) → still `dependentCount: 0`, with a `note`.
- `Command/Command.php` (real) → still `dependentCount: 93` (unchanged — no dependents lost). The `riskReasoning` count itself now reads `"29 production callers"` rather than `"93"`: unrelated to this PR, #926 (merged since) improved PHP test-file detection (capitalized `Tests/`), so 64 of the 93 are now correctly recognized as test files that were previously miscounted as production. The label mechanism this PR adds is doing exactly what it's supposed to — reporting whatever the (now more accurate) production count actually is.

**Checked #934 doesn't subsume any of this PR's fix, independently (not just by reading its PR body):** built #934's branch (`fix/929-direct-importer-test-priority`) in isolation, force-reindexed a fresh tokio copy, and confirmed `benches/copy.rs` still returns 80 fabricated dependents against #934's build alone — it fixes a different `matchesFile` strategy (Strategy 5, Python bare-module matching applied cross-language) than the one this PR's tokio case goes through (Strategy 2, bare-single-segment leniency, fed by Rust's `self::`/`super::` collapsing to an ambiguous string). Zero file-level overlap between the two PRs either.

**Checked this PR's `targetIndexed`-based `note` against #927's manifest-based `note`, now that both are live on `main` and testable together (not just asserted):** added a test (`does not duplicate or contradict the #927 manifest note when both mechanisms would fire`) covering the common case where a nonexistent path is both "not in the manifest" and "has no chunks" — confirmed only #927's note appears (the more authoritative "is this path even part of the indexed project" signal), never a second, redundant explanation. `buildDependentsResponse` now checks the manifest note first and only falls back to this PR's chunk-based note for the narrower case the manifest doesn't cover (a path the manifest lists as indexed that produced zero chunks in the current scan). Reproduced live against real corpora post-rebase: the fake PHP path's `note` reads with #927's exact wording ("not found in the index: ..."), not this PR's ("has no chunks anywhere in the index").

All six gates re-verified green on the rebased tree: format, lint (0 errors), typecheck, build, full test suite (parser 1441, core 378, review 1701, action 41, cli 1293 — all green), `lien delta` (exit 0, no crossings).

---

### Lien Review

| 🧠 mental load | 1 | +0 |

> [!NOTE]
> **Low Risk**
>
> This PR fixes two root causes of fabricated dependency graphs in `get_dependents`: (1) Rust `self::`/`super::` imports are now resolved against the importer's actual file-to-module location rather than collapsing to ambiguous bare strings, and (2) `findDependents` now verifies the target file exists in the index before running fuzzy matching, returning `targetIndexed: false` and an explicit note for unresolvable paths. It also clarifies risk reasoning by relabeling "N callers" to "N production callers". The changes are well-scoped, backward-compatible when the new `importerFile` parameter is omitted, and covered by targeted tests for the exact boundary cases (unindexed target, real isolated file, basename-collision nonexistent path, and Rust leaf/mod.rs resolution shapes).
>
> - Rust `self::`/`super::` resolution now uses importer-aware module semantics via `resolveRustRelativeModulePath`
> - `findDependents` gates fuzzy search on `allChunksByFile.has(normalizedTarget)` and surfaces `targetIndexed`
> - Risk reasoning relabeled to "N production callers" to explain the deliberate production-only scope

<sup>Reviewed by [Lien Review](https://lien.dev) for commit 2129a3c. Updates automatically on new commits.</sup>

---

Attested: degraded:budget_starved · 657K/814K tokens

### DIFF
--- a/.changeset/fix-fabricated-dependency-graphs.md
+++ b/.changeset/fix-fabricated-dependency-graphs.md
@@ -44,3 +44,8 @@ Also reconciles the reported `dependentCount`/`riskReasoning` mismatch (e.g.
 production-only callers by design (test callers shouldn't weigh into risk
 the same way), just without saying so — now labeled "N production callers"
 explicitly.
+
+`targetIndexed`'s `note` defers to #927's manifest-based `note` whenever both
+would fire for the same target (the overwhelming common case — a
+nonexistent path is both "not in the manifest" and "has no chunks"): only
+one note is ever shown, never two competing explanations of the same zero.
--- a/.changeset/fix-fabricated-dependency-graphs.md
+++ b/.changeset/fix-fabricated-dependency-graphs.md
@@ -1,0 +1,46 @@
+---
+"@liendev/parser": patch
+"@liendev/lien": patch
+---
+
+#928: `get_dependents` fabricated dependency graphs via language-blind
+bare-module matching — a confident wrong answer, not a degraded one, since
+the tool exists specifically to catch "is this safe to change?" before an
+edit.
+
+Two independent shapes, found via the foreign-repo dogfood (#917):
+
+- **Rust `self::`/`super::` collapsed to a directory-less string.**
+  `RustImportExtractor` stripped these keywords down to a bare or merely
+  `../`-prefixed specifier with no knowledge of the importer's real location,
+  which `matchesFile`'s generic bare-identifier leniency (designed for the
+  legitimate `crate::auth` -> `src/auth.rs` convention) then had to guess at
+  match time — and could coincidentally match an unrelated same-named file
+  elsewhere with a single leading directory. Reproduced on tokio-rs/tokio:
+  `benches/copy.rs` (a leaf benchmark nothing can import) fuzzy-matched
+  `tokio/src/fs/mod.rs`'s `self::copy` and
+  `tokio/src/io/util/copy_bidirectional.rs`'s `super::copy`, fabricating 80
+  dependents. Fixed by resolving `self::`/`super::` precisely against the
+  importer's own file-to-module-aware location (`resolveRustRelativeModulePath`
+  in `ast/languages/rust.ts`) instead of the old lossy string convention —
+  this also makes real Rust cross-file dependency tracking MORE accurate
+  (the 79 real callers of `tokio::fs::copy` are now correctly attributed to
+  `tokio/src/fs/copy.rs`, not to the unrelated benchmark).
+- **No existence check before the fuzzy search.** A nonexistent path that
+  collides on a namespace/directory suffix with a real file silently
+  inherited that file's entire graph — `src/Command/Command.php` (guessed,
+  doesn't exist) returned the same 93 dependents as the real
+  `Command/Command.php`, because `matchesFile`'s multi-segment boundary
+  strategy has no cap on extra leading target directories (deliberately, to
+  support e.g. PHP PSR-4 vendor prefixes) — there is no purely textual way to
+  tell a real deep path from a fabricated one with the same suffix. Fixed by
+  checking the target actually has chunks in the index before running the
+  fuzzy search at all (`get_dependents`'s new `targetIndexed` result field);
+  an unresolvable target now always comes back with zero dependents and an
+  explicit `note`, never someone else's graph.
+
+Also reconciles the reported `dependentCount`/`riskReasoning` mismatch (e.g.
+`dependentCount: 80` next to `"14 callers"`): the reasoning already counted
+production-only callers by design (test callers shouldn't weigh into risk
+the same way), just without saying so — now labeled "N production callers"
+explicitly.
--- a/packages/cli/src/cli/api-delta-cmd.test.ts
+++ b/packages/cli/src/cli/api-delta-cmd.test.ts
@@ -409,6 +409,28 @@ describe('apiDeltaCommand — enrichment when an index is present', () => {
     await fs.mkdir(realIndexDir, { recursive: true });
     await fs.writeFile(path.join(realIndexDir, 'structural.db'), '', 'utf-8');
 
+    // The target file itself must have a chunk in the stub index (#928) --
+    // get_dependents/findDependents now requires the requested target to be
+    // indexed at all before running its fuzzy dependent search, to stop an
+    // unresolvable path from silently inheriting an unrelated file's graph.
+    // A real index always has at least one chunk for a.ts itself; this stub
+    // must model that too, or the target reads as "not indexed" and
+    // dependentCount comes back 0 regardless of caller.ts below.
+    const targetChunk = {
+      content: 'export function formatUser(user) { return user.name; }',
+      metadata: {
+        file: 'a.ts',
+        startLine: 1,
+        endLine: 1,
+        type: 'function',
+        language: 'typescript',
+        symbolName: 'formatUser',
+        symbolType: 'function',
+        exports: ['formatUser'],
+      },
+      score: 0,
+      relevance: 'not_relevant',
+    };
     const callerChunk = {
       content: 'formatUser(x);',
       metadata: {
@@ -429,7 +451,7 @@ describe('apiDeltaCommand — enrichment when an index is present', () => {
     vi.mocked(coreModule.createVectorDB).mockResolvedValueOnce({
       initialize: vi.fn().mockResolvedValue(undefined),
       getCurrentVersion: vi.fn().mockReturnValue(1),
-      scanAll: vi.fn().mockResolvedValue([callerChunk]),
+      scanAll: vi.fn().mockResolvedValue([targetChunk, callerChunk]),
     } as unknown as Awaited<ReturnType<typeof coreModule.createVectorDB>>);
 
     await expect(apiDeltaCommand({ format: 'json', file: 'a.ts' })).rejects.toThrow('__exit__:0');
--- a/packages/cli/src/mcp/handlers/dependency-analyzer.test.ts
+++ b/packages/cli/src/mcp/handlers/dependency-analyzer.test.ts
@@ -120,6 +120,11 @@ describe('findDependents', () => {
         createChunk('rack-protection/lib/rack/protection/base.rb', {
           imports: ['rack/protection'],
         }),
+        // The requested target must itself be indexed (#928) -- otherwise
+        // this test would pass vacuously (empty dependents for an
+        // unresolved target) rather than actually exercising the anti-fan-
+        // out guard it's named for.
+        createChunk('rack-protection/lib/rack/protection/xss_header.rb'),
       ]);
 
       const result = await findDependents(
@@ -138,6 +143,10 @@ describe('findDependents', () => {
         createChunk('rack-protection/lib/rack/protection/base.rb', {
           imports: ['rack/protection'],
         }),
+        // The requested target must itself be indexed (#928) for the fuzzy
+        // dependent search to run at all -- a real umbrella entry point is
+        // always its own chunk in a real index.
+        createChunk('rack-protection/lib/rack/protection.rb'),
       ]);
 
       const result = await findDependents(
@@ -159,6 +168,8 @@ describe('findDependents', () => {
       // real gin clone) by applying Ruby's stricter anchor unconditionally.
       mockDB.scanAll.mockResolvedValue([
         createChunk('render/html.go', { imports: ['internal/fs'] }),
+        // The requested target must itself be indexed (#928).
+        createChunk('internal/fs/fs.go'),
       ]);
 
       const result = await findDependents(mockDB as any, 'internal/fs/fs.go', mockLog);
@@ -174,6 +185,8 @@ describe('findDependents', () => {
       mockDB.scanAll.mockResolvedValue([
         createChunk('render/html.go', { imports: ['pkg/sub'] }),
         createChunk('lib/pkg/consumer.rb', { imports: ['pkg/sub'] }),
+        // The requested target must itself be indexed (#928).
+        createChunk('pkg/sub/child.go'),
       ]);
 
       const result = await findDependents(mockDB as any, 'pkg/sub/child.go', mockLog);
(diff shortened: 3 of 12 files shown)


alfhen:
**Merging.** Independently verified against your rebased build (own worktree resolved from the PR's `headRefName`, fresh `npm run build`, `lien index --force` on fresh corpus copies):

```
tokio  benches/copy.rs            80 → 0
FAKE   src/Command/Command.php    93 + contradictory note → 0, low, ONE coherent note
REAL   Command/Command.php        93 preserved (prod 29 / test 64), risk critical
```

The third line is the one that makes this safe to merge. A fix for "returns too many dependents" could easily have been a fix that returns too few; the real file keeps its entire graph, and `riskReasoning` now reads `"29 production callers"` rather than `93` because #926's capitalized-`Tests/` detection correctly reclassified 64 of them. Different number, more accurate, not a regression.

### On the note collision — you fixed a real problem, not a hypothetical one

I asked you to *test* whether your note duplicated #927's rather than assert it, and it turned out they both wrote to the same `response.note`. Giving #927's manifest note precedence with yours as fallback, plus a test firing both mocks and asserting only #927's wording survives, is the right resolution. Confirmed live: the fake path emits #927's text alone.

Verified on main before your fix, for the record — the contradiction was real:

> `dependentCount: 93`, `riskLevel: "critical"`, **and** `note: ⚠ Lien: not found in the index` — in one response.

### And you resolved the four-flag concern rather than adding to it

I'd worried about a fourth "untrustworthy zero" vocabulary item landing on `get_dependents`. `targetIndexed` is internal-only — used solely as the condition to emit the note, never surfaced as a response field, so a model learns nothing new. That's better than what I asked for: #927 remains the single consumer-facing message and your check simply stops the bad data being produced. #927 explains; this prevents.

I was wrong earlier to suggest these two might be redundant. They're complementary and both were needed.

### Triage

All checks green including the E2E matrix. Lien Review: Stable, none introduced. Inline: 0. **CodeRabbit: `Review rate limited`** — never ran, same external constraint as #927/#934/#936; recording rather than reading the tick as coverage.

Two process notes worth preserving from your report: the stale-`packages/parser/dist` trap after a rebase (missing #931's new exports looked like 20 self-inflicted test failures until a full rebuild), and catching your own mis-threaded Rust parameter only by rebuilding against a real corpus. Both are the same lesson this whole dogfood kept teaching — the measurement substrate lies more often than the code does.

PR #939, opened 2026-07-29
docs(cli): document get_dependents' unindexed-target note at model-facing surfaces

alfhen:
## Summary

Small follow-up to #937, which merged before this doc pass landed on top of it.

`get_dependents`'s `note` field (existing since #927, extended by #937 with a second, chunk-based fallback source — see #937's PR body for the full precedence rationale) was documented at the *human*-facing surface (`packages/site/docs/guide/mcp-tools.md`) but not at either surface a model actually reads:

- `SERVER_INSTRUCTIONS` (`packages/cli/src/mcp/instructions.ts`) — sent on every MCP connection. Already documented `dependentAttributionIncomplete` (#936) but said nothing about `note`.
- The `get_dependents` tool description (`packages/cli/src/mcp/tools.ts`) — already documented `note`'s existence, but not that it can come from either of two checks (manifest lookup or chunk presence) and is deliberately never doubled.

This closes that gap, and expands the site docs paragraph to match.

## Why this is worth its own PR

#937 merged (as #937/8c3b2ce9) before this commit was added on top of it in the same branch — by the time the doc work was ready, the branch had already been merged and deleted. Re-based this commit onto current `origin/main` directly; it's the exact same change, no logic touched.

## Gates

- `npm run format:check` / `npm run lint` (0 errors) / `npm run typecheck` / `npm run build` / `npm test` (cli 1295, all green) / `lien delta` (exit 0, no complexity-affecting changes — text-only)
- `packages/cli/src/mcp/tools.test.ts` and `packages/cli/src/mcp/instructions.claude-md-sync.test.ts` (the two tests that would catch a doc/schema mismatch) explicitly re-run and green.

Docs-only change; no dogfood beyond the gates above applies (nothing in the tool's actual behavior changed, only what a model is told about it).

---

### Lien Review

> [!NOTE]
> **Low Risk**
>
> Docs-only follow-up that documents the `note` field semantics for `get_dependents` at the model-facing `SERVER_INSTRUCTIONS` and tool-description surfaces, plus the user-facing site guide. The prose accurately reflects the implementation: a `note` is emitted when the requested `filepath` cannot be resolved in the index, produced by either a manifest lookup (`findUnindexedPaths`) or a zero-chunks check (`!analysis.targetIndexed`), with the manifest-based note taking precedence so only one note is ever returned. No code behavior changed.
>
> - Added `note` semantics to `SERVER_INSTRUCTIONS` in `packages/cli/src/mcp/instructions.ts`
> - Expanded the `get_dependents` tool description in `packages/cli/src/mcp/tools.ts` to explain the two fallback checks and single-note guarantee
> - Mirrored the same explanation in `packages/site/docs/guide/mcp-tools.md`

<sup>Reviewed by [Lien Review](https://lien.dev) for commit 9f66459. Updates automatically on new commits.</sup>

### DIFF
--- a/packages/cli/src/mcp/instructions.ts
+++ b/packages/cli/src/mcp/instructions.ts
@@ -33,6 +33,11 @@ symbol:
   riskLevel of "low" mean Lien could not SEE the callers, not that there are none —
   grep before concluding the symbol is unused. riskLevel is not adjusted for it.
 
+  If a result carries a note, filepath is not resolvable in the index at all
+  (never indexed, misspelled, or a typo'd directory prefix) — dependentCount: 0
+  and riskLevel: "low" then mean "unresolved", not "confirmed safe to change".
+  Fix the path (try search_code or list_functions) before trusting the result.
+
 For discovery ("where is X?", "how does Y work?"), call search_code FIRST.
 It runs full-text BM25 keyword search over code, docstrings, and camelCase-split
 identifiers. Query with concrete KEYWORDS, identifiers, and domain terms
--- a/packages/cli/src/mcp/tools.ts
+++ b/packages/cli/src/mcp/tools.ts
@@ -144,7 +144,7 @@ Returns:
 - dependents[]: { filepath, isTestFile, usages[]? }
 - complexityMetrics: { averageComplexity, maxComplexity, highComplexityDependents[] }
 - totalUsageCount?: number (when symbol parameter provided)
-- note?: present and explicit when \`filepath\` has no entry in the index at all — a dependentCount of 0 / riskLevel "low" can otherwise look identical whether the file genuinely has no dependents or the path was never indexed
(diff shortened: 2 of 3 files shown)


alfhen:
**Merging.** Verified the claim rather than just the wording, because "if a result carries a note" is the kind of statement that can be quietly overbroad.

The discovery-tool sweep earlier in this dogfood found `note` is **overloaded** on `search_code`/`find_similar`/`list_functions`, where it also carries response-budget truncation warnings. So a blanket "a note means the path is unresolvable" would have been wrong there. Checked whether it's true on this path:

```
get-dependents.ts        note used only for unindexed-target (#927 manifest, #937 chunk-based,
                         mutually exclusive via the precedence added in #937)
dependency-analyzer.ts   no truncation/budget note in the dependents path
```

Accurate for `get_dependents`, and it sits in that tool's own section rather than the shared preamble. Good.

The wording is also actionable rather than merely descriptive — *"Fix the path (try search_code or list_functions) before trusting the result"* tells a model what to do next, which is the whole point of putting it where a model reads.

This closes the last gap in the family: the *behaviour* landed in #937, and the two surfaces a model actually reads now describe it. Worth noting that #936 had the same shape — a correct fix documented only in `packages/site`, i.e. for humans — so this is twice in one campaign that a signal shipped invisible to its consumer. That pattern is itself worth watching for in review.

Checks green, Lien Review Good, 0 inline. CodeRabbit rate-limited again (recorded, not read as coverage).

Issue #940, opened 2026-07-29
Unify get_dependents' attribution-caveat flags (symbolAttributionDegraded / dependentAttributionIncomplete / note) behind one shape


alfhen:
## Summary

`get_dependents` now carries four separate mechanisms that each mean some version of "this zero (or this count) is not a fully trustworthy answer":

1. `symbolAttributionDegraded` + `symbolAttributionNote` (#931) — symbol-level attribution couldn't confirm call sites; the response widened to the file-level answer instead.
2. `dependentAttributionIncomplete` + `dependentAttributionNote` (#936) — a file-level query returned zero dependents in a language whose real callers need no per-file import (C#'s `global using`).
3. `note` (#927, extended by #937) — `filepath` isn't resolvable in the index at all (not in the manifest, or has zero chunks in the current scan).

That's three distinct flag/note pairs (four including the two independent sources behind #3, which #937 made single-output but didn't unify structurally) with different names, different shapes, and no shared vocabulary — a model has to learn each one individually, and `riskLevel` still reads `"low"` under all of them regardless of which fired.

## Where this came from

Raised during review of #937 (the `get_dependents` fabricated-graph fix): #937's `note` and #927's `note` needed reconciling into a single non-duplicated field (done — see #937's PR body for the precedence rule), which surfaced the broader question of whether this whole family should share a structure.

## Proposed shape (for discussion, not a commitment)

Something like a single optional field:

```ts
attributionCaveat?: {
  reason: 'unresolved-target' | 'symbol-attribution-degraded' | 'dependent-attribution-incomplete';
  note: string;
}
```

so a model (or any consumer) can check one field's presence rather than three, and the `reason` enum gives a stable, greppable identifier per case instead of three differently-shaped booleans. This would be a breaking change to `get_dependents`'s response shape, so it needs its own design pass and a migration plan (probably additive first, with the existing flags deprecated rather than removed immediately) — not something to fold into a bug-fix PR.

## Related

- #933 — `riskLevel` can itself contradict its own component signals (e.g. `complexityRiskBoost: "critical"` under a `riskLevel: "low"` verdict). Same underlying shape: an aggregate/headline field that doesn't fully reflect the caveats sitting next to it. Worth deciding both together, since a unified `attributionCaveat` might also want to feed into whatever fixes #933.

## Non-goals for this issue

[...]

PR #941, opened 2026-07-29
fix(parser,cli): resolve get_dependents risk-verdict contradiction, unify attribution caveats

### LINKED ISSUE #933: riskLevel can report "low" while its own components say critical

alfhen:
## Summary

Closes #933 and #940 — a coherence pass on `get_dependents`, done together
per the tracking issues since #940's unification touches the exact surface
#933's fix lives in.

### #933 — riskLevel contradicted its own components

`computeBlastRadiusRisk` (the shared risk primitive used by `get_dependents`,
`lien delta`, `annotate`, and review's blast-radius injection) only ever
factored dependent complexity into the verdict through
`hasHighComplexityUncovered`, which requires an untested dependent to fire
at all. A file with a **critical**-complexity dependent but **zero**
untested dependents therefore came back `riskLevel: "low"` while its own
`complexityMetrics.complexityRiskBoost` read `"critical"` — confirmed on
`symfony/console`'s `Cursor.php` (4 fully-tested production callers, max
complexity 31).

**Chosen invariant**: a `high`/`critical` `complexityRiskBoost` now floors
the verdict **one tier below its own severity** (`critical` → at least
`high`, `high` → at least `medium`), regardless of test coverage — testing
lowers the odds of a *silent* break, it doesn't shrink the blast radius.
I picked this over "riskLevel must not be low above some dependent count"
because it's the more surgical fix for the actual root cause (complexity
was computed but not wired into classification) and doesn't require picking
an arbitrary new count threshold. A monotonicity test in
`blast-radius-risk.test.ts` asserts `riskLevel` is never more than one tier
below `complexityRiskBoost` across a spread of scenarios.

The one-tier floor (not "clamp to the boost's own level") is what keeps the
already-correct gin `bytesconv.StringToBytes` case (1 untested dependent,
critical complexity, already `"high"` via `hasHighComplexityUncovered`)
from escalating to `"critical"` — it's the regression guard in the table
below.

### #940 — three attribution-caveat flags, one question

`symbolAttributionDegraded`/`symbolAttributionNote` (#931),
`dependentAttributionIncomplete`/`dependentAttributionNote` (#936), and
`note` (#927/#937) all mean some version of "this count/zero isn't a
verified clear." I judged the proposed shape **correct to build as-is**:

- The three are **mutually exclusive by construction** — verified by
  reading the actual control flow (`checkDependentAttributionIncomplete`
  early-returns when `symbol` is set or `!targetIndexed`;
  `symbolAttributionDegraded` requires `chunksByFile.size > 0`, which is
  false whenever `!targetIndexed`). A single `reason` enum loses no
  information versus three booleans.
- It's a genuine ergonomic win, not just cosmetic: a model checks one
  field's presence instead of three ORed conditions, and gets a stable,
  greppable identifier per case.
- This is a **breaking change** to the response shape with **no
  deprecation window**. Accepted given: `@liendev/lien` is pre-1.0 (0.71.x),
  the three flags being replaced are only weeks old, and the project has a
  track record of breaking MCP/response-shape changes without shims
  (LanceDB/embeddings removal, native-parser swap, etc.). An additive
  deprecated-fields-linger approach would add permanent surface for zero
  benefit at this stage.
- `targetIndexed` stays internal (drives the `unresolved-target` note,
  never promoted to a response field), per the non-goal in #940.

New shape: `attributionCaveat?: { reason: 'unresolved-target' |
'symbol-attribution-degraded' | 'dependent-attribution-incomplete'; note:
string }`.

### Where this is documented (model-facing, not just packages/site)

This exact campaign shipped a correct fix twice that was invisible to its
consumer (documented only in `packages/site`, with zero mentions in
`tools.ts`/`instructions.ts`). This PR updates all of:

- `packages/cli/src/mcp/tools.ts` — `get_dependents`'s tool description
- `packages/cli/src/mcp/instructions.ts` — server instructions sent to
  every connecting client
- `packages/cli/src/mcp/schemas/dependents.schema.ts` — the `symbol`
  param's own description (previously referenced the old
  `symbolAttributionDegraded` field name)
- `packages/site/docs/guide/mcp-tools.md` — human-facing docs

Verified against the **built bundle** (`packages/cli/dist/index.js`), not
just source: `attributionCaveat` appears 9×, the invariant sentence ("even
if every dependent is fully tested") is present, and the old field names
only remain where they should — the internal `DependencyAnalysisResult`
computation in `dependency-analyzer.ts`/`annotate-cmd.ts`, which is a
separate surface out of scope for #940 (annotate's hook output never
carries `symbol`, so it never had `symbolAttributionDegraded` to unify in
the first place; it keeps its own `dependentAttributionIncomplete` field
unchanged).

## Dogfood evidence (real corpora, before/after)

Built the native parser locally, indexed each repo fresh
(`lien index --force`) immediately before every measurement, per the
protocol in `docs/dogfood/`. The shared corpus at
`/tmp/lien-dogfood-460173c9/gin` had uncommitted changes from a concurrent
agent's task (a renamed symbol) at measurement time, so the gin
measurements below use a clean private clone instead — same commit
(`34dac20`), verified via `git status --short` empty.

### console `Cursor.php` symbol `Cursor` — the #933 case

| | before | after |
|---|---|---|
| `riskLevel` | `"low"` | **`"high"`** |
| `riskReasoning` | `["4 production callers", "max complexity 31"]` | `[..., "critical-complexity dependent regardless of test coverage"]` |
| `complexityMetrics.complexityRiskBoost` | `"critical"` | `"critical"` (unchanged) |

### gin `bytesconv.go` symbol `StringToBytes` — regression guard, must stay `high`

| | before | after |
|---|---|---|
| `riskLevel` | `"high"` | **`"high"` (unchanged)** |
| `riskReasoning` | `["4 production callers", "1 untested", "max complexity 27", "untested high-complexity dependent"]` | identical |

### console `Cursor.php` symbol `__construct` — #940 symbol-attribution-degraded, combined with #933

```json
{
  "riskLevel": "high",
  "riskReasoning": ["4 production callers", "max complexity 31", "critical-complexity dependent regardless of test coverage"],
  "attributionCaveat": {
    "reason": "symbol-attribution-degraded",
    "note": "\"__construct\" doesn't appear in Cursor.php's tracked top-level exports ..."
  }
}
```

### serilog `Alignment.cs` (file-level) — #940 dependent-attribution-incomplete, unaffected by #933

`riskLevel: "low"` (no complexity data — correctly unaffected),
`attributionCaveat.reason: "dependent-attribution-incomplete"`.

### console `src/Command/Command.php` (nonexistent) — #940 unresolved-target, unaffected

`attributionCaveat.reason: "unresolved-target"`, note text unchanged
(`"not found in the index"` — #927's manifest-based wording, precedence
over #928's still intact).

### gin `binding/binding.go` — healthy file, entirely unaffected

`riskLevel: "medium"`, `complexityRiskBoost: "low"` — before and after
identical (floor only ever activates for `high`/`critical` boosts).

## Test plan

- [x] `npm run format:check` — pass
- [x] `npm run lint` — 0 errors (pre-existing warnings only, unrelated)
- [x] `npm run typecheck` — pass across all packages
- [x] `npm run build` — pass across all packages
- [x] `npm test` — 1296 CLI tests + parser/core/review/action suites, all pass
- [x] `node packages/cli/dist/index.js delta` — exit 0 (worsened-but-under-threshold only; `buildDependentsResponse` actually improved: cognitive 8 → 3)
- [x] `npm run docs:build` — pass, rendered HTML spot-checked for the new `attributionCaveat` section
- [x] New monotonicity + regression-guard tests in `blast-radius-risk.test.ts` and `get-dependents.test.ts`
- [x] Real-corpus dogfood before/after on 5 cases (table above), including one full end-to-end MCP call combining both fixes

---

### Lien Review

> [!WARNING]
>
> This PR fixes a coherence bug in `get_dependents` by wiring `complexityMetrics.complexityRiskBoost` into `computeBlastRadiusRisk` as a floor, so high/critical complexity can no longer be silenced to `low` by full test coverage. It also unifies three attribution-caveat flags into a single `attributionCaveat: { reason, note }` response field. The changes are well-tested for the parser/cli paths, but the review-side blast-radius caller was not updated to pass the new `complexityRiskBoost` parameter, leaving the original contradiction bug alive in PR review reports.
>
> - `computeBlastRadiusRisk` now accepts optional `complexityRiskBoost` and floors `riskLevel` one tier below high/critical boosts
> - `get_dependents` response replaces separate caveat flags with `attributionCaveat: { reason, note }`
> - Review's `buildEntry`/`computeGlobalRisk` still omit `complexityRiskBoost`, so the fix does not propagate to PR review blast-radius scoring

<sup>Reviewed by [Lien Review](https://lien.dev) for commit 0558218. Updates automatically on new commits.</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved blast-radius risk scoring so critical or high complexity cannot be understated by test coverage.
  * Risk explanations now clarify when complexity influences the result regardless of coverage.

* **New Features**
  * Added a unified `attributionCaveat` response containing a reason and explanatory note.
  * Consolidated unresolved-target, degraded symbol attribution, and incomplete dependent attribution warnings.

* **Documentation**
  * Updated tool guidance, parameter descriptions, server instructions, and documentation to reflect the revised risk semantics and attribution response format.
  * Removed legacy attribution warning fields; this is a breaking response-shape change.

### DIFF
--- a/.changeset/riskverdict-attribution-unify.md
+++ b/.changeset/riskverdict-attribution-unify.md
@@ -1,6 +1,6 @@
 ---
-"@liendev/parser": patch
-"@liendev/lien": patch
+"@liendev/parser": minor
+"@liendev/lien": minor
 ---
 
 Fix `get_dependents`' risk verdict contradicting its own components (#933),
--- a/.changeset/riskverdict-attribution-unify.md
+++ b/.changeset/riskverdict-attribution-unify.md
@@ -1,0 +1,43 @@
+---
+"@liendev/parser": patch
+"@liendev/lien": patch
+---
+
+Fix `get_dependents`' risk verdict contradicting its own components (#933),
+and unify its three attribution-caveat flags into one field (#940).
+
+**#933**: `computeBlastRadiusRisk` only ever factored complexity into the
+verdict through `hasHighComplexityUncovered`, which requires an untested
+dependent to fire at all — so a file with zero untested dependents but a
+`critical`-complexity caller came back `riskLevel: "low"` while its own
+`complexityMetrics.complexityRiskBoost` read `"critical"` (confirmed on
+symfony/console's `Cursor.php`: 4 fully-tested production callers, max
+complexity 31, `riskLevel: "low"`). `riskLevel` is the field
+`instructions.ts` tells agents to gate on before editing an exported symbol,
+so the wrong verdict misled even though the underlying counts were correct.
+
+Fixed by adding an optional `complexityRiskBoost` input to
+`computeBlastRadiusRisk`: a `high`/`critical` boost now floors the verdict
+one tier below its own severity (`critical` → at least `high`, `high` → at
+least `medium`), regardless of test coverage — testedness lowers the odds
+of a *silent* break, it doesn't shrink the blast radius. The
+untested-and-high-complexity case (`hasHighComplexityUncovered`) already
+reaches full severity on its own and is unaffected; verified against gin's
+`bytesconv.StringToBytes` (1 untested, critical complexity) staying `high`,
+not escalating to `critical`.
+
+**#940**: `symbolAttributionDegraded`/`symbolAttributionNote`,
+`dependentAttributionIncomplete`/`dependentAttributionNote`, and `note` all
+meant some version of "this count isn't a verified clear," with three
+different names and shapes for a model to learn — and were mutually
+exclusive by construction, so a single field always sufficed. Replaced with
+one optional `attributionCaveat: { reason, note }`, where `reason` is
+`'unresolved-target' | 'symbol-attribution-degraded' |
+'dependent-attribution-incomplete'`. This is a breaking response-shape
+change with no deprecation window — acceptable pre-1.0, and the fields were
+only weeks old. `targetIndexed` stays internal, as before.
+
+Both `tools.ts`'s tool description and `instructions.ts`'s server
+instructions (the two surfaces a connecting model actually reads) are
+updated accordingly, along with the `symbol` parameter's own schema
+description and the `get_dependents` docs page.
--- a/packages/cli/src/cli/annotate-cmd.ts
+++ b/packages/cli/src/cli/annotate-cmd.ts
@@ -374,6 +374,7 @@ async function computeAnnotationData(
     // separately on the display line below — different signal.
     maxDependentComplexity: result.complexityMetrics.maxComplexity,
     hasHighComplexityUncovered,
+    complexityRiskBoost: result.complexityMetrics.complexityRiskBoost,
   });
 
   return {
--- a/packages/cli/src/cli/api-delta-cmd.ts
+++ b/packages/cli/src/cli/api-delta-cmd.ts
@@ -101,6 +101,7 @@ function computeRisk(analysis: DependencyAnalysisResult): string {
     uncoveredDependents: uncoveredProductionDependents,
     maxDependentComplexity: maxComplexity > 0 ? maxComplexity : undefined,
     hasHighComplexityUncovered,
+    complexityRiskBoost: complexityMetrics.complexityRiskBoost,
   }).level;
 }
 
--- a/packages/cli/src/mcp/handlers/get-dependents.test.ts
+++ b/packages/cli/src/mcp/handlers/get-dependents.test.ts
@@ -260,8 +260,14 @@ describe('handleGetDependents', () => {
       );
     });
 
-    it('should keep risk low when complexity is high but all dependents are tested', async () => {
-      // hasHighComplexityUncovered only fires when uncovered > 0.
+    it('should not report low risk when complexity is critical, even with full test coverage (#933)', async () => {
+      // Regression test for #933: a critical complexityRiskBoost must never
+      // be paired with riskLevel "low" -- being tested lowers the odds of a
+      // *silent* break, it doesn't shrink the blast radius of a
+      // critical-complexity caller. Before the fix, hasHighComplexityUncovered
+      // (the only path complexity fed into the verdict) required uncovered > 0
+      // to fire at all, so this exact shape came back "low" despite its own
+      // complexityRiskBoost reading "critical".
       vi.mocked(findDependents).mockResolvedValue(
         createMockAnalysis({
           dependents: [{ filepath: 'src/a.ts', isTestFile: false }],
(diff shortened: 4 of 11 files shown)


alfhen:
## Both CodeRabbit findings were valid — fixed. Independent verification below.

**Finding 2 (instructions.ts:44) — real, and the same defect class this campaign exists to close.**

```
instructions.ts:44    "grep before concluding the symbol is unused"     ← wrong noun
tools.ts:166          "the file is unused"                              ← correct
get-dependents.ts:41  "a file-level query (no `symbol`)"                ← the actual condition
```

`dependent-attribution-incomplete` only ever fires on file-level queries, so model-facing text was disagreeing with its sibling doc about which parameter the caveat concerns. Corrected, and the file-level-only condition is now stated outright rather than left to inference.

Worth noting how it broke on the first attempt: I wrote `` `symbol` `` in backticks, and `SERVER_INSTRUCTIONS` is a template literal that uses **no backticks anywhere** — the unescaped backtick terminated the string and killed the build. Matched the file's double-quote convention instead.

**Finding 1 (changeset) — premise off, substance right. Changed patch → minor.**

CodeRabbit said the project is at `0.1.0`; it's actually `0.71.1`. But the underlying question stands: this removes three response fields and adds one, and at `0.x` a minor is what signals breaking. Repo precedent agrees decisively — **71 minor releases against 14 patches**.

## Verified independently, on a clean private corpus

The shared corpus had my own T3 trial's uncommitted rename in it mid-session, so I cloned fresh — the same hazard this PR's author hit and correctly routed around.

| Case | Result |
|---|---|
| console `Cursor.php` symbol `Cursor` — the #933 case | `low` → **`high`**, boost `critical` no longer contradicted, reasoning gained *"critical-complexity dependent regardless of test coverage"* |
| gin `StringToBytes` — **regression guard** | **stayed `high`**, not pushed to `critical` |
| `Cursor::__construct` | `high` + `symbol-attribution-degraded` — both fixes compose |
| serilog `Alignment.cs` | `low` + `dependent-attribution-incomplete` |
| nonexistent path | `low` + `unresolved-target` |
| gin `binding/binding.go` — control | `medium`, boost low, no caveat — untouched |

All three caveat reasons fire on the right case and are mutually exclusive. Old flags gone from the response.

## On the one-tier-floor choice

The author's reasoning holds up empirically, and the measure-first order is why. A naive `riskLevel = max(count-level, complexityRiskBoost)` would have pushed gin's `StringToBytes` to `critical` and broken the regression guard — the author found that by measuring gin's real numbers *before* choosing the rule, rather than discovering it in review. Flooring one tier below the boost keeps `critical` reachable only via the existing >20-dependent path while tested+critical caps at `high`, which is the defensible reading: being tested reduces the chance of a *silent* break, not the blast radius.

Also good: it found and fixed a test that **asserted the bug** — *"should keep risk low when complexity is high but all dependents are tested"* — renamed with the expectation flipped. A test encoding the defect is how these survive review.

CodeRabbit genuinely ran on this one (`Review completed`, not rate-limited as on #927/#934/#936/#937), so this is the first PR in the campaign with both reviewers actually reporting.

Issue #904, opened 2026-07-25
Python relative imports (from .module import X) never match anything — matchesPythonModule's regex rejects a leading dot, undercounting flask.globals dependents 5 vs 16 real


alfhen:
## Summary

`matchesPythonModule` in `packages/parser/src/utils/path-matching.ts` gates on `/^[A-Za-z_]\w*(\.[A-Za-z_]\w*)+$/` — the string must *start* with a letter/underscore. Python's relative-import syntax (`from .module import X`, `from ..pkg.module import Y`) is extracted by `PythonImportExtractor` as a **leading-dot** specifier (`.module`, `..pkg.module` — confirmed via chunk metadata below), which fails that regex at the very first character, so `matchesPythonModule` never fires for it. None of `matchesFile`'s other strategies fill the gap either: strategy 3's relative-import handling only strips a `./` or `../` *slash-based* prefix (`normalizedImport.replace(/^(\.\.?\/)+/, '')`), which is JS/TS-shaped and never matches Python's dot-without-slash form, so `cleanedImport === normalizedImport` and nothing changes; strategies 1/2's generic boundary matching can't align a literal `.` character against a `/`-separated target path either, since gitignore/path boundaries are `/`, not `.`. Net result: a leading-dot Python import matches **zero** files, full stop, via any of the six matching strategies.

This is a different mechanism from #901 (bare, dot-free `import pkg` — that string has *no* dot at all) and from #859 (already-fixed raw-statement-text bug) — here the extracted specifier is exactly right (verified below) and even *does* contain a dot, but the regex's start-anchor rejects the leading one that Python's own relative-import syntax requires.

## Repro (pallets/flask)

```
git clone --depth 1 https://github.com/pallets/flask.git
cd flask
lien init --editor claude-code --legacy && lien index -f
```

`src/flask/app.py` imports `globals.py` the idiomatic, PEP-8-recommended way — a relative import:
```python
from .globals import _cv_app, app_ctx, g, request, session
```
Chunk metadata confirms clean, correct extraction:
```sql
sqlite> SELECT imports, importedSymbols FROM chunks WHERE file = 'src/flask/app.py' LIMIT 1;
[..., ".globals", ...] | {..., ".globals": ["_cv_app","app_ctx","g","request","session"], ...}
```
[...]

PR #942, opened 2026-07-29
fix(parser): resolve bare same-directory self-import specifier "." (#935)

### LINKED ISSUE #935: test-associations: bare self-import specifier "." is never resolved, so a same-directory barrel re-export test never matches its target
Found as a side effect while fixing #929 (PR #934) — surfaced once that
fix removed an unrelated false-positive match that had been masking it.

## The bug

A test file that imports its own directory's barrel via the bare current-
directory specifier `'.'` (e.g. `import { x } from '.'`) never gets
associated with the sibling file(s) that barrel re-exports, even when the
import genuinely resolves to exactly the target.

Confirmed on a real indexed honojs/hono corpus:

```
$ grep -nE "^import|from '" src/middleware/jsx-renderer/index.test.tsx
2:import { expectTypeOf } from 'vitest'
3:import { html } from '../../helper/html'
4:import { Hono } from '../../hono'
5:import type { FC } from '../../jsx'
6:import { Suspense } from '../../jsx/streaming'
7:import { jsxRenderer, useRequestContext } from '.'
```

Line 7's `'.'` specifier is a genuine same-directory self-import — the
test directly exercises `src/middleware/jsx-renderer/index.ts`'s own
exports. But the stored chunk metadata for this test file keeps it
completely unresolved:

```
$ sqlite3 structural.db "SELECT DISTINCT file, imports FROM chunks WHERE file LIKE '%jsx-renderer/index.test.tsx%';"
src/middleware/jsx-renderer/index.test.tsx|["vitest","src/helper/html","src/hono","src/jsx","src/jsx/streaming","."]
```

Every other relative import in the same array (`'../../helper/html'`,
`'../../hono'`, etc.) is resolved to a workspace-relative path by the
extractor before it reaches chunk metadata — only the bare `.` is left
as the literal string `"."`.

`annotate` confirms the practical effect — no coverage at all is reported
for the file the barrel re-exports:

```
$ node <cli> annotate src/middleware/jsx-renderer/index.ts
  • No test coverage.
```

Same shape recurs for `src/middleware/secure-headers/index.test.ts`
(`import { NONCE, secureHeaders } from '.'`).

## Why nothing in `matchesFile` catches it

- Strategy 3 (relative-import cleanup) only strips a leading `./` or `../`
  — a bare `.` with no trailing slash doesn't match
  `/^(\.\.?\/)+/`, so `cleanedImport === normalizedImport` and the whole
  branch is skipped.
- Strategy 5 (`matchesPythonModule`)'s guard regex
  (`/^[A-Za-z_]\w*(\.[A-Za-z_]\w*)*$/`) requires the string to start with a
  letter/underscore, so a bare `.` is rejected immediately regardless of
  language.
- No other strategy is relevant to a bare `.`.

So a same-directory self-import via `.` structurally cannot match anything
today, regardless of language.

## Suggested fix direction

This looks like it belongs in the import-extraction/resolution layer (the
TS/JS import extractor, alongside whatever already resolves `'./foo'` and
`'../foo'` to workspace-relative paths), not in `matchesFile` itself — a
bare `.` (and probably bare `./`) should resolve to
`<importer's own directory>/index` the same way Node/TS module resolution
treats it, before it ever reaches chunk metadata. Worth checking whether
other relative-import edge cases (e.g. a bare `..`) have the same gap.

## Repro

Corpus: honojs/hono, already indexed at the time of writing (see #929's
provisioning). Files: `src/middleware/jsx-renderer/index.test.tsx` /
`src/middleware/jsx-renderer/index.ts`,
`src/middleware/secure-headers/index.test.ts` /
`src/middleware/secure-headers/secure-headers.ts` (one level of barrel
re-export removed).

alfhen:
## Summary
- Fixes #935: a bare same-directory self-import specifier (`import { x } from '.'`) was stored verbatim as the literal string `"."` in chunk metadata and never resolved to a path, so a same-directory barrel re-export test never associated with the target it directly imports.
- `resolveRelativeImport` only recognized specifiers starting with `./` or `../`. JS/TS's import extractor stores the raw source text as-is, so a bare `.` (no leading slash) fell straight through unresolved.
- Extended `resolveRelativeImport` to also match the bare, slash-free `.`/`..` themselves via a single anchored regex (`RELATIVE_IMPORT_PATTERN`) — Node/TS module resolution treats them as "this directory"/"the parent directory" exactly like their slash-suffixed forms. This is a narrow fix at the shared import-resolution layer (not `matchesFile`, per the issue's suggested direction), so it also benefits `get_dependents`, which shares the same pipeline.
- Python's own leading-dot relative imports (`.foo`, `..pkg`) are unaffected — `PythonImportExtractor` already converts those to `./`/`../`-prefixed form upstream (#904), so the new bare-dot branch only ever fires for languages (JS/TS today) whose extractor stores the raw specifier verbatim. Verified no leniency was reintroduced for any other language (see dogfood evidence below).
- Simplified the check to a single anchored regex rather than four OR'd string comparisons — `lien delta` actually reports this function's cognitive complexity *improving* (2 → 1) versus HEAD, not worsening.

## Dogfood evidence (real hono corpus, honojs/hono @ 224d2f5)

Built two CLIs — pre-fix (HEAD~1, native parser built) and this branch — and force-reindexed the same corpus copy with each before every measurement (learned the hard way: my first pass measured a stale/broken index and had to be redone from scratch).

**The two files named in the issue, before vs after:**

Before (`node <old-cli> annotate <file>`):
```
Lien impact for src/middleware/jsx-renderer/index.ts:
  • 8 files import this — ...; risk: high (8 callers, max complexity 32, untested high-complexity dependent).
  • No test coverage.

Lien impact for src/middleware/secure-headers/secure-headers.ts:
  • 9 files import this — ...; risk: high (9 callers, max complexity 32, untested high-complexity dependent).
  • No test coverage.
```

After (`node <new-cli> annotate <file>`):
```
Lien impact for src/middleware/jsx-renderer/index.ts:
  • 10 files import this — ...; risk: high (10 callers, max complexity 32, critical-complexity dependent regardless of test coverage).
  • Test coverage: src/middleware/jsx-renderer/index.test.tsx.

Lien impact for src/middleware/secure-headers/secure-headers.ts:
  • 11 files import this — ...; risk: high (11 callers, max complexity 32, critical-complexity dependent regardless of test coverage).
  • Test coverage: src/middleware/secure-headers/index.test.ts.
```

The dependent count also moves (+2 each) — the test files' own bare-`.` self-import is now correctly counted as a dependency edge too, a `get_dependents` accuracy bonus from fixing the shared resolution layer.

**25-file coverage sample, reindexed fresh before each measurement:**

| repo (lang) | before | after | flips |
|---|---|---|---|
| hono (TS) | 13/25 (52%) | 18/25 (72%) | +5, all `none → confident` |
| gin (Go) | 8/25 (32%) | 8/25 (32%) | 0 |
| flask (Python) | 9/25 (36%) | 9/25 (36%) | 0 |

Manually diffed all 5 hono flips against source — every one is a genuine bare-dot self-import newly resolving correctly:
- `context-storage/index.test.ts` / `pretty-json/index.test.ts`: the test file itself does `from '.'`.
- `cloudflare-workers/index.ts`, `jwk/index.ts`, `timeout/index.ts`: a sibling `*.test.ts` in the same directory imports the barrel via `from '.'` (`websocket.test.ts`, `jwk/index.test.ts`, `timeout/index.test.ts` respectively).

Zero flips on Go/Python confirms no cross-language leniency was reintroduced (mirrors #934's guard-scoping precedent).

## Test plan
- [x] `packages/parser/src/utils/path-matching.test.ts`: 3 new unit tests for `resolveRelativeImport`'s bare `.`/`..` handling + an end-to-end `matchesFile` pairing on the real hono shape (133/133 passing)
- [x] `packages/parser/src/ast/symbols.test.ts`: new integration test through the real TS parser + `extractImports` with a filepath, confirming `.` resolves to the importer's directory rather than staying literal (102/102 passing)
- [x] Full gate chain: `format:check`, `lint` (0 errors), `typecheck`, `build`, `test` (parser 378, cli 1296, review 1701, action 41 — all passing), `lien delta` (improved, not worsened)
- [x] Dogfood: real before/after against the honojs/hono corpus (see above) + cross-language regression check (gin, flask)

No new MCP response fields or documented limitations — this is a pure accuracy fix to existing `testAssociations`/`get_dependents` output, invisible at the schema level, so no `tools.ts`/`instructions.ts` doc changes are needed.

---

### Lien Review

> [!NOTE]
> **Low Risk**
>
> This PR fixes a narrow import-resolution bug where bare `.`/`..` specifiers (e.g. `import { x } from '.'`) were stored verbatim instead of being resolved to the importer's directory. The change is localized to `resolveRelativeImport` in `packages/parser/src/utils/path-matching.ts`, replacing the `./`/`../` prefix check with an anchored regex that also matches bare `.`/`..` while still excluding Python-style dotted imports like `.module` or `..pkg.thing`. The implementation correctly resolves the bare specifiers via `path.posix.join`/`normalize` and strips the trailing slash, and the added unit and integration tests cover the exact repro shapes. Documentation in the changeset and code comments accurately describes the behavior and the upstream Python conversion guard.
>
> - Added `RELATIVE_IMPORT_PATTERN` regex to `resolveRelativeImport` so bare `.` and `..` are treated as relative self/parent-directory imports
> - Added unit tests for `.`/`..` resolution and an end-to-end `matchesFile` test using the real hono repro shape
> - Added an integration test through the TypeScript parser confirming `extractImports` resolves `'.'` against the importer path

<sup>Reviewed by [Lien Review](https://lien.dev) for commit e7067f0. Updates automatically on new commits.</sup>

### DIFF
--- a/.changeset/bare-dot-self-import.md
+++ b/.changeset/bare-dot-self-import.md
@@ -1,0 +1,37 @@
+---
+"@liendev/parser": patch
+"@liendev/lien": patch
+---
+
+#935: a bare same-directory self-import specifier (`import { x } from '.'`)
+was never resolved, so a same-directory barrel re-export test never
+associated with the sibling file(s) it directly imports.
+
+Root cause: `resolveRelativeImport` only recognized specifiers starting
+with `./` or `../`. JS/TS's import extractor stores the raw source text
+verbatim, and `import { x } from '.'`'s specifier is the literal string
+`"."` — no leading slash, so it fell through unresolved and was stored in
+chunk metadata exactly as written. None of `matchesFile`'s five strategies
+can match an unresolved `"."` against anything.
+
+Reproduced on a real indexed honojs/hono corpus:
+`src/middleware/jsx-renderer/index.test.tsx` imports its own directory's
+barrel via `from '.'`, and `lien annotate` reported "No test coverage" for
+`src/middleware/jsx-renderer/index.ts` despite the test directly exercising
+it. Same shape on `src/middleware/secure-headers/index.test.ts`.
+
+Fix: `resolveRelativeImport` now also matches the bare, slash-free `.`/`..`
+themselves (Node/TS module resolution treats them as "this directory" and
+"the parent directory", exactly like their slash-suffixed forms) via a
+single anchored regex, `RELATIVE_IMPORT_PATTERN`. This resolves to the
+importer's own directory (or its parent) the same way an already-supported
+`./` /`../` specifier does, so downstream matching needs no changes.
+Python's own leading-dot relative imports (`.foo`, `..pkg`) are unaffected —
+`PythonImportExtractor` already converts those to the `./`/`../`-prefixed
+form before this function runs (#904), so the new bare-dot case only ever
+fires for languages (JS/TS today) whose extractor stores the raw specifier
+as-is.
+
+Verified end-to-end on the real hono repro (both files now report their
+own test file under "Test coverage") and via a 25-file coverage sample
+across hono, gin, and flask confirming no regression.
--- a/packages/parser/src/ast/symbols.test.ts
+++ b/packages/parser/src/ast/symbols.test.ts
@@ -641,6 +641,24 @@ from typing import Optional
       expect(imports).toContain('os');
       expect(imports).toContain('typing');
     });
+
+    it('resolves a bare same-directory self-import against the importer path (#935)', () => {
+      // `import { jsxRenderer } from '.'` -- the honojs/hono repro. The raw
+      // TS extractor stores the source text verbatim ("."), so resolution to
+      // the importer's own directory only happens when a filepath is passed
+      // through to resolveImportSpecifier/resolveRelativeImport.
+      const content = `import { jsxRenderer } from '.';`;
+
+      const parseResult = parseAST(content, 'typescript');
+      const imports = extractImports(
+        parseResult.tree!.rootNode,
+        'typescript',
+        'src/middleware/jsx-renderer/index.test.tsx',
+      );
+
+      expect(imports).toContain('src/middleware/jsx-renderer');
+      expect(imports).not.toContain('.');
+    });
   });
 
   describe('extractImports - Go grouped imports (#863)', () => {
--- a/packages/parser/src/utils/path-matching.test.ts
+++ b/packages/parser/src/utils/path-matching.test.ts
@@ -997,6 +997,30 @@ describe('resolveRelativeImport', () => {
     expect(bResolved).toBe('packages/parser/src/dependency-analyzer');
     expect(aResolved).not.toBe(bResolved);
   });
+
+  it('resolves a bare same-directory self-import (#935)', () => {
+    // `import { x } from '.'` in JS/TS — the extractor stores the raw source
+    // text verbatim (unlike Python, which converts its own bare-dot form to
+    // "./" before this function ever runs), so a literal "." must resolve the
+    // same way "./" already does: to the importer's own directory, with
+    // nothing joined after it.
+    expect(resolveRelativeImport('src/middleware/jsx-renderer/index.test.tsx', '.')).toBe(
+      'src/middleware/jsx-renderer',
+    );
+  });
+
+  it('resolves a bare parent-directory self-import (#935)', () => {
+    expect(resolveRelativeImport('src/middleware/jsx-renderer/nested/foo.ts', '..')).toBe(
+      'src/middleware/jsx-renderer',
+    );
+  });
+
+  it('matches the real hono repro end-to-end via matchesFile (#935)', () => {
+    // src/middleware/jsx-renderer/index.test.tsx: `import { jsxRenderer } from '.'`
+    // must associate with src/middleware/jsx-renderer/index.ts.
+    const resolved = resolveRelativeImport('src/middleware/jsx-renderer/index.test.tsx', '.');
+    expect(matchesFile(resolved, 'src/middleware/jsx-renderer/index')).toBe(true);
+  });
 });
 
 describe('resolveWorkspaceImport', () => {
--- a/packages/parser/src/utils/path-matching.ts
+++ b/packages/parser/src/utils/path-matching.ts
@@ -631,33 +631,50 @@ function matchesPHPNamespace(importPath: string, targetPath: string): boolean {
 }
 
 /**
+ * Matches a relative import specifier: `./x`, `../x`, or the bare, slash-free
+ * `.`/`..` themselves (anchored so a longer dotted specifier like `.module`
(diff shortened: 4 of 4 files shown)


alfhen:
**Merging.** Verified independently, and reporting one unexplained discrepancy rather than adopting the PR's number.

## Confirmed

Built this branch **with `build:native`** and sanity-checked that parsing actually works before trusting any number — `Indexing complete (378/378)`, native binding present. That check exists because of the trap the author hit and reported: a scratch "before" CLI built without native silently failed to parse *every* file, producing plausible-looking numbers from a completely broken index.

Both repro files resolve correctly now:

```
src/middleware/jsx-renderer/index.ts
  • Test coverage: src/middleware/jsx-renderer/index.test.tsx.
src/middleware/secure-headers/secure-headers.ts
  • Test coverage: src/middleware/secure-headers/index.test.ts.
```

Controls — **exact match to baseline, zero flips**, as claimed:

| repo | baseline | mine |
|---|---|---|
| gin | 8/25 | **8/25** |
| flask | 9/25 | **9/25** |

That matters most here: a fix to shared import resolution could easily have leaked into other languages, and it didn't. The author's reasoning for why is sound — Python's leading-dot imports are converted to `./`/`../` by `convertPythonRelativeImport` upstream, so the new bare-dot branch only ever fires for JS/TS, and no gated language parameter was needed.

## Unexplained discrepancy — flagging, not blocking

hono: baseline 13/25. The PR reports **18/25**; I measure **16/25**.

I checked the obvious cause and it isn't that — my corpus copy and the shared one are byte-identical (both clean, both at `224d2f5`, both 324 `.ts` files), and I used this branch's own `provision.mjs` and `mcp-call.mjs`. So identical inputs, different results, cause unknown.

It doesn't change the merge decision: direction agrees, both are improvements, the two repro files verify directly, and the controls are clean. But given this campaign produced nine measurement failures — one of them the author's own broken-native run in this very PR — I'd rather record an unresolved +3-vs-+5 than quietly adopt whichever number reads better. If anyone re-measures, mine is the conservative figure.

`patch` bump is right — this is import-resolution behaviour, not a response-shape change, matching #929/#934's precedent for this subsystem.

Checks green, Lien Review Good, 0 inline. CodeRabbit rate-limited again (recorded, not read as coverage).

## Process note for the fleet

The author flagged that this worktree's branch was silently swapped mid-session. The reflog shows what happened: the harness checked out **another agent's** branch (`fix-csharp-typeref-dependents`) into this worktree, this PR's commit landed on it, and it was then renamed to `fix-935-bare-dot-self-import`. I verified this PR's diff contains only #935's files, so nothing was absorbed — but the other agent's branch ref is gone and its work was left uncommitted in a third worktree. It's been told where to find it. Worth knowing that concurrent agents can silently inherit each other's branch checkouts.

PR #943, opened 2026-07-29
fix(parser,cli): recover C# type-reference dependents for get_dependents (#930 part 2)

alfhen:
## Summary

Part 2 of #930. #932/#936/#938 stopped `get_dependents` from fabricating
dependents out of C#'s `GlobalUsings.cs` boilerplate and made the resulting
zero honest (`dependentAttributionIncomplete`) — but a file with `global
using` in scope still reported `dependentCount: 0` / `riskLevel: "low"` even
when it had real callers. Honest-and-blind is better than confidently-wrong,
and worse than correct.

- Adds `findCSharpTypeReferenceDependents` (`@liendev/parser`): for a target
  file whose type name is declared exactly once project-wide (so a
  same-named reference elsewhere can't be an unrelated declaration — the C#
  compiler itself would refuse to build over that ambiguity), scan every
  other C# file's source text for an identifier-boundary occurrence of that
  name. Only attempted when the import graph found zero dependents for a
  file-level query on an `enclosingNamespaceAccess` language (C# today) —
  mirrors the tier-1/tier-2 discipline from #926/#923's Go/Java/Swift
  same-shaped signals, but recovers *dependents* (`get_dependents`), not
  test-associations (which stay `lien annotate`-only).
- Recovered entries are tagged `confidence: "inferred"` on `DependentInfo` —
  a new optional field, never folded in unhedged next to an import-verified
  dependent.
- Adds a new `attributionCaveat` reason, `"dependent-attribution-partial"`,
  so callers know the count is a recovered lower bound, not a
  verified/complete answer. `dependentAttributionIncomplete` now only fires
  when this fallback *also* finds nothing.
- Purely additive: no existing field removed, renamed, or reshaped.
- `tools.ts` and `instructions.ts` (the two model-facing surfaces) updated
  and verified against the built bundle.

### Design judgment

1. **Extension point**: a new signal-generator module
   (`csharp-type-reference-signals.ts`) parallel to
   `go-same-directory-tests.ts`/`java-same-package-tests.ts`/
   `swift-symbol-usage-signals.ts`, but wired into the CLI's
   `dependency-analyzer.ts` (`get_dependents`) instead of
   `test-associations.ts` — a deliberate precedent break, since #930's
   remaining gap is specifically that `get_dependents` itself (not just
   `annotate`) reports a false all-clear.
2. **Type reference detection**: no new AST extraction needed. Existing
   symbol extraction already records each type declaration's name
   (`symbolType: 'class' | 'interface'` + `symbolName`); a project-wide
   uniqueness check on that name, combined with an identifier-boundary text
   scan of `chunk.content`, is the cheapest sound approach — verified
   against a real serilog clone with zero false positives project-wide
   (`grep -rlw` for both `Alignment` and `Padding`, not just the 5 known
   files).
3. **Confidence tiering**: mirrors #926 — recovered dependents are tagged
   `confidence: "inferred"`, never merged unhedged. Unlike Swift's
   call-site signal, no extra "type-shaped driver" gate is needed: every
   candidate here is already a real type declaration by construction, never
   a method/property name, so that collision class doesn't apply the same
   way.
4. **`attributionCaveat`**: `dependentAttributionIncomplete` (zero signal at
   all) and the new `dependentAttributionPartial` (fallback recovered
   something) are mutually exclusive by construction — the former requires
   the *final* `dependents.length` to be zero, the latter requires it to be
   positive.

## Verification (real MCP tool, fresh serilog/serilog clone, own build)

`src/Serilog/Parsing/Alignment.cs` — before (main): `dependentCount: 0`,
`riskLevel: "low"`. After:
```
dependentCount: 6 | riskLevel: medium
dependents: MessageTemplateTextFormatter.cs, MessageTemplateParser.cs,
            PropertyToken.cs, MessageTemplateRenderer.cs, Padding.cs,
            MessageTemplateParserTests.cs
caveat: dependent-attribution-partial
```
All 5 real production dependents + the real test dependent recovered, all
tagged `confidence: "inferred"`.

`src/Serilog/Parsing/PropertyToken.cs` (control — genuine import-verified
test dependent) — unaffected: `dependentCount: 1`, `riskLevel: low`, its one
real dependent carries no `confidence` field, no `attributionCaveat`.

Non-C# control (hono, fresh clone) — `src/router.ts`: 42 dependents via
ordinary import matching, zero `confidence` tags, no caveat — zero
cross-language impact.

25-file serilog sample: 21/25 recovered real dependents via the new signal
(risk correctly escalated from `low` to `medium`/etc.), 2/25 honestly left
`dependent-attribution-incomplete` (no recoverable signal), 2/25 already had
real import-based dependents (untouched), 0 errors.

## Gates

`format:check` / `lint` (0 errors) / `typecheck` / `build` / `test` (all
packages, 1457+378+1701+1302+41+27 = green) / `lien delta --base origin/main`
(1 crossing found and fixed via extraction into `resolveDependents`, exit 0
on the final run) — all run in a fresh worktree with its own `npm install` +
`build:native`.

## Test plan

- [x] New unit tests for `findCSharpTypeReferenceDependents` (11 cases,
      including a full reproduction of the 5-dependent serilog fixture)
- [x] New CLI tests for the enrichment gate + `attributionCaveat` branch
- [x] Full monorepo test suite green
- [x] `lien delta` clean
- [x] Live MCP tool verification against a real, freshly-cloned C# repo
      (before/after, control, non-C# control, 25-file sample)

## Summary by CodeRabbit

- **New Features**
  - Improved C# dependent-file detection when `global using` prevents import-graph attribution.
  - Recovered dependents are identified through type references and marked as inferred.
  - Added a `dependent-attribution-partial` caveat to clarify when results are lower-bound estimates.

- **Bug Fixes**
  - Fixed cases where valid C# dependents were incorrectly reported as zero.
  - Preserved incomplete-attribution warnings when recovery finds no matches.

- **Documentation**
  - Updated `get_dependents` guidance to explain inferred matches, confidence, and attribution caveats.

---

### Lien Review

> [!NOTE]
> **Low Risk**
>
> This PR adds a C# type-reference fallback to recover dependents that `global using` hides from the import graph. The new `findCSharpTypeReferenceDependents` signal scans uniquely-declared C# type names across the corpus and tags recovered dependents with `confidence: 'inferred'`. The CLI wires this into `get_dependents` only when the import graph returns zero file-level dependents for an `enclosingNamespaceAccess` language, and surfaces a new `dependent-attribution-partial` caveat. The implementation is additive, well-gated, and consistent with the documentation: the C# parser collapses struct/record/enum declarations to `symbolType: 'class'`, so the implementation's class/interface check does cover all five declaration kinds mentioned in the docs. No breaking changes or logic errors were found.
>
> - Added `findCSharpTypeReferenceDependents` parser signal for C# type-reference-based dependent recovery
> - Wired the fallback into CLI `get_dependents` with `confidence: 'inferred'` and `dependent-attribution-partial` caveat
> - Updated model-facing `tools.ts` and `instructions.ts` to describe the new caveat and confidence tier

<sup>Reviewed by [Lien Review](https://lien.dev) for commit 7044b20. Updates automatically on new commits.</sup>

---

Attested: degraded:budget_starved · 669K/736K tokens

### DIFF
--- a/.changeset/csharp-type-reference-dependents.md
+++ b/.changeset/csharp-type-reference-dependents.md
@@ -1,0 +1,43 @@
+---
+"@liendev/parser": patch
+"@liendev/lien": patch
+---
+
+Recover real C# `get_dependents` dependents lost to `global using` (#930,
+part 2). #932/#936 stopped the tool from fabricating dependents out of
+`GlobalUsings.cs` boilerplate and made the resulting zero honest
+(`dependentAttributionIncomplete`), but a file with `global using` in scope
+still reported `dependentCount: 0` / `riskLevel: "low"` even when it had
+real callers — honest-and-blind, not correct. Confirmed on a fresh
+serilog/serilog clone: `Alignment.cs` has 5 real production dependents and
+1 real test dependent, none reachable via a per-file import.
+
+Adds a lower-confidence recovery signal, `findCSharpTypeReferenceDependents`
+(`@liendev/parser`): for a file whose type name is declared exactly once
+project-wide (so a same-named reference elsewhere can't be an unrelated
+declaration — the C# compiler itself would refuse to build over that
+ambiguity), scan every other C# file's source text for an
+identifier-boundary occurrence of that name. Only attempted when the import
+graph found zero dependents for a file-level query on an
+`enclosingNamespaceAccess` language (C# today).
+
+Recovered dependents are tagged `confidence: "inferred"` on `DependentInfo`
+(a new optional field, absent on every ordinary import-verified dependent)
+and never folded in unhedged — the response also gets a new
+`attributionCaveat` reason, `"dependent-attribution-partial"`, explaining
+that the count is a recovered lower bound, not a verified/complete answer.
+`dependentAttributionIncomplete` now only fires when this recovery attempt
+*also* finds nothing. Both are purely additive: no existing field is
+removed, renamed, or changed shape.
+
+Verified end-to-end via the real MCP `get_dependents` tool against a fresh
+serilog/serilog clone: `Alignment.cs` goes from `dependentCount: 0` /
+`"low"` to all 5 real production dependents + its test dependent (`medium`
+risk); `PropertyToken.cs` (a file with a genuine import-verified test
+dependent) is unaffected; a 25-file serilog sample recovered dependents for
+21 files, left 2 honestly `dependent-attribution-incomplete` (no
+recoverable signal), and left 2 real import-based hits untouched; a
+TypeScript control (hono) confirmed zero cross-language impact.
+
+`tools.ts` and `instructions.ts` (the two model-facing surfaces) are
+updated accordingly.
--- a/packages/cli/src/mcp/handlers/dependency-analyzer.test.ts
+++ b/packages/cli/src/mcp/handlers/dependency-analyzer.test.ts
@@ -14,14 +14,18 @@ function createChunk(
     complexity: number;
     callSites: Array<{ symbol: string; line: number }>;
     symbolName: string;
+    symbolType: 'function' | 'method' | 'class' | 'interface';
+    content: string;
     startLine: number;
     endLine: number;
   }> = {},
 ): SearchResult {
   return {
-    content: overrides.callSites
-      ? overrides.callSites.map(cs => `  ${cs.symbol}()`).join('\n')
-      : 'test content',
+    content:
+      overrides.content ??
+      (overrides.callSites
+        ? overrides.callSites.map(cs => `  ${cs.symbol}()`).join('\n')
+        : 'test content'),
     metadata: {
       file,
       startLine: overrides.startLine ?? 1,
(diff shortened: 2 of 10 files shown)


alfhen:
**Merging.** Verified against independently-derived ground truth, and the one inline finding resolved.

## The recovery is exactly right — zero false positives

I established ground truth myself with `grep -rlw Alignment src --include='*.cs'` before looking at the tool's answer:

```
src/Serilog/Formatting/Display/MessageTemplateTextFormatter.cs
src/Serilog/Rendering/MessageTemplateRenderer.cs
src/Serilog/Rendering/Padding.cs
src/Serilog/Parsing/PropertyToken.cs
src/Serilog/Parsing/MessageTemplateParser.cs
```

`get_dependents` on this branch returns **all five**, plus the genuine test user, every one tagged `confidence: "inferred"`, `attributionCaveat.reason: dependent-attribution-partial`, risk escalated `0/low` → `6/medium`. Before, on main: `dependentCount: 0, riskLevel: "low"`.

Controls hold:
- `PropertyToken.cs` — 1 import-verified dependent, **no** `confidence` tag, **no** caveat. The fallback correctly did not fire on a file that already had a real edge.
- hono `src/utils/jws.ts` — 14 dependents, **0** confidence-tagged, no caveat. Zero cross-language impact.
- `dependent-attribution-partial` present in `tools.ts`, `instructions.ts`, and the built bundle.

## The inline finding was a false positive — and worth documenting anyway

CodeRabbit read `isCSharpTypeDeclarationChunk` as narrower than its doc comment: checking only `'class'|'interface'` while claiming class/struct/record/enum/interface. Reasonable inference, wrong conclusion — `csharp.ts` collapses all five declaration node types into one case, so they all surface as `'class'`.

Checked empirically rather than by reading: serilog's `public readonly struct Alignment` reports `symbolType: 'class'`, which is *why* its dependents are recovered at all. If the reviewer had been right, this PR's headline case could not have worked.

I've added the explanation instead of changing working code, plus something the check itself made visible: in that same corpus `Alignment` appears **twice more** with `symbolType: 'method'` — properties on `PropertyToken` and on the struct. Restricting the declaration set to types is therefore what makes the "exactly one file declares this name" uniqueness gate function; without it those two would break uniqueness and *suppress* a real recovery. The narrow filter is load-bearing, not a limitation.

## On the design judgment

Three calls I'd have made the same way:

- **Breaking the Go/Java/Swift precedent deliberately** and wiring into `dependency-analyzer.ts` rather than `test-associations.ts` — correct, because #930's remaining gap is specifically that `get_dependents` reports the false all-clear, so recovery has to land on that path.
- **The `wordBoundaryRe` bug found en route** is the good kind of finding: reusing `doc-reference-matching.ts`'s pattern treats `.` as identifier-continuation (right for markdown path prose), which silently failed to match `Alignment.Apply(...)` — the dominant real usage shape. That would have produced a plausible partial recovery rather than an obvious failure.
- **Scope honestly stated**: the fallback only runs when the import graph found *zero* dependents, so a mixed old/new-style C# project can still miss recoverable dependents on a file that already has one real edge. Documented as a widening point rather than swallowed.

Checks green, Lien Review Good, CodeRabbit completed (actually ran, unlike most of this campaign).

PR #954, opened 2026-07-29
fix(cli): stop name-filtered test runs from silently disabling the verification nudge

alfhen:
## The bug

`classifyTestCommand` (`packages/cli/src/utils/test-run-matcher.ts`) had exactly
two outcomes: `scoped` (a path-like token found after the runner keyword) or
`broad` (none found). A run scoped by test **NAME** rather than file/directory
(`pytest -k expr`, `dotnet test --filter expr`, `rspec -e name`, `mocha --grep
pattern`, `go test -run regex`, a bare `cargo test name`, `vitest`/`jest -t`)
has no path-like token at all, so it fell into `broad` by construction.
`computeUnverifiedFiles` treats any `broad` run as "the whole edit set was
exercised," so one arbitrarily-named, completely unrelated test run silently
marked every edited file in the session as verified — the false-"tests ran"
direction, which disables the Stop-hook nudge (`recap-stop.sh`) invisibly.

Confirmed on the released 0.72.0 (see dogfood evidence below): an edit with no
test run at all correctly nags; the identical edit followed by
`pytest -k test_totally_unrelated_name` goes completely silent.

## Which runners actually reproduce (verified, not assumed)

All 8 named in the bug report reproduce, confirmed by direct calls to
`classifyTestCommand` before the fix:

| Command | Before | After |
| --- | --- | --- |
| `pytest -k test_totally_unrelated_name` | `broad:true` | `broad:false` |
| `dotnet test --filter FullyQualifiedName~Some.Unrelated.Test` | `broad:true` | `broad:false` |
| `rspec -e "some unrelated example"` | `broad:true` | `broad:false` |
| `mocha --grep 'unrelated pattern'` | `broad:true` | `broad:false` |
| `go test -run TestUnrelated` | `broad:true` | `broad:false` |
| `cargo test test_unrelated_name` | `broad:true` | `broad:false` |
| `vitest -t "unrelated name"` | `broad:true` | `broad:false` |
| `jest -t "unrelated name"` | `broad:true` | `broad:false` |

Compound case (`pytest -k name tests/test_helpers.py`) correctly stays
scoped-by-path in both before and after — a path present must not be
downgraded by a name filter, which it wasn't and still isn't.

**Bonus find, same underlying bug**: `swift test --filter HTTPHeadersTests`
also reproduced (`--filter` was globally invisible via
`WORKSPACE_SCOPE_FLAGS`), fixed by the same mechanism. Updated the one
existing unit test that pinned this as `broad:true`.

## The model chosen, and why

**Option (a)** from the brief: a name-filtered run is a third classification
outcome — "ran, scope unknown" — that is neither `broad` nor a `scopeTokens`
contributor. `computeUnverifiedFiles` only has ONE call site that branches on
`.broad` (itself), so this needed no interface change to `TestRunClassification`
or `computeUnverifiedFiles` at all: a run that's `{isTestRun: true, broad:
false, scopeTokens: []}` already behaves exactly like "no run observed" for
coverage purposes, for free.

Recognition is a **per-runner-family flag allow-list**
(`PYTEST_NAME_FILTER_FLAGS`, `RSPEC_NAME_FILTER_FLAGS`, etc. in
`test-run-matcher.ts`), not one global set, and `cargo test`'s bare positional
argument is special-cased to exactly that runner
(`POSITIONAL_NAME_FILTER_RUNNERS`). Two collisions a naive global/generic
implementation would have hit, both now pinned as regression tests:

- **tox's `-e ENV`** (environment selector, must stay `broad`) vs. **rspec's
  `-e NAME`** (example filter) — identical spelling, unrelated meaning. A
  global `-e` flag would have wrongly flipped `tox -e py311` to name-filtered.
- **`cargo nextest run`**'s required `run` subcommand keyword, and Gradle's
  `-x <task>`/`--exclude-task <task>` exclusion value — both bare tokens in
  the same textual position a name filter's value would sit, but with
  unrelated meaning. A generic "any bare leftover token is a name filter"
  rule (my first draft) would have wrongly reclassified both.

Option (c) (matching the filter expression against real associated test
names) is out of scope — the brief called it "probably out of scope" and I
agree: the filter is a plain-text expression/regex evaluated by the target
runner (e.g. pytest's `-k` supports Python boolean expressions over test
names/markers), not a simple string to compare against file names, and
getting it right per-ecosystem is a much larger project for a small marginal
precision gain over the "just don't drop the nag" fix here.

## What's out of scope / not attempted

- Maven's `-Dtest=ClassName` filter uses the same failure mode (starts with
  `-D`, so it's silently skipped as a flag today) but wasn't in the brief's
  runner list and isn't fixed here.
- Any deeper file-name-matching against the filter expression's own text
  (option c above).

## Verification

- Unit tests: `packages/cli/src/utils/test-run-matcher.test.ts` — 151 tests
  (was 124), covering: the flag form and equals-form for each of the 8
  runners, the compound name+path case, the swift `--filter` fix, the
  tox/cargo-nextest/gradle negative-collision regressions, and a new
  `computeUnverifiedFiles` describe block pinning that a name-filtered run
  contributes no coverage (alone, alongside a real scoped run, and alongside
  a genuinely broad run).
- All six gates green locally: `format:check`, `lint` (pre-existing warnings
  only, 0 errors), `typecheck`, `build`, `test` (1461+378+1701+1332+41+27,
  all packages), `lien delta` (exit 0 — two functions "worsened" in cognitive
  complexity, none crossed a threshold).

### Dogfood evidence (mandatory, foreign repo, real hook stdin)

Run against `/tmp/lien-dogfood-460173c9/flask` (clean working tree,
re-indexed with `index --force` before measuring), driven through the real
`test-reminder.sh`/`test-run-note.sh` hook scripts with the true PostToolUse
stdin shape — not just direct CLI calls for the Bash-command classification
step:

```
# CONTROL -- edit recorded, no test run at all: correctly nags
$ lien verify-tests note-edit --session S1 --file src/flask/helpers.py
Lien: you changed src/flask/helpers.py -- associated tests: tests/test_helpers.py, ...
$ lien verify-tests report --session S1
Before finishing: these files you edited this session have associated tests I
did not observe running in a Bash command:
  - src/flask/helpers.py -> tests/test_helpers.py (+35 more)
...

# BEFORE (pre-fix build, same repo, same commands): silent -- bug reproduced
$ lien verify-tests note-edit --session before --file src/flask/helpers.py
Lien: you changed ... (same as above)
$ echo '{"session_id":"before","tool_name":"Bash","tool_input":{"command":"pytest -k test_totally_unrelated_name"},"cwd":"..."}' | test-run-note.sh
$ lien verify-tests report --session before
(completely silent)

# AFTER (fixed build, same repo, same commands): nags -- bug fixed
$ lien verify-tests note-edit --session after --file src/flask/helpers.py
Lien: you changed ... (same as above)
$ echo '{"session_id":"after","tool_name":"Bash","tool_input":{"command":"pytest -k test_totally_unrelated_name"},"cwd":"..."}' | test-run-note.sh
$ lien verify-tests report --session after
Before finishing: these files you edited this session have associated tests I
did not observe running in a Bash command:
  - src/flask/helpers.py -> tests/test_helpers.py (+35 more)
...
```

The "before" build was produced by stashing this PR's diff, rebuilding
`packages/cli`, running the repro, then popping the stash and rebuilding
again -- not the npm-published binary, because concurrent agents in this
session had corrupted the shared npx cache (`npx -y @liendev/lien@0.72.0`
was resolving to a stale 0.65.0 despite the registry confirming 0.72.0 is
current). This gives an equally valid, more controlled A/B since it isolates
exactly this diff.

Regression checks on the fixed build, same repo:
- Genuinely broad run (`python -m pytest`, no filter) -> `report` still silent
  (correctly suppresses).
- File-scoped run naming an unrelated file (`pytest
  tests/test_totally_unrelated_module.py`) -> `report` still nags (correctly
  does NOT suppress).

Flask repo left clean (`git status --porcelain` empty) -- no repo files were
touched, only `lien` CLI state under `~/.lien`.

## Note on Lien MCP tool usage

Per CLAUDE.md, ran `get_files_context` before editing and `get_dependents`
before touching the exported `classifyTestCommand`/`TestRunClassification`/
`computeUnverifiedFiles` symbols. The `plugin:lien` MCP server itself
returned a stale index throughout this session (pinned to a different
branch/commit, unaffected by `lien index --force` run from this worktree --
a known multi-agent-fleet quirk, not something this PR could fix), so
dependent analysis was cross-checked manually via `grep` for
`classifyTestCommand`/`computeUnverifiedFiles`/`.broad` call sites across
`packages/`; only `verify-tests-cmd.ts` and `recap-cmd.ts` (import-only)
consume these exports, both unaffected by the interface (no signature
change).

---

### Lien Review

> [!NOTE]
> **Low Risk**
>
> This PR fixes a real bug where name-filtered test commands (e.g., `pytest -k`, `go test -run`, `cargo test <name>`) were misclassified as broad runs, causing the test-verification nudge to be silently disabled. The fix introduces a third classification outcome — name-filtered (`isTestRun: true, broad: false, scopeTokens: []`) — which `computeUnverifiedFiles` already treats as "no coverage evidence," so no caller interface changes were required. The implementation uses per-runner-family flag allow-lists and runner-specific value-skip sets to avoid known collisions (tox `-e` vs. rspec `-e`, cargo-nextest's `run` keyword, cargo build-config flags, pnpm workspace `--filter`). The change is extensively regression-tested, including negative guards and the disputed scope-broadening cases (`go test -run Foo ./...`, `cargo test foo --workspace`), which the author has documented as intentionally name-filtered per the fail-safe bias. No code defects were found.
>
> - Name-filtered runs now return `{isTestRun: true, broad: false, scopeTokens: []}` instead of `{broad: true}`, preventing false suppression of the verification nudge.
> - Added per-runner-family name-filter flag detection and cargo-test-specific positional name filter handling with value-skip sets for cargo and pnpm.
> - Public `TestRunClassification` interface unchanged; callers `verify-tests-cmd.ts` and `recap-cmd.ts` require no updates.

<sup>Reviewed by [Lien Review](https://lien.dev) for commit 83dd774. Updates automatically on new commits.</sup>

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved test-run detection for name-filtered commands such as `pytest -k`, `jest -t`, and similar options.
  * Prevented filtered test runs from being incorrectly treated as full-suite runs or suppressing verification advisories.
  * Corrected handling of value-taking `cargo test` options to avoid misclassifying their values as test scopes.

* **Documentation**
  * Updated guidance and examples for test-run verification behavior, including end-to-end scenarios.

### DIFF
--- a/docs/architecture/test-verification-nudge.md
+++ b/docs/architecture/test-verification-nudge.md
@@ -276,6 +276,61 @@ vs. rspec `-e`, cargo-nextest's `run` keyword vs. `cargo test`'s positional,
 and now `cargo test`'s own compile-config flags vs. its own positional. All
 three are now pinned as regression tests in `test-run-matcher.test.ts`.
 
+##### A fourth instance: pnpm's own `--filter`/`-F` workspace selector (2026-07-29)
+
+Review of the cargo fix above found one more case of the same hazard, then a
+second, independent review of *that* claim corrected its direction: `pnpm
+test --filter <selector>` (the workspace selector placed AFTER the script
+name, rather than the `pnpm --filter <selector> test` form already absorbed
+whole by its own anchored `RUNNER_PATTERNS` entry) has a selector value that
+routinely contains `/` for a scoped package name (`@hono/core`). Without a
+dedicated skip, that value was independently scanned by the generic
+path-token check and misread as a real scope-narrowing FILE — flipping a
+genuinely broad run to a falsely narrow "scoped to `./@hono/core`" (the
+review's first pass called this a suppression risk; a second, independently
+verified pass established the direction is actually the opposite — a false
+NAG, not a false silence, since a broad-run-misread-as-scoped can only ever
+*reduce* claimed coverage, never expand it). Fixed by adding
+`PNPM_TEST_VALUE_FLAGS` (`--filter`, `-F`), skipped as flag+value via
+`valueSkipFlagsFor` exactly like `CARGO_TEST_VALUE_FLAGS`, but gated on the
+generic `pnpm test`/`pnpm test:script` match specifically (`isPnpmTestRunner`)
+so dotnet/swift's unrelated test-NAME use of the identical `--filter`
+spelling is untouched.
+
+The same review also surfaced, independently of the misclassification bug,
+that the `--filter=selector` single-token form placed BEFORE the script
+(`pnpm --filter=@hono/core test`) matched no `RUNNER_PATTERNS` entry at all —
+pnpm's own docs show this `=` form as the canonical way to write an
+exclusion selector (`--filter=!foo`), so a real, documented, non-exotic
+invocation always nagged. Fixed by widening the dedicated anchored pnpm
+pattern to `(?:--filter|-F)(?:\s+\S+|=\S+)` (both the space and `=` forms,
+both the long flag and its documented short alias).
+
+A sweep of every other per-runner name-filter flag's `=` form (`pytest
+-k=`/`-m=`, `rspec -e=`/`--example=`, `mocha --grep=`/`-g=`, `go test
+-run=`, `vitest`/`jest -t=`/`--testNamePattern=`, `dotnet`/`swift
+--filter=`) found no equivalent gap: `isNameFilterFlag`'s existing
+`token.startsWith(`${flag}=`)` check already recognizes all of them —
+verified directly against `classifyTestCommand`, all still correctly
+name-filtered. A separate finding while investigating (`turbo test
+--filter=web`) is NOT a `--filter` bug at all: `turbo` is not a recognized
+runner in `RUNNER_PATTERNS` at all, so the command is simply unclassified
+and falls through to the safe fail-open default (nags because no run was
+ever recorded) — unrelated to this fix and out of scope (adding `turbo`
+support was never requested and isn't a regression).
+
+This is the **fourth** instance of the identical short-flag/value-collision
+hazard class in this file. A generalized "flag arity table" (mapping every
+recognized flag per runner to a consumes-value/semantics tag) was considered
+in place of a fourth targeted allow-list entry, but rejected for now: the
+existing per-concern, per-runner-family lookup idiom (`nameFilterFlagsFor`,
+`valueSkipFlagsFor`) already generalizes to this fourth case with one small,
+independently testable addition each, matching this module's stated
+conservative-allow-list philosophy; a unified table would be a structural
+reorganization with no additional correctness benefit over what's here,
+and would cost a larger, riskier diff for a module already this
+security-sensitive to the nudge's credibility.
+
 ## C. `lien verify-tests <subcommand>`
 
 A command group (`packages/cli/src/cli/verify-tests-cmd.ts`), like `lien
--- a/docs/architecture/test-verification-nudge.md
+++ b/docs/architecture/test-verification-nudge.md
@@ -137,7 +137,7 @@ interface TestRunClassification { isTestRun: boolean; broad: boolean; scopeToken
    or ends in a source extension (reusing `getSupportedExtensions()` from
    `@liendev/parser` rather than hand-maintaining a second extension list —
    a test file shares its language's extension, so "ends in a source
-   extension" already covers the test-ish case). Two token classes are
+   extension" already covers the test-ish case). Three token classes are
    excluded from ever counting as a scoping argument even when they'd
    otherwise qualify: a workspace-scope flag's value (`-w`/`--workspace`,
    e.g. `@liendev/core`, which contains a `/`), and a config-flag's value
@@ -330,6 +330,64 @@ conservative-allow-list philosophy; a unified table would be a structural
 reorganization with no additional correctness benefit over what's here,
 and would cost a larger, riskier diff for a module already this
 security-sensitive to the nudge's credibility.
+
+##### A scope-broadening flag does NOT make a name-filtered run `broad` — selection and scope are independent axes (2026-07-30)
+
+A review pass suggested that `go test -run TestFoo ./...` /
+`go test -run TestFoo .` and `cargo test foo --workspace` /
+`cargo test foo --all` should classify as `broad`, on the reasoning that
+`./...`/`--workspace`/`--all` are "broad-scope indicators." **This is wrong,
+and the classifier's existing behavior (no change needed) is correct** —
+verified directly against `classifyTestCommand` and independently re-verified
+against a real Go module (gin) with `internal/bytesconv` (4 test functions,
+none named `TestFoo`):
+
+```
+go test -run TestFoo ./...      => name-filtered (NAGS)  -- executes none of bytesconv's 4 tests
+go test -run TestFoo .          => name-filtered (NAGS)
+cargo test foo --workspace      => name-filtered (NAGS)
+cargo test foo --all            => name-filtered (NAGS)
+go test ./...                   => broad (SILENT)  -- no name filter, real whole-suite coverage
+cargo test --workspace          => broad (SILENT)
+cargo test --all                => broad (SILENT)
+```
+
+The reasoning: **which packages/crates get compiled and which tests within
+them actually execute are two independent axes.** `./...`/`--workspace`/
+`--all` answer the first question (compile everything reachable) — genuinely
+scope-broadening, correctly contributing no scoping restriction on their
+own. `-run TestFoo` / a bare positional test name answer the second,
+completely orthogonal question (of the tests that got compiled, run only
+ones matching this name) — and that restriction does not evaporate just
+because the first axis was maximally broadened. `go test -run TestFoo ./...`
+compiles every package in the module and then, within each one, runs only
+tests whose name matches `TestFoo` — the overwhelming majority of real test
+functions across the module (e.g. all 4 in `bytesconv`, none named `TestFoo`)
+never execute. Treating this as `broad` would make it functionally identical
+to running nothing at all while still marking every edited file in the
+session "verified" — precisely the false-"tests ran" bug this whole feature
+exists to close, just reached via a scope flag instead of a bare name.
+
+The classifier already gets this right by construction, not by any special
+case: a scope-broadening flag either lands in a genuinely orthogonal
+skip-flag set (`--workspace`/`-w` in the global `WORKSPACE_SCOPE_FLAGS`,
+consumed as flag+value and contributing nothing) or is excluded from
+path-likeness entirely (`./...` via `GLOB_ALL_TOKENS`) — in both cases it
+simply produces **no scoping evidence of its own**, neither for nor against
+name-filtering. The *separate* name-filter evidence (`-run`'s presence, or
+cargo's bare positional) is tracked independently and is what actually
+decides `broad` vs. name-filtered here; a scope flag appearing in the same
+command never suppresses that independently-collected evidence. This is
+exactly the axis-independence the `--workspace`/`--all`/`./...` case
+requires, achieved for free by not having any code path that treats "a
+scope flag is present" as license to ignore other evidence in the same
+segment. **Do not "fix" this by making a scope-broadening flag force
+`broad` regardless of a name filter also present in the same command** — that
+is, verbatim, the false-silence bug from the top of this document, just
+triggered by a different flag.
+
+Regression tests for all seven commands above are pinned in
+`test-run-matcher.test.ts`.
 
 ## C. `lien verify-tests <subcommand>`
 
--- a/docs/architecture/test-verification-nudge.md
+++ b/docs/architecture/test-verification-nudge.md
@@ -204,7 +204,7 @@ naming conventions (Python's `test_foo.py`, Go's `foo_test.go`, Ruby's
 `foo_spec.rb`) fall back to exact-basename matching, which already covers
 the common case of a run naming the real associated test file.
 
-#### Name-filtered runs are neither `broad` nor `scoped` (2026-07-29, #946)
+#### Name-filtered runs are neither `broad` nor `scoped` (2026-07-29)
 
 A run scoped by test **name** rather than file/directory (`pytest -k expr`,
 `dotnet test --filter expr`, `go test -run regex`, a bare `cargo test name`,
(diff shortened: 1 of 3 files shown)


alfhen:
## Follow-up: fixed the `cargo test` build-config-flag regression from review

Confirmed the review finding before touching anything:

```
cargo test --features foo            => broad:false  (WRONG -- was silently name-filtered)
cargo test -p my_crate                => broad:false  (WRONG)
cargo test --manifest-path ./Cargo.toml => broad:false, scopeTokens:["./Cargo.toml"] (WRONG -- bogus scope token)
cargo test --target x86_64-unknown-linux-gnu => broad:false (WRONG)
cargo test --profile release          => broad:false  (WRONG)
cargo test -j 4                       => broad:false  (WRONG)
```

Root cause: `cargo test`'s bare-positional detection (added by the previous
commit to catch `cargo test some_name`) couldn't tell a real positional test
name apart from the VALUE of a build/compile-config flag sitting in the same
textual position. Fixed by adding `CARGO_TEST_VALUE_FLAGS` /
`valueSkipFlagsFor` in `test-run-matcher.ts` -- these flags (`--features`/
`-F`, `-p`/`--package`, `--exclude`, `--manifest-path`, `--lockfile-path`,
`--target`, `--target-dir`, `--profile`, `-j`/`--jobs`, `--color`,
`--message-format`) are now skipped as flag+value, scoped to `cargo test`
only (not global -- some of these, like `-j`, aren't safe to blanket-skip
for every runner).

`--test <name>` (cargo's specific-integration-test-binary filter) is
deliberately left OUT of the skip set: it genuinely narrows which tests run,
unlike the flags above, so its value still flows through as a bare token and
lands on the same safe name-filtered (not broad) outcome as a plain
positional filter -- there's no infrastructure here to resolve a bare target
name against a real file path, so "keep nagging" is the correct fallback,
not "silently broad."

All confirmed fixed after the change (direct `classifyTestCommand` calls):

```
cargo test --features foo             => broad:true   (fixed)
cargo test -F foo                     => broad:true
cargo test --features=foo             => broad:true
cargo test -p my_crate                => broad:true
cargo test --package my_crate         => broad:true
cargo test --exclude other_crate      => broad:true
cargo test --manifest-path ./Cargo.toml => broad:true
cargo test --lockfile-path ./Cargo.lock => broad:true
cargo test --target x86_64-unknown-linux-gnu => broad:true
cargo test --target-dir ./target      => broad:true
cargo test --profile release          => broad:true
cargo test -j 4                       => broad:true
cargo test --jobs 4                   => broad:true
cargo test --color always             => broad:true
cargo test --message-format json      => broad:true
cargo test --test integration_test    => broad:false  (intentional -- see above)
cargo test some_name                  => broad:false  (unchanged -- original fix)
cargo test --features foo some_name   => broad:false  (compound: flag-value skip doesn't eat the real positional)
cargo nextest run                     => broad:true   (unaffected -- no regression)
tox -e py311                          => broad:true   (unaffected -- no regression)
```

Also re-dogfooded end-to-end in a **different** foreign repo this time
(`/tmp/lien-dogfood-460173c9/tokio`, a real Rust codebase, re-indexed with
`index --force`), through the real `test-run-note.sh` hook with true stdin:

- Edit `tokio/src/sync/mutex.rs` + `cargo test --features rt-multi-thread` ->
  `report` silent (correctly suppresses -- the regression is fixed).
- Same edit + `cargo test totally_unrelated_test_name` (different session) ->
  `report` nags (the original name-filter fix still works, unregressed by
  this follow-up).

tokio repo left clean (`git status --porcelain` empty).

Added 17 new unit tests (151 -> 168) pinning every flag above plus the
compound case and the `--test` design decision. All six gates green again
(`format:check`, `lint`, `typecheck`, `build`, `test` -- 1349 in `packages/cli`
vs 1332 before, `lien delta` exit 0 -- one new trivial helper, no threshold
crossings). `classifyTestCommand`'s own cognitive complexity (already at
12/15 per Lien's own hook warning before this change) was kept flat by
routing the new logic through an extracted `valueSkipFlagsFor` helper and
threading one more argument, rather than adding branches to that function
directly.

New SHA: `ceb3fbdcdc6821d10e5a8dc948af5c2d32fe9acb`

alfhen:
## Follow-up: fixed the pnpm `--filter` regression, with a direction correction

Confirmed the review's core claim before touching anything, computed directly
via `classifyTestCommand`:

```
pnpm test --filter hono           => broad:true                       (fine, plain name)
pnpm test --filter @hono/core     => broad:false, scopeTokens:["@hono/core"]  (BUG)
pnpm --filter hono test           => broad:true                       (fine -- dedicated anchored pattern)
pnpm --filter @hono/core test     => broad:true                       (fine -- dedicated anchored pattern)
turbo test --filter=web           => isTestRun:false                  (NOT a --filter bug -- see below)
```

**Two corrections to the review as relayed to me:**

1. The review's own example (`pnpm test --filter <pkg>`) does reproduce, but
   only when the value looks path-like (contains `/`, i.e. a scoped package
   name like `@hono/core`) -- a plain name (`hono`) already classified
   correctly as `broad:true` before this fix. Root cause: `--filter` was
   removed from the global `WORKSPACE_SCOPE_FLAGS` set in the very first
   commit of this PR specifically to stop it colliding with dotnet/swift's
   *different* meaning of the same flag spelling (a test-NAME filter) --
   correct move, but it left the pnpm-specific "filter after the script"
   ordering with no skip at all, so a slash-containing value fell through to
   the generic path-token scan.

2. `turbo test --filter=web` is **not** a `--filter`-handling bug at all --
   `turbo` isn't in `RUNNER_PATTERNS`, so the whole command is unclassified
   and falls through to the safe fail-open default (nags because no run was
   ever recorded, same as any unrecognized command). I did not add `turbo` as
   a new recognized runner -- that's new scope nobody has asked for and isn't
   a regression from anything in this PR.

**Fix**: added `PNPM_TEST_VALUE_FLAGS` (`--filter`, `-F`) via
`valueSkipFlagsFor`, gated on the generic `pnpm test`/`pnpm test:script`
match specifically (`isPnpmTestRunner`) -- not global, so dotnet/swift's
unrelated test-NAME `--filter` is untouched.

**A second, independent gap** surfaced while investigating and verified
against pnpm's own docs (fetched via context7,
[pnpm.io/filtering](https://pnpm.io/filtering)): `pnpm --filter=<selector>
test` (the single-token `=` form, filter BEFORE the script) matched **no**
runner pattern at all before this fix -- pnpm's docs show this exact form as
the canonical way to write an exclusion selector (`--filter=!foo`), so a
real, documented, non-exotic invocation always nagged. Fixed by widening the
dedicated anchored pnpm pattern to accept both the space and `=` forms, plus
pnpm's documented `-F` alias for `--filter`.

**Swept every other per-runner name-filter flag's `=` form** as asked
(`pytest -k=`/`-m=`, `rspec -e=`/`--example=`, `mocha --grep=`/`-g=`, `go
test -run=`, `vitest`/`jest -t=`/`--testNamePattern=`, `dotnet`/`swift
--filter=`) -- all already correctly name-filtered via the existing
`isNameFilterFlag`'s `token.startsWith(`${flag}=`)` check. No gap found
there; verified directly, not assumed.

**On the "flag arity table" question**: considered it, didn't do it. This is
the fourth instance of the same hazard class, but each instance so far has
been a small, independently testable addition to the existing per-concern
lookup functions (`nameFilterFlagsFor`, `valueSkipFlagsFor`) rather than a
sign the pattern itself is straining. A unified table would be a larger,
riskier diff for the same correctness outcome. Said so explicitly in the
architecture doc in case the next reviewer disagrees.

All 22 combinations verified directly against `classifyTestCommand` (both
orderings x both forms x plain/scoped name x the `-F` alias x the
`--filter=!excluded` exclusion form), 13 added as permanent regression
tests (181 total, was 168). Re-dogfooded in `hono` (the same repo used to
find this) through the real `test-run-note.sh` hook, re-indexed with `index
--force`:

- Edit `src/utils/url.ts` + `pnpm test --filter @hono/core` -> `report`
  silent (regression fixed).
- Same edit + `pnpm --filter=@hono/core test` (separate session) -> `report`
  silent (second gap fixed).
- Same edit + `vitest -t some_totally_unrelated_name` (separate session) ->
  `report` still nags (real name-filter behavior unregressed).

hono repo left clean. All six gates green (`format:check`, `lint`,
`typecheck`, `build`, `test` -- 1362 in `packages/cli`, was 1349 --
`lien delta` exit 0, `classifyTestCommand` not even listed in the delta
output this time -- its complexity stayed untouched).

New SHA: `28d37921567d85f25e976350141c4c723331df38`

alfhen:
## Final round: doc-truth fix + durable rebuttal, no code changes needed

Independently verified all five triaged findings before touching anything.

**Stale (2)** — confirmed already fixed by earlier commits in this PR
(`ceb3fbdc` for cargo `--features`, `28d37921` for pnpm `--filter`). Nothing
to do.

**Wrong (2)** — verified directly against `classifyTestCommand`, matching
the independent gin-corpus measurement:

```
go test -run TestFoo ./...      => name-filtered (NAGS)
go test -run TestFoo .          => name-filtered (NAGS)
cargo test foo --workspace      => name-filtered (NAGS)
cargo test foo --all            => name-filtered (NAGS)
go test ./...                   => broad (SILENT)
cargo test --workspace          => broad (SILENT)
cargo test --all                => broad (SILENT)
```

No code change needed — the existing implementation already gets this
right, by construction: `./...`/`--workspace`/`--all` either land in the
`WORKSPACE_SCOPE_FLAGS` skip set or are excluded from path-likeness
(`GLOB_ALL_TOKENS`), contributing no scoping evidence of their own; the
name-filter evidence (`-run`'s presence, or cargo's bare positional) is
tracked independently and isn't suppressed by a scope flag appearing
alongside it. Adopting the review's suggestion ("keep broad regardless of
positional name") would reintroduce the exact false-silence bug this PR
exists to fix.

Added a durable "selection vs. scope are independent axes" section to
`docs/architecture/test-verification-nudge.md` explaining this, plus 7
pinning regression tests (188 total, was 181) covering all 4 name-filtered
cases and their 3 broad contrasts, so a future review has both the reasoning
and the test suite as evidence before re-raising this.

**Real (1)** — fixed the doc-truth error: `docs/architecture/test-verification-nudge.md`
said "Two token classes are excluded" (introduced in this PR's first commit)
while the list right below it enumerates three (workspace-scope flag
values, config-flag values, Go's glob-all convention). Restored "Three."

Re-dogfooded in `gin` (the repo used to verify this round) through the real
`test-run-note.sh` hook, re-indexed with `index --force`: edited
`internal/bytesconv/bytesconv.go` (4 real test functions, none named
`TestFoo`), then `go test -run TestFoo ./...` -> `report` nags (correct);
same edit + `go test ./...` (separate session) -> `report` silent (correct
contrast). Repo left clean.

All six gates green — this round changed only docs + tests, so `lien delta`
reports zero complexity changes across the repo.

New SHA: `83dd774965cdd6bb18108e050ddb03e024456ee6`

PR #992, opened 2026-07-31
refactor(parser,cli): move findDependents into parser, decoupled from VectorDBInterface

alfhen:
## Why

`findDependents` (formerly `packages/cli/src/mcp/handlers/dependency-analyzer.ts:741`) is the hardened, actively-maintained dependency analysis behind the `get_dependents` MCP tool — but it sat in `cli`, the top of the package stack (`parser` ← `core` ← `cli`), so nothing below could reuse it. `packages/review` (which depends on `parser` only, not `core`) grew its own weaker dependency graph as a result. This PR moves `findDependents` into `@liendev/parser`, generic and chunk-based, so it's finally shareable.

This is the mirror image of #987 (`getLienHome`/`getIndexDir`/`extractRepoId` → `core`): that PR moved Lien-product/infra concerns *out* of `parser` to keep it a pure source-analysis library; this one moves source-code analysis *into* `parser` for the same reason, from the other direction.

## The database dependency was an illusion (re-verified on this branch)

| checked | result |
|---|---|
| `@liendev/core` imports in the 1,494-line file | `import type` only — no runtime dependency |
| `VectorDBInterface` methods used | exactly one: `vectorDB.scanAll()` |
| `indexVersion` | a cache key only |
| `SearchResult`-only fields (`score`, `relevance`) used | none — only `CodeChunk` fields |
| `scanPaginated` consumed anywhere in the repo | never |

`scanAll()` was an injection point, not a data dependency. Inverting it (parser takes chunks, cli fetches them) removes the coupling.

## Merge vs. sibling module — I merged

`packages/parser/src/dependency-analyzer.ts` already held `analyzeDependencies` (used by `get_complexity`/`ComplexityAnalyzer`). I merged `findDependents` into the **same file** rather than a new sibling module:

- Doing so let the actually-duplicated helpers become **single, generic, private implementations** shared by both algorithms (see below) instead of needing to `export` them from one file just so a sibling module could import them — that would grow parser's public API surface for pure implementation details.
- `analyzeDependencies`'s own body is **untouched** (only its private helper *signatures* gained `<T extends CodeChunk>`, inferred transparently at its existing call sites) — confirmed by `lien delta`: `analyzeDependencies` doesn't even appear in the delta output, i.e. zero complexity change. Its cognitive-14/15 headroom is undisturbed.
- The two algorithms stay clearly separated by a section comment and distinct top-level exports (`analyzeDependencies` vs `findDependents`), so this isn't a single 2,000-line blob — it's one file with two related families.

## The six duplicated pairs — per-function drift decisions

Brief note: mid-task, PR #973 merged upstream and #987 also landed — I rebased onto both (`git merge --ff-only origin/main` twice) before writing any code. #973 had already generic-ified `addFuzzyMatchChunks`/`findDependentChunks` and deleted the cli copies, so those two were **not** among the six I had to reconcile (confirmed: only 6 duplicate-named functions remained, not 8, once rebased).

| function | decision | why |
|---|---|---|
| `calculateComplexityRiskBoost` | **merged, unchanged logic** | byte-identical between the two files; only the return-type name differed (`RiskLevel` alias vs local literal union) — structurally identical, safe to unify. |
| `createPathNormalizer` | **kept parser's parameterized version** (`(workspaceRoot: string) =>`) | cli's version silently read `process.cwd()` internally; parser's takes it explicitly. `analyzeDependencies`'s own tests inject a fixed `/test/workspace` — a `process.cwd()`-reading version would break that. `findDependents` now also takes `workspaceRoot` as a required, explicit parameter for the same testability reason (a real, disclosed API design choice on the new function, not forced by the merge — see below). |
| `groupChunksByFile` | **kept parser's parameterized version** | same reasoning as above; cli's recomputed `process.cwd()` per call instead of accepting it as an argument. |
| `buildReExportGraph` | **kept parser's simpler `string[]`-returning version** | cli's returned `ReExporter[]` (`{filepath, reExportedSymbols}`), but I verified (grepped every call site) `reExportedSymbols` was populated and then **never read again** — every downstream consumer only used `.filepath`. Parser's version already discarded that dead data. Eliminates the now-unused `ReExporter` interface. |
| `calculateFileComplexities` | **merged, unchanged logic** | logic was identical; only the intermediate-variable style differed and the type name differed (`FileComplexity` vs `FileComplexityInfo` — structurally identical). Kept parser's exported name (`FileComplexity` had zero external consumers, confirmed by grep). |
| `calculateOverallComplexityMetrics` | **kept both, deliberately** | REAL behavioral drift: parser's returns `undefined` when there's no complexity data (pinned by `analyzeDependencies`'s own test: `expect(result.complexityMetrics).toBeUndefined()`); cli's always returned a concrete zeroed object (`get_dependents`'s consumers dereference `.maxComplexity` etc. unconditionally). Both contracts are load-bearing for different callers, so I did not unify them. Added `calculateComplexityMetricsOrDefault`, a **thin wrapper** (`calculateOverallComplexityMetrics(...) ?? <zeroed default>`) for `findDependents`'s always-present need, instead of copying the ~25-line computation a third time. |

## Design choices beyond direct relocation (disclosed)

1. **`findDependents` takes `Iterable<T>` chunks + explicit `workspaceRoot`, not a prebuilt scan index.** It builds its own import index internally every call — the same self-contained style as its `analyzeDependencies` sibling (which also takes raw chunks and rebuilds its index every call, no caching). `Iterable<T>` (not `T[]`) keeps `scanPaginated` pluggable later without another signature change, per the brief's prompt.
2. **`scanCache` in the cli wrapper now caches the raw chunk array**, not a prebuilt import-index Map (today's design cached the built `{importIndex, allChunksByFile}` maps). This is a real, disclosed shift: a cache **hit** now still pays the O(n) in-memory index-build cost (cheap; the expensive part, per the removed `scanAllChunks` comment, was always the DB read — `vectorDB.scanAll()` — which the cache still fully avoids on hit). I did not fabricate a benchmark for this repo's index size; the E2E suite reindexed the CLI's own repo plus MediatR (1449 chunks), Requests (841 chunks), and Zod (5078 chunks) without any timeout or slowdown. No test asserts on cross-call index-build caching (only on `vectorDB.scanAll()` call counts), so this doesn't break `dependency-analyzer.test.ts`'s "import index caching" suite. Flagging this explicitly since it's the one place I deviated from "byte-for-byte same architecture," even though the observable API and results are unchanged.

## Scope discipline — what I did NOT do

- Did not restructure the import index to preserve raw specifiers (separate planned phase).
- Did not wire `packages/review` up to use the shared `findDependents` (separate phase — this PR only makes it reachable).
- Noticed but left alone: `get-dependents.ts`'s doc comments referencing `dependency-analyzer.ts` needed light touch-ups (3 spots) now that the referenced functions live in `@liendev/parser`, not the sibling cli file — fixed as pure prose, no behavior touched. Also updated 2 stale "the CLI's dependency-analyzer.ts" references in `csharp-type-reference-signals.ts` now that there's no separate CLI copy.

## Dogfood — before/after `get_dependents`, byte-identical

Measured against **the same commit** (`4b5efb62`, before my change) for both sides, toggling only the 5 changed files via `git checkout <sha> -- <paths>` — not two different points in git history — since this is a fleet session with many concurrent agents landing PRs, and an earlier attempt (comparing across two different rebase points) produced misleading diffs caused entirely by *other* merged PRs changing the dependency graph, not by this change. Re-ran the tight comparison twice for stability; both identical, including the full sorted `dependents` filepath list per query (not just counts):

```
packages/parser/src/utils/path-matching.ts
  dependentCount=34 prod=26 test=8 risk=high caveat=null        [before == after]

packages/parser/src/dependency-analyzer.ts
  dependentCount=29 prod=22 test=7 risk=high caveat=null        [before == after]

packages/cli/src/mcp/handlers/dependency-analyzer.ts
  dependentCount=6 prod=3 test=3 risk=low caveat=null           [before == after]

packages/parser/src/dependency-analyzer.ts (symbol=findDependentChunks)
  dependentCount=1 prod=1 test=0 risk=low caveat=null           [before == after]

packages/cli/src/mcp/handlers/get-dependents.ts
  dependentCount=3 prod=2 test=1 risk=medium caveat=null        [before == after]

packages/parser/src/index.ts
  dependentCount=26 prod=20 test=6 risk=medium caveat=null      [before == after]

packages/cli/src/mcp/handlers/dependency-analyzer.ts (symbol=findDependents)
  dependentCount=6 prod=3 test=3 risk=low caveat=null           [before == after]

packages/parser/src/types.ts
  dependentCount=48 prod=36 test=12 risk=high caveat=null       [before == after]
```

Method: spawned the real `node packages/cli/dist/index.js serve` over stdio, connected a real MCP `Client`, called the real `get_dependents` tool for both file-level and symbol-level queries.

## Gates

| Gate | Result |
|---|---|
| `format:check` | pass |
| `lint` | **0 errors** (37 pre-existing warnings, none in touched files) |
| `typecheck` | pass |
| `build` | pass |
| `npm test` | **5290 tests, 0 failures** |
| `lien delta` | **exit 0** — 2 improved (`getOrScanChunks` time, `findDependents` cognitive 3→0 on the cli side, since it's now a thin wrapper), 3 worsened-but-not-crossed (minor "time to understand" cost from generic signatures on `groupChunksByFile`/`calculateFileComplexities`/`buildReExportGraph`), 34 relocated. `analyzeDependencies` does not appear in the output at all — zero complexity change. |

## E2E (language-most-affected matrix)

| Suite | Result |
|---|---|
| `test:e2e:csharp` (MediatR, 160 files / 1449 chunks) | pass |
| `test:e2e:python` (Requests, 51 files / 841 chunks) | pass |
| `test:e2e:typescript` (Zod, 454 files / 5078 chunks) | pass |

## What moved where

- `packages/parser/src/dependency-analyzer.ts`: gains `findDependents`, `FindDependentsResult<T>`, `DependentInfo`, `SymbolUsage`, `ComplexityMetrics` (all generic/chunk-based), plus every private helper the old cli file had (now generic over `<T extends CodeChunk>`), and `calculateComplexityMetricsOrDefault`.
- `packages/parser/src/index.ts`: now also exports `findDependents`, `FindDependentsResult`, `ComplexityMetrics`, `DependentInfo`, `SymbolUsage`.
- `packages/cli/src/mcp/handlers/dependency-analyzer.ts`: shrinks from 1494 lines to a ~95-line wrapper — `scanCache`/`clearDependencyCache`/`getOrScanChunks` (vectorDB-backed caching) plus `findDependents` (fetches chunks, delegates to parser). `DependencyAnalysisResult`/`ComplexityMetrics`/`DependentInfo` are now type aliases/re-exports of parser's generic types, so `get-dependents.ts`, `annotate-cmd.ts`, and `api-delta-cmd.ts` needed zero changes.

Changeset included (`@liendev/parser` minor — new public API; `@liendev/lien` patch).

---

### Lien Review

> [!NOTE]
> **Low Risk**
>
> This PR is a clean structural refactor: it moves `findDependents` and its helpers from `packages/cli` into `packages/parser` as generic, chunk-based functions, and reduces the CLI file to a thin `vectorDB.scanAll()` wrapper. The algorithm logic is preserved, the public API surface is intentionally expanded in `@liendev/parser`, and the CLI re-exports the types it needs. No breaking changes were found: the only removed export (`FileComplexity`) has zero surviving references, and the disclosed cache architecture change (raw chunks instead of a prebuilt import index) does not alter observable tool output.
>
> - Moved `findDependents` engine into `@liendev/parser` as `Iterable<T>`-in/chunk-out, decoupled from `VectorDBInterface`
> - CLI `dependency-analyzer.ts` is now a ~95-line wrapper that fetches/caches chunks and delegates to parser
> - Shared helpers between `analyzeDependencies` and `findDependents` are now single generic implementations

<sup>Reviewed by [Lien Review](https://lien.dev) for commit c7576b8. Updates automatically on new commits.</sup>


alfhen:
Quantifying the disclosed trade-off, since the PR body flags it without a number.

**The change:** `scanCache` went from holding the prebuilt indexes to holding raw chunks:

```
before:  { importIndex, allChunksByFile, ... }   → a cache hit skipped the index build
after:   { indexVersion, chunks }                → a cache hit rebuilds it each call
```

`lien serve` is long-lived, so this is a real per-call cost on the 2nd..Nth `get_dependents` in a session, which is the hot path (CLAUDE.md makes the tool mandatory before symbol changes).

**Measured on this repo** — replicating exactly the work a cache hit used to skip (path normalization with the same memoizing wrapper, plus building `importIndex` and the per-file map over every chunk), 5 consecutive runs:

```
chunks: 9,864
index-build cost per call: 7, 7, 6, 5, 2 ms   → median 6 ms
```

**Verdict: acceptable, and not worth re-complicating the cache for.** 6 ms against `findDependents`' own 83–161 ms (measured separately on this repo) and an MCP round-trip is noise. Even a repo 10× this size lands around 60 ms, still well inside the budget.

And the simplification buys something real: `@liendev/parser` keeps **no mutable module state**, and index construction lives with the analysis that owns it, matching `analyzeDependencies`'s existing style. Caching a prebuilt index in the CLI would mean the CLI knowing parser's internal representation — precisely the coupling this move exists to remove.

If a much larger repo ever makes this measurable, the fix is local and obvious: memoize the built index inside parser keyed on chunk-set identity, without the CLI needing to know the shape. Not needed now.

Good call disclosing it rather than burying it — that's what made it cheap to check.

Issue #994, opened 2026-07-31
Duplication refactor: remaining Phases 3, 4, 5 (roadmap — the plan doc is gitignored)


alfhen:
Tracking issue for the unfinished half of the 2026-07-31 duplication refactor.

**Why this issue exists:** the full plan lives at `.wip/duplication-refactor-plan.md`, and `.wip/` is gitignored (`.gitignore:96`, zero tracked files). So the roadmap currently exists on one machine only. Everything needed to act is restated below.

Completed and shipped in **0.74.0**: Phase 0 (#985, #981), Phase 0.5 (#982 #983 #984 #986 #989 #990 #991), Phase 1 (#984, #993), Phase 2a (#987), Phase 2b (#992, #973).

---

## Phase 3 — the import index throws away what the guards need  ⭐ recommended next

**Strongest evidence of real cost of anything remaining.** `packages/parser/src/utils/path-matching.ts:378-388` documents four match-side call paths deliberately NOT routed through the `importMatchesTarget` primitive that #886 introduced:

- `buildImportIndex` (parser) and `indexImportEntry`/`addChunkToImportIndex` (cli) — call `isUnresolvableWholeModuleImport` directly
- `findDependentChunks`'s fuzzy loop, and `buildReExportGraph` — stay on raw `matchesFile`

**Why they can't be routed:** the index is `Map<normalizedSpecifier, Chunk[]>` and building it discards (a) the raw specifier and (b) per-importer identity — one bucket can hold chunks from many importer files in different languages. `importMatchesTarget(rawSpecifier, importerFile, target, normalize)` needs both. So the same three guards (#884 whole-module, #887 single-file-vs-package, #929/#953 Python bare-package) get expressed in two shapes: once per-importer up front, once split across build-time and per-chunk match-time.

From the source: *"`findDependentChunks`'s own doc comment for how it independently derives the #887 distinction per-chunk (it has no single importer file)."*

**Proposed fix:** store `{ chunk, rawSpecifier }` entries instead of bare chunks. Keep the normalized key and the build-time `isUnresolvableWholeModuleImport` prune (it's an early-drop optimization; retaining the raw specifier doesn't require giving it up). Then every entry carries `rawSpecifier` **and** `chunk.metadata.file`, so the fuzzy loop can call `importMatchesTarget` directly and all four sites collapse onto the primitive. Cost: one string reference per index entry.

**Dated evidence this keeps biting:**
[...]

Issue #953, opened 2026-07-29
fix(parser): bare-directory relative imports (../.., ../utils) fabricate get_dependents edges


alfhen:
## Bug

\`get_dependents\` reports fabricated \`hops:1\` dependents and inflates \`riskLevel\` for any file whose corpus contains a relative import that resolves to a bare directory path instead of a real file.

Confirmed on the published 0.72.0 CLI against \`getlien/hono\` (via a foreign-repo dogfood pass): \`src/middleware/jwt/index.ts\` contains only:

\`\`\`ts
import type { JwtVariables } from './jwt'
export { jwt, verifyWithJwks, verify, decode, sign } from './jwt'
export { AlgorithmTypes } from '../../utils/jwt/jwa'
import type {} from '../..'          // dots-only, empty type-only import
\`\`\`

\`get_dependents({filepath: "src/utils/color.ts"})\` and \`get_dependents({filepath: "src/utils/url.ts"})\` both report \`jwt/index.ts\`, \`ip-restriction/index.ts\`, and \`bun/conninfo.ts\` as direct (\`hops:1\`) dependents, with no caveat. Neither file has any relationship to jwt/ip-restriction/conninfo — the grep count for "url"/"color" in those files is 0.

## Root cause

\`resolveRelativeImport\` (packages/parser/src/utils/path-matching.ts) correctly joins \`'../..'\` against \`src/middleware/jwt\` to produce the bare path \`src\` — but never resolves that bare directory name to its real entry point (\`src/index.ts\`). Left as bare \`src\`, the specifier then fuzzy-matches through \`matchesFile\`'s Strategy 5 (Python bare-module \`matchesParentPythonPackage\`, since \`src\` satisfies the bare-word regex) against EVERY file nested under \`src/\`, for a TypeScript importer that has no Python semantics at all. This is the exact repro already described (but left unguarded for this call path) in \`matchesFile\`'s own #929 doc comment: \`findDependentChunks\`'s fuzzy loop (in both \`packages/parser/src/dependency-analyzer.ts\` and \`packages/cli/src/mcp/handlers/dependency-analyzer.ts\`) calls raw \`matchesFile\` rather than the guarded \`importMatchesTarget\`, so the #929 \`allowPythonModuleMatching\` guard never applies there.

## Fix

[...]

PR #1011, opened 2026-07-31
refactor(review): route dependency-graph edge resolution through parser's guarded primitives

alfhen:
## Summary

Issue #994, Phase 5. `packages/review/src/dependency-graph.ts` had its own `resolveImportPath` — relative-only, JS/TS-only, a hardcoded 6-extension list — sharing zero path-matching with `@liendev/parser`'s guarded, 12-language `importMatchesTarget`/`matchesFile`. Every guard fix parser gained (#884, #886, #887, #929, #934, #953, #955) never reached review's `blast_radius` signal.

- **Deleted `resolveImportPath`.** Replaced with `importMatchesTarget`-based resolution, verified against the symbol's known export locations (`exportIndex`) instead of a corpus-wide file scan — same O(n)-once/O(1)-query architecture, now multi-language.
- **Kept review's own BFS untouched.** `bfsTransitiveCallers` (via parser's `walkBounded`) is unchanged — hop semantics stay call-graph hops throughout. Only per-edge resolution changed.
- **C# gating bug fixed.** `addSameNamespaceEdges`'s same-directory heuristic ran for `csharp` (`!['typescript','javascript'].includes(lang)`) despite its own comment naming only PHP/Python/Rust. C# now routes through `findCSharpTypeReferenceDependents` (#930/#971) instead.
- **New import-only fallback** (`buildImportOnlyEdges`) for a real gap a mid-task measurement surfaced (see "Course correction" below): a class/type-shaped seed referenced only via constructor injection — never a literal call to the class's own name — resolved to **zero** callers under the old call-site-only model, even with real, verified dependents.
- **Provenance tag added** (`EdgeProvenance`: `same-file` / `import-verified` / `import-only` / `symbol-name-match` / `oop-method-import` / `namespace-inferred`) on every `CallerEdge`/`BlastRadiusDependent`, surfaced as a new **Confidence** column in `<blast_radius>` and in `get_dependents`'s JSON output — so the agent can weight a verified edge above a name-matched guess.

## Course correction mid-task

The original brief recommended calling parser's `findDependents` directly per BFS node. I measured this instead of implementing it as specified:

```
single findDependents call (symbol, depth1): 39.16 ms
240x findDependents calls total: 6131.74 ms, avg: 25.55 ms/call
```
(9,864-chunk corpus — this repo at the PR #981 capture point.) `findDependents` rebuilds its whole import index every call with no way to memoize across different (file, symbol) queries, so this cost scales with corpus size and would also land on the **live** `get_dependents` agent tool's per-call latency (called synchronously, mid-investigation, possibly many times per review). I pushed back on the literal recommendation and instead routed only the **per-edge guards** (`importMatchesTarget` et al., all already exported from parser) through review's existing one-time index build, preserving O(n)-once/O(1)-query performance while still sharing every language guard. This also sidesteps the "both call it `hops` but count different graphs" ambiguity entirely — review's BFS never touches `findDependents`'s own hop/depth machinery.

Separately, the coordinator ran an independent measurement mid-task and found `getCallers` resolves via `exports`-listed symbols at 0% for PHP (0/152) vs 45-98% via a chunk's own `symbolName` — root-caused to PHP classes being called via their *methods*, never their own class name as a literal call site. I reproduced this independently (see below) before building on it, and it changes the fix's shape: `provenance` alone isn't enough if a real dependent still returns zero. That's what `buildImportOnlyEdges` is for — recovering the case that call-site-only matching structurally cannot see.

## Reproduced the coordinator's finding independently

Captured a real PR fixture (#1003, this branch's own testbed edits — see below) and ran `findDependents` directly against parser, no review code involved:

```js
findDependents(chunks, 'lien-review-testbed/rust/src/config.rs', ..., 'Config', 1) → 0 dependents
```
Confirmed the analogous PHP case pre-Phase-5 (`buildDependencyGraph` on `main`): `getCallers('PricingService.php', 'PricingService')` → **0** callers, `risk: low`, despite 4 real files depending on it. Matches the coordinator's report; not repeated here as a re-measurement, just independently confirmed before building the fix on top of it.

## Measurement (before/after, zero LLM cost)

Deterministic `computeBlastRadius` checkers — no `OPENROUTER_API_KEY`, never swept into `--calibrate` (named `.blast-radius.ts`, not `.assertions.ts`):

| Fixture | Seed | Before (main@099c1d78) | After |
|---|---|---|---|
| PR #981 (Python, pre-existing) | `check_required_fields` | 3 direct / 12 total, risk=high | **unchanged** — 3 direct / 12 total, risk=high |
| PR #1003 (PHP, new) | `PricingService` (class) | **0 dependents**, risk=low | **2 direct dependents** (both `import-only`), risk=medium |
| PR #1003 (Rust, new) | `Config` (struct) | 0 dependents, risk=low | **unchanged** — 0 dependents, risk=low (see below) |

```
$ npx tsx packages/review/test/harness/fixtures/blast-radius/pr981-python-check-required-fields.blast-radius.ts
OK — check_required_fields@lien-review-testbed/python/pipeline/validator.py: 3 direct dependents match baseline, 12 total, risk=high.

$ npx tsx packages/review/test/harness/fixtures/blast-radius/pr1003-php-pricingservice-guard.blast-radius.ts
OK — PricingService@lien-review-testbed/php/app/Services/PricingService.php: 2 direct dependents (all import-only, PRE-#994-Phase-5 this was 0), risk=medium.

$ npx tsx packages/review/test/harness/fixtures/blast-radius/pr1003-rust-config-guard.blast-radius.ts
OK — Config@lien-review-testbed/rust/src/config.rs: 0 dependents, risk=low (pinned known-gap baseline — see module doc for the parser-level path-depth cap this is blocked on).
```

**Rust is an honest non-improvement, not a bug in this PR.** `Config`'s real dependents (`cache.rs`, `analyzer.rs`, `main.rs`, ...) all reference it via a bare `use crate::config::Config;` → `importedSymbols: { config: ['Config'] }`. `lien-review-testbed/rust/` nests each language's sample crate 2 directories below the repo root, so matching `config` against `lien-review-testbed/rust/src/config` needs 3 leading path segments — over parser's `matchesAtBoundaryPrecise`'s deliberate `maxLeadingSegments: 1` cap (#868/#883, meant for a single `src/`-style prefix). Confirmed this is a **pre-existing, already-shared parser limitation**, not something this PR introduced or could fix by routing through parser more thoroughly:
```js
importMatchesTarget('config', 'lien-review-testbed/rust/src/analyzer.rs', normalize('lien-review-testbed/rust/src/config.rs'), normalize) // false
importMatchesTarget('config', 'src/analyzer.rs', normalize('src/config.rs'), normalize) // true — same shape, 1 level deep
findDependents(chunks, 'lien-review-testbed/rust/src/config.rs', ..., 'Config', 1) // 0 dependents — parser's OWN get_dependents engine, same corpus
```
Filed as a follow-up rather than patched around in review — loosening the shared cap locally would reintroduce exactly the kind of review-side path-matching divergence this phase exists to eliminate. Almost certainly a testbed-nesting artifact (a real crate's `src/` is usually 0-1 directories from its repo root), not a general Rust concern — but the fixture pins the measured truth, not the probable case, and documents this thoroughly (see its module doc).

## PHP testbed fixture PR (merged separately)

lien-review-testbed had no PHP/Rust PR history to `capture-pr.ts --sha` against (only direct "scenario/*" commits, never opened as PRs — confirmed via `gh api .../commits/<sha>/pulls` → `[]`). Opened and merged #1003 (`test(review-testbed): tighten PricingService and Config guards`) specifically to create one, mirroring #981's precedent. Lien Review itself caught a real bug in my first draft (returning a negative total unchanged violated `applyDiscount`'s own "never goes below zero" doc contract) — fixed before merge, re-captured against the corrected SHA (`77820895`).

## Gates

```
npm run format:check   ✓
npm run lint           ✓ 0 errors (37 pre-existing warnings, none in touched files)
npm run typecheck      ✓
npm run build          ✓
npm test                ✓ all test files pass (parser/core/cli/review/action — 226 files)
lien delta              ✓ exit 0 (no NEW-over-threshold or crossed function; see below)
```

`lien delta` shows several `worsened`/`new` entries but nothing crosses a threshold — the biggest new function (`collectImportOnlyFileKeys`, cognitive 14) was refactored down from an initial 17 specifically to stay clear of the 15 threshold. `resolveImportPath` (cognitive 14) is fully removed; `resolveChunkImports` improved 13→5.

Added 5 new unit tests to `dependency-graph.test.ts` (import-only recovery, no-fabrication guard, per-file dedup, C# same-directory exclusion, C# type-reference-fallback wiring) plus `provenance` assertions on 5 existing tests covering each tier. 28/28 pass.

## Harness evidence (byte-identical census, no `--calibrate`)

`build-prompts.ts` always passes `blastRadius: null` — prompt assembly deliberately does not include blast-radius output, so it cannot exercise this PR's change surface. Ran it anyway, before/after, on the 3 tracked (non-gitignored) fixtures to confirm **zero unintended effect on prompt assembly**:

```
sha256 before == after, all 3:
0c94b54734e4c37fdeea386d023948476e9e7273cc44631dcf176f63a40fee65  doc-truth/accurate-doc.fixture.json
5cf9d29cf625b0cd8b0303e0ac9d4a3d414a83c3368ceb588f9db39dd202d781  doc-truth/renamed-stale-claim.fixture.json
17c007762d9fea46bfa404206bcbfabebe8c2fd4139794361f66b19f688d2191  boundary-change/placeholder.fixture.json
```
byte-identical, 3/3 — via `npm run test:harness:build-prompts -w @liendev/review`. The real evidence for the actual change (blast-radius computation) is the deterministic `.blast-radius.ts` checkers above, which directly exercise `computeBlastRadius`/`buildDependencyGraph` with zero LLM spend.

## Design decisions / pushback

- **Did not literally call `findDependents` per BFS node** — measured 25ms/call cost, would scale with corpus size and hit the live tool. Routed the guards through instead (see "Course correction").
- **Did not extend the same-namespace fallback to be call-site-free.** PHP's `OrderService`/`CheckoutService` reference `PricingService` implicitly (same namespace, no `use` statement, no matching call site) and remain an unrecovered, documented gap — see the PHP fixture's module doc. A same-directory match with zero call-site OR import evidence at all would be a materially weaker signal than the import-only fallback (which requires a *verified* import statement); adding it risked reintroducing a false-hub class of bug this whole phase exists to remove. Flagged as a known limitation, not silently claimed as fixed.
- **`review`/`action` are private/unpublished** — no changeset (no `packages/parser`/`core`/`cli` changes in this PR).

Reference: issue #994, Phase 5.

---

### Lien Review

➡️ **Stable** - 7 pre-existing issues in touched files (none introduced).
🔗 **Impact**: 1 high-risk file(s) with 1 total dependents
| ⏱️ time to understand | 7 | +10 |

> [!NOTE]
> **Low Risk**
>
> This PR refactors review's dependency-graph edge resolution to route through @liendev/parser's guarded `importMatchesTarget` primitives, deletes the old JS/TS-only `resolveImportPath`, adds a provenance tag to every `CallerEdge`/`BlastRadiusDependent`, and introduces import-only and C# type-reference fallbacks. The changes are well-scoped and internally consistent: the new `provenance` field threads correctly through `blast-radius.ts`, `blast-radius-render.ts`, and the `get_dependents` agent tool; `buildDependencyGraph`'s new optional `workspaceRoot` parameter is supplied by both production call sites; and the fallback ordering (direct call-site edges → import-only edges → C# fallback → empty) is sound. The only surviving references to the removed `resolveImportPath` are historical comments in test files, not live code usages, so there is no breaking change.
>
> - Replaced review's hardcoded relative-import resolver with parser's multi-language `importMatchesTarget` guards.
> - Added `EdgeProvenance` tiers and surfaced them as a Confidence column in blast-radius output and `get_dependents` JSON.
> - Added import-only fallback to recover class/type seeds with verified imports but no literal call sites.
> - Fixed C# same-directory gating bug and wired in parser's `findCSharpTypeReferenceDependents` fallback.

<sup>Reviewed by [Lien Review](https://lien.dev) for commit 3457830. Updates automatically on new commits.</sup>

---

Attested: degraded:budget_starved · 674K/805K tokens

### DIFF
--- a/packages/review/src/blast-radius-render.ts
+++ b/packages/review/src/blast-radius-render.ts
@@ -6,6 +6,7 @@
  */
 
 import type { BlastRadiusReport, BlastRadiusEntry, BlastRadiusDependent } from './blast-radius.js';
+import { isPreciseProvenance } from './dependency-graph.js';
 
 const LEVEL_LABEL: Record<string, string> = {
   low: 'LOW',
@@ -26,7 +27,9 @@ const NON_SUBSTITUTION_NOTE =
   'Tests column = a test file exists for that dependent, NOT that it covers ' +
   "this PR's change; Complexity = a static metric, not a correctness check. " +
   'This table does not substitute for get_files_context / read_file on a ' +
-  'dependent when you need to confirm the change is actually handled there.';
+  'dependent when you need to confirm the change is actually handled there. ' +
+  'Confidence: verified = a real import/call-site edge; inferred = a ' +
+  'name-matched, same-namespace, or text-matched guess — weight it lower.';
 
 /**
  * Render a blast-radius report as markdown wrapped in a `<blast_radius>` XML tag.
@@ -67,8 +70,8 @@ function renderSummary(report: BlastRadiusReport): string {
 
 function renderTable(entries: BlastRadiusEntry[]): string[] {
   const rows: string[] = [];
-  rows.push('| Seed (changed) | Hops | Dependent | Tests | Complexity |');
-  rows.push('|---|---|---|---|---|');
+  rows.push('| Seed (changed) | Hops | Dependent | Tests | Complexity | Confidence |');
+  rows.push('|---|---|---|---|---|---|');
 
   for (const entry of entries) {
     const seedLabel = formatSeed(entry);
@@ -79,12 +82,23 @@ function renderTable(entries: BlastRadiusEntry[]): string[] {
     });
     for (const dep of deps) {
       rows.push(
-        `| \`${seedLabel}\` | ${dep.hops} | ${formatDependent(dep)} | ${dep.hasTestCoverage ? '✓' : '✗'} | ${formatComplexity(dep)} |`,
+        `| \`${seedLabel}\` | ${dep.hops} | ${formatDependent(dep)} | ${dep.hasTestCoverage ? '✓' : '✗'} | ${formatComplexity(dep)} | ${formatConfidence(dep)} |`,
       );
     }
   }
 
   return rows;
+}
+
+/**
+ * `provenance` is absent only for hand-built test fixtures that predate
+ * #994 Phase 5 (see `BlastRadiusDependent.provenance`'s doc comment) —
+ * rendered as "verified" rather than surfacing that gap to the agent, since
+ * every real `computeBlastRadius` output always sets it.
+ */
+function formatConfidence(dep: BlastRadiusDependent): string {
+  if (!dep.provenance || isPreciseProvenance(dep.provenance)) return 'verified';
+  return 'inferred';
 }
 
 function formatSeed(entry: BlastRadiusEntry): string {
--- a/packages/review/src/blast-radius.ts
+++ b/packages/review/src/blast-radius.ts
@@ -8,7 +8,7 @@
 
 import type { CodeChunk, RiskLevel, BlastRadiusRisk } from '@liendev/parser';
 import { findTestAssociationsFromChunks, computeBlastRadiusRisk } from '@liendev/parser';
-import type { DependencyGraph } from './dependency-graph.js';
+import type { DependencyGraph, EdgeProvenance } from './dependency-graph.js';
 
 // ---------------------------------------------------------------------------
 // Types
@@ -34,6 +34,17 @@ export interface BlastRadiusDependent {
   complexity?: number;
   /** True when a test file imports the dependent's source file. */
   hasTestCoverage: boolean;
+  /**
+   * How this dependent edge was resolved — carried through from
+   * `dependency-graph.ts`'s `CallerEdge.provenance` (#994 Phase 5) so the
+   * reviewing agent can weight a verified import/call-site edge above a
+   * name-matched guess or an inferred, no-import structural signal, instead
+   * of treating every row in `<blast_radius>` as equally solid. See
+   * `EdgeProvenance`'s doc comment for what each tier means. Optional only
+   * so hand-built test fixtures don't all need one; every dependent produced
+   * by `computeBlastRadius` itself always sets it.
+   */
+  provenance?: EdgeProvenance;
 }
 
 export interface BlastRadiusEntry {
@@ -184,6 +195,7 @@ function buildDependent(
     callSiteLine: edge.callSiteLine,
     complexity: complexity && complexity > 0 ? complexity : undefined,
     hasTestCoverage: coveredFiles.has(edge.caller.filepath),
+    provenance: edge.provenance,
   };
 }
 
--- a/packages/review/src/dependency-graph.ts
+++ b/packages/review/src/dependency-graph.ts
@@ -4,15 +4,73 @@
  * Resolves caller/callee relationships using imports, exports, and callSites
  * without any vector DB. Used by the bug-finding plugin to find all callers
  * of changed functions across the full repo.
+ *
+ * Phase 5 of the duplication refactor (issue #994): per-edge resolution now
+ * routes through @liendev/parser's guarded, multi-language path-matching
+ * primitives (`importMatchesTarget` and friends) instead of this module's own
+ * `resolveImportPath` (deleted — it only ever handled relative JS/TS imports
+ * with a hardcoded 6-extension list). The BFS traversal itself
+ * (`bfsTransitiveCallers`, via `walkBounded`) already came from parser and is
+ * untouched: what changed is how a single edge gets resolved, not how hops are
+ * counted. Review's graph stays symbol/call-site-level throughout (a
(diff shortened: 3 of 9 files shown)


alfhen:
Filed #1012 for the pre-existing parser-level path-matching limitation the Rust fixture (`pr1003-rust-config-guard.blast-radius.ts`) documents as a known gap — `matchesAtBoundaryPrecise`'s `maxLeadingSegments` cap breaks bare-module resolution when a crate's `src/` sits more than 1 directory below the repo root (as in `lien-review-testbed/rust/`). Not in scope for this PR; follow-up only.

Issue #1053, opened 2026-08-01
real-projects.test.ts edge-count floors only catch ~90-98% collapse — a real 78.5% regression (Monolog, #1028 gate) still passes


alfhen:
## Summary

The #1023/#1041 collapse-detector floors in
`packages/cli/test/e2e/real-projects.test.ts` (`expectedMinDependencyEdges`
per project) are explicitly documented as "collapse detectors, not precision
targets" — but adversarial testing (HEAD a676e663) shows the margin is wide
enough that they miss large, real, partial regressions, not just small ones.
Filing this so the gap is quantified and tracked, not to dispute the
documented design intent.

## Empirical result (real regression, not synthetic)

Reverting the PHP `namespaceStyleImports` gate from #1028/#1032 (flipping
`packages/parser/src/ast/languages/php.ts`'s `namespaceStyleImports: true`
back to `false` — the exact matcher-flag mechanism #1028 exists to gate) on
Monolog:

```
baseline:  186 edges, 85 orphans / 100 files checked
regressed:  40 edges, 85 orphans / 100 files checked   (-78.5%)
```

`expectedMinDependencyEdges: 20` for Monolog **still passed** (40 ≥ 20, 2x
margin) despite a real, measured 78.5% loss in resolved edges — the exact
regression class #1004/#1028 exist to catch a recurrence of.

## Arithmetic for other corpora (re-verified against current file)

| Project | floor | measured (comment) | loss needed to trip floor |
|---|---|---|---|
| Zod | 50 | ~1265 | 96.0% |
| Express | 50 | ~2924 | 98.3% |
| MediatR | 20 | ~578 | 96.5% |
| Requests | 20 | ~353 | 94.3% |
| Sinatra | 10 | ~109 | 90.8% |

A real 60% loss on any of these (e.g. Zod 1265→506) would leave the corpus
at roughly 10x the floor — nowhere close to tripping it.

## Aside: found while measuring — a stale comment

Monolog's own comment says "measured ~405 edges" but the actual 100-file
sampled measurement (same code, same sampling logic, unmodified HEAD) reads
186. Likely explanation: 232 files exceeds the 100-file sweep cap, so ~405
may be an earlier full-corpus (or differently-sampled) measurement rather
than this sampled path — not itself a regression, but worth refreshing so
the comment matches what the test actually measures today.

## Also checked

- Rust/Anyhow's floor (1) has a stale rationale comment ("measured ~39
[...]

Issue #1044, opened 2026-08-01
findDependents: re-export transitive resolution is non-deterministic across identical re-indexes (zod core.ts/classic/schemas.ts)


alfhen:
## Summary

While doing the before/after behavior-preservation verification for ADR-015
(#1038), I found that `findDependents`'s re-export transitive-dependent
resolution is **non-deterministic across repeated `lien index --force` runs
of the same, unmodified source tree** -- i.e. this is unrelated to the
ADR-015 change itself (confirmed: the ADR-015 branch only edits declarative
per-language boolean fields, touching zero scanning/BFS/re-export code).

## Repro

Corpus: `zod` (https://github.com/colinhacks/zod), any recent clone.

1. `lien index --force` the repo into a fresh, isolated `LIEN_HOME`.
2. Query `findDependents` for `packages/zod/src/v4/core/core.ts`.
3. Repeat steps 1-2 in a **new** fresh `LIEN_HOME` (same source tree, same
   built CLI/parser, nothing else changed).

Observed: whether `packages/zod/src/v4/classic/schemas.ts` appears in the
dependents list **flips** between runs. Confirmed flipping 2/2 times with the
exact same code and source tree (`present`, then `absent` on the very next
independent re-index). `result.truncated` is `false` in both cases, so it is
not the BFS `maxNodes` cap.

`classic/schemas.ts` never directly imports `core.ts` -- the log line `Found
73 additional dependents via re-export chains` confirms this specific
dependent is only reachable through `findTransitiveDependents`'s re-export
BFS (`dependency-analyzer.ts`), which visits re-exporters via
`buildReExportGraph`'s iteration over `allChunksByFile` (a `Map` whose
insertion order follows file-scan order). If that scan order isn't
guaranteed stable run-to-run (e.g. concurrent file reads completing in
varying order before being inserted into the map), the BFS's `visited` Set
can mark a file as visited via a different chain than expected, silently
dropping a legitimate dependent depending on which chain wins the race.

## Why this matters

Structural dependency analysis that gives a different answer for
`get_dependents` on the exact same code between two ordinary re-indexes is a
correctness/trust problem independent of any specific matching rule -- a user
re-running `lien index` (e.g. after a `git pull` that touched unrelated
files) could see a real dependent silently disappear or reappear with no
underlying code change.

## Suggested next step

Audit `buildReExportGraph`/`findTransitiveDependents`/`processTransitiveChunk`
[...]

Issue #1040, opened 2026-08-01
fix(parser): C# test-association resolution is 0/160 files on MediatR — enclosing-namespace test convention has no test-associations.ts case


alfhen:
## Summary

While extending the CLI E2E suite's output assertions (#1029, Workstream 3) to test associations, C# measures **0/160 files** with any resolved test association on MediatR — despite the corpus shipping a real, substantial `MediatR.Tests` project. Confirmed not a sampling artifact (checked all 160 indexed files via `findTestAssociationsFromChunks`, not just a 100-file sample).

## Mechanism

C#'s dominant convention doesn't need an explicit `using` for a test file to reference the code under test, the same underlying shape as Go's same-directory (#902) and Java's same-package (#925) no-import test conventions — but for C#'s **enclosing namespace**, not same-directory/same-package.

Repro: `test/MediatR.Tests/PipelineTests.cs`:

```csharp
namespace MediatR.Tests;

using System.Collections.Generic;
using System.Threading.Tasks;
using Shouldly;
using Xunit;
```

No `using MediatR;` anywhere in the file, yet it references `MediatR`'s root types (`IRequestHandler`, `IPipelineBehavior`, etc.) unqualified — legal because C# namespace lookup walks outward through enclosing namespaces (`MediatR.Tests` is lexically enclosed by `MediatR`), so nothing requires an explicit import. This is exactly the shape `hasSameDirectoryTestConvention`/`hasSamePackageTestConvention` exist to paper over for Go/Java in `packages/parser/src/test-associations.ts` — but there is no C# case.

C# **does** have a mechanism that models enclosing-namespace access already: `enclosingNamespaceAccess` (used by `dependency-analyzer.ts`'s general `findDependents` resolution, per #930 and friends — six issues' worth of real lexical namespace scoping with shadowing). That mechanism is not reused by `test-associations.ts`'s `collectImportMatchedTests`, which only knows about `hasGoSameDirectoryConvention`/`hasJavaSamePackageConvention`. So C#'s general dependency-edge resolution is comparatively strong (MediatR measures ~578 edges via `findDependents`), while its test-association resolution is silently zero.

## Distinction from existing known-zero cases

- **Kotlin** (`sameUnitAccessWithoutImport`, #1005) and **Swift** (#869) are already documented, accepted zero-test-association gaps.
[...]

Issue #1005, opened 2026-07-31
fix(parser): Java/Kotlin dotted FQN imports never map to file paths, and only C# has enclosing-namespace dependents — 100% orphan rate on JavaPoet/Klaxon


alfhen:
## Summary

Two separate mechanisms leave Java, Kotlin and Swift with **zero** dependency edges on real projects. Both are distinct from #1002 (PHP PSR-4) and #1000 (Rust `mod`), though they share the family.

Measured on the corpora the CLI E2E suite already clones:

```
corpus-java-javapoet      43 files   0 edges   43 orphans (100.0%)
corpus-kotlin-klaxon     104 files   0 edges  103 orphans (100.0%)
corpus-swift-swiftyjson   27 files   0 edges   27 orphans (100.0%)
```

All three E2E jobs pass — see #1004 for why.

## Mechanism 1 — dotted FQNs are never converted to paths (verified for Java)

Java/Kotlin imports are dot-separated fully-qualified names. `normalizePath` leaves them untouched:

```
sample file: src/test/java/ClassNameNoPackageTest.java
sample imports: ["com.google.common.truth.Truth.assertThat","com.google.common.truth.Truth",
                 "com.squareup.javapoet.ClassName","org.junit.Test"]

normalizePath("com.squareup.javapoet.Util.checkArgument") = com.squareup.javapoet.Util.checkArgument   <-- dots intact
normalizePath("src/main/java/com/squareup/javapoet/TypeName.java") = src/main/java/com/squareup/javapoet/TypeName
```

A dot-separated specifier can never match a slash-separated path, so no Java import resolves to any file — at the repo root **or** at the source root (`src/main/java`), both verified.

Contrast PHP, where the extractor emits backslashes and `normalizePath` converts them (`Psr\Log\LoggerInterface` -> `Psr/Log/LoggerInterface`), and Python, which has a dedicated dotted-path matcher (`path-matching.ts:505+`, "django.http -> django/http/response.py"). Java/Kotlin have neither.

Note the fix also needs the source-root strip (`src/main/java/`, `src/main/kotlin/`) — the FQN `com.squareup.javapoet.TypeName` corresponds to `src/main/java/com/squareup/javapoet/TypeName.java`. That is build-system knowledge (Maven/Gradle), analogous to what `php-psr4.ts` does for Composer and `rust-crate-map.ts` for Cargo. The conventional Maven/Gradle layout is standard enough to hard-code as a candidate prefix list without reading a manifest.

[...]

PR #1061, opened 2026-08-03
fix(parser): resolve Java/Kotlin dotted-FQN imports via Maven/Gradle source roots (#1046, #1005 Mechanism 1)

### LINKED ISSUE #1046: get_dependents: Java/Kotlin dotted-FQN imports never resolve via importMatchesTarget (Strategy 5 gated to Python-only by #929)
## Summary

While verifying ADR-015 (#1038, PR #1045), independent runtime probing
(toggling flags against the real built `packages/parser/dist` matcher, not
just reading source) found that Java and Kotlin's normal, precise, per-file
`import` statements **never resolve** through `importMatchesTarget` today --
not a gap in some edge case, the ordinary case.

## Root cause

`matchesFile`'s Strategy 5 (`matchesPythonModule`) is the only strategy that
understands a DOTTED fully-qualified name (`com.squareup.javapoet.Util`).
Java's `JavaImportExtractor` and Kotlin's `KotlinImportExtractor` both store
the raw dotted FQN verbatim (`java.ts:312-321`, `kotlin.ts` mirrors it) --
neither converts it to a slash path, and no other strategy in
`path-matching.ts` understands dots. #929 gated Strategy 5 off for every
non-Python importer via `hasPythonModuleSemantics`
(`path-matching.ts:550`, `allowPythonModuleMatching`), to fix a real false-hub
bug on TypeScript (a barrel import coincidentally matching Python's dotted-
module syntax). That fix was correct for the TypeScript case, but it also
cut off the *only* mechanism that ever let a Java/Kotlin dotted import
resolve at all -- there never was a Java/Kotlin-specific equivalent.

## Confirmed

Toggling `allowPythonModuleMatching` on for a Java/Kotlin importer while
probing the real matcher:

```
com.squareup.javapoet.Util -> src/main/java/com/squareup/javapoet/Util.java
  Strategy 5 ON:  true
  today (OFF):    false

com.beust.klaxon.internal.ConverterFinder -> (real klaxon file)
  Strategy 5 ON:  true
  today (OFF):    false
```

## Why the existing test suite doesn't catch this

`path-matching.test.ts`'s own Java/Kotlin fixture
(`testMatchesFile('com.example.Utils', 'com/example/Utils.kt') === true`,
around line 421) calls the low-level `matchesFile` directly, which defaults
`allowPythonModuleMatching` to `true` (its own pre-#929 behavior, preserved
for direct callers) -- it never exercises the gated, language-aware
production path (`importMatchesTarget`) that `get_dependents` actually uses.
So the test suite asserts the WRONG layer and stays green while the real
path returns nothing.

## Impact

`get_dependents` for Java/Kotlin currently relies entirely on
`samePackageTestConvention` (Java, test-association only) /
`sameUnitAccessWithoutImport` (Kotlin, honesty-caveat only) for any
recovery -- there is no real per-file dependent recovery for an ordinary,
explicit, cross-package `import` statement in either language. This is
consistent with (and a plausible root cause of, not just correlated with)
`KNOWN_ZERO_EDGE_LANGUAGES` already tracking Java/Kotlin/Swift as zero-edge
in `packages/cli/test/e2e/real-projects.test.ts` (#1005).

## Suggested next step

A Java/Kotlin-aware dotted-FQN matcher, mirroring `matchesPythonModule`'s
shape but gated to Java/Kotlin importers only (the same per-language-gate
pattern `singleFileImports`/#887, the Python guard/#929, and
`namespaceStyleImports`/#1028 all already use) -- not reusing Strategy 5
unconditionally, which is exactly what caused the original bug it was
gated to fix.

## Scope note

Not blocking ADR-015/#1038 (PR #1045) -- that PR's own three fields
(`wholeModuleImports`, `singleFileImports`, `namespaceStyleImports`) don't
touch Strategy 5 or `allowPythonModuleMatching` at all, and this gap
pre-dates that PR by #929 (well before this ADR). Filed separately per that
PR's own verification requirements, discovered as a side effect of
independently probing the real matcher rather than trusting the existing
test suite's assertions about it.

alfhen:
## Summary

Fixes #1046 and #1005's Mechanism 1: Java and Kotlin store the raw dotted
FQN specifier verbatim (`com.squareup.javapoet.TypeName`), and `matchesFile`'s
only dotted-aware strategy (Python module matching) is correctly gated off
for non-Python importers since #929 — so a Java/Kotlin cross-package
`import` never resolved to a dependent edge at all. `get_dependents` was
silently at or near 0 edges for 2 of 11 supported languages on real corpora.

**Mechanism 2 (#1005's same-package-without-import gap) is explicitly out of
scope** — noted below, not touched.

## What was built

`packages/parser/src/jvm-source-root.ts` — locates a project's conventional
Maven/Gradle source-set directories (`src/main/java`, `src/main/kotlin`,
`src/test/java`, `src/test/kotlin`, found at **any depth** via glob, since a
multi-module Gradle build nests each module's source set under its own
subdirectory — confirmed on Klaxon's real layout, `klaxon/src/main/kotlin/...`,
not a top-level `src/main/kotlin/...`) and resolves a dotted FQN to a
workspace-relative path **only when the resolved candidate exists on disk as
a real `.java`/`.kt` file**. Mirrors the existing `php-psr4.ts` /
`rust-crate-map.ts` / `python-src-layout.ts` pattern exactly.

Wired into `resolveImportSpecifier` (`ast/symbols.ts`) as a fourth
manifest-root resolution step, gated to `java`/`kotlin` in `chunker.ts`'s
`buildManifestRoots` (refactored into a per-language builder dispatch table
in the same change, to keep that function's own complexity flat as
languages are added — `lien delta` flagged the original if-chain approaching
threshold).

**This does NOT touch `path-matching.ts`.** Resolution happens entirely
upstream of `matchesFile`: once a dotted FQN resolves to a real file path,
it satisfies the existing exact-match strategy directly. No new
fuzzy-matching strategy, no Strategy-5 re-enable, no new gate in
`importMatchesTarget` — the ADR-015 matcher-path fields (`wholeModuleImports`,
`singleFileImports`, `namespaceStyleImports`) are unaffected and stay
exactly as they were for both languages.

### Deliberately NOT resolved (scope boundary, avoiding the #928 trap)

- **Wildcard package imports** (`import com.example.*`) — these name a
  package *directory*, not a single file. Resolving them would require a
  different (directory-containment) mechanism; left as an honest gap rather
  than guessed at.
- **Kotlin function/property-level dotted imports** (e.g.
  `import com.beust.klaxon.internal.firstNotNullResult`) — `KotlinImportExtractor`
  has no `static`-keyword-equivalent marker to tell a truncated top-level
  declaration apart from a package path (documented in its own doc comment,
  pre-existing gap, not introduced by this PR).
- Both are existence-gated no-ops: the specifier passes through unresolved,
  exactly as before this fix. Confirmed via unit tests
  (`jvm-source-root.test.ts`).

### Unit + integration tests (packages/parser)

- `jvm-source-root.test.ts` (16 tests) — resolver unit tests: single-module
  and multi-module (Klaxon-shaped) layouts, mixed Java/Kotlin interop,
  external-library non-fabrication, basename-collision safety, wildcard
  non-resolution, static-member-candidate handling, cache behavior.
- `jvm-source-root-integration.test.ts` (4 tests) — end-to-end through the
  **real production path** (`chunkByAST` → `resolveImportSpecifier` →
  `analyzeDependencies`, the same function `get_dependents` calls), not a
  hand-built chunk fixture. **Verified two of these tests fail on pre-fix
  code**: temporarily commented out the `java`/`kotlin` entries in
  `chunker.ts`'s dispatch table and reran — both "resolves a real
  cross-package import" tests failed (`expected [] to include '...'`) while
  the two anti-fabrication tests stayed green either way, as they should.
- `path-matching.test.ts` — the existing Java/Kotlin fixture at
  `testMatchesFile('com.example.Utils', 'com/example/Utils.kt')` calls
  `matchesFile` directly (defaults `allowPythonModuleMatching: true`),
  which never exercised the gated `importMatchesTarget` production path —
  exactly the flaw #1046 identified. Added a clarifying comment plus a new
  `#1046 Java/Kotlin dotted-FQN specifiers` block that pins the real
  boundary: a raw (unresolved) dotted specifier still correctly returns
  `false` through `importMatchesTarget` (Strategy 5 stays closed), while
  the *resolved* slash-path form (what `jvm-source-root.ts` produces)
  matches through the unmodified existing strategies.

Full parser suite: 56 files / 1702 tests passing.

### Real corpora, before/after, 5 repeated measurements each

Shallow-cloned JavaPoet, Klaxon, SwiftyJSON directly (not just via the E2E
harness) and measured with an isolated `LIEN_HOME` per run, using the same
`computeDependencyStats`/`countDependents` logic the E2E suite uses
(`findDependents`, the exact production engine). Numbers were **identical
across all 5 repeats** for both languages — no #1044 flap observed for
these corpora.

| Corpus | Before | After (5/5 identical) |
|---|---|---|
| JavaPoet (43 files) | 0 edges, 43 orphans (100.0%) | **18 edges**, 40 orphans (93.0%) |
| Klaxon (104 files, 100 sampled) | 0 edges, 100 orphans (100.0%) | **4 edges**, 97 orphans (97.0%) |
| SwiftyJSON (27 files) | 0 edges, 27 orphans (100.0%) | 0 edges, 27 orphans (100.0%) — unchanged, expected (Swift untouched) |

Baseline measured against a separate `git worktree` checked out at this
branch's pre-fix parent commit (03232e60), built independently — not
assumed from the issue text.

**JavaPoet caveat, exactly as flagged in the brief**: JavaPoet's 43 files
sit in a single flat package (`com.squareup.javapoet`), so most cross-file
references carry **no import statement at all** (Mechanism 2, out of
scope) — this fix cannot and does not touch that. The 18 resolved edges
are real: JavaPoet's ~13 test files cross-package-import their subjects
under `src/test/java` → `src/main/java`, and those now resolve correctly
(verified end-to-end via a live MCP `get_dependents` call — see below).

**Klaxon's 4 edges, verified by hand**: traced every one of Klaxon's 8
real non-wildcard cross-package `import` lines by hand. 3 resolve to
`internal/ConverterFinder.kt`/`token/Token.kt`/`Converter.kt` (exactly 4
edges once de-duplicated by target file — `TokenType` lives in the same
file as `Token`). The remaining lines are 3 wildcard imports and 1
function-level import, both deliberately unresolved (see scope boundary
above). This is a **correct, fully-accounted-for count**, not a partial
fix undercounting silently.

### Live dogfooding — real MCP `get_dependents` call

Indexed JavaPoet and called the actual MCP tool (not the test harness):

```
get_dependents({filepath: "src/main/java/com/squareup/javapoet/ClassName.java"})
→ dependentCount: 1, dependents: [{filepath: "src/test/java/ClassNameNoPackageTest.java", isTestFile: true}]
```

And confirmed the pre-existing (not new — already shipped, verified still
correct) `attributionCaveat` honesty mechanism still fires correctly for a
same-package zero-dependent file post-fix:

```
get_dependents({filepath: "src/main/java/com/squareup/javapoet/TypeSpec.java"})
→ dependentCount: 0, attributionCaveat: { reason: "dependent-attribution-incomplete", note: "...Java/Kotlin's same-package visibility..." }
```

### Regression check — all 11 real-project corpora, full E2E suite

`npm run test:e2e -w packages/cli` (requests/zod/express/monolog/anyhow/chi/
mediatr/sinatra/javapoet/klaxon/swiftyjson): **198/198 tests passing**, no
regressions. `chunker.ts`'s `buildManifestRoots` refactor (if-chain →
per-language dispatch table) is behavior-preserving for every other
language — confirmed both by this full E2E run and by the untouched
per-resolver unit test suites (php-psr4, python-src-layout, rust-crate-map,
go-module, js-directory-index) all staying green.

`npm run test:e2e:java -w packages/cli` and
`npm run test:e2e:kotlin -w packages/cli` also run individually per the
task brief — both green, matching the table above exactly.

### E2E suite updated (`real-projects.test.ts`)

- Java/Kotlin graduated out of `KNOWN_ZERO_EDGE_LANGUAGES`'s exact-zero
  tripwire into real `expectedMinDependencyEdges` floors — **tight**, not
  the file's usual "far below measured" convention (that convention is
  #1053's own root cause elsewhere in this file): JavaPoet floor `9`
  (50% of measured 18), Klaxon floor `2` (50% of measured 4) — each tight
  enough to catch a ~50% regression, not just total collapse.
- Swift stays in the tripwire set with an explicit note explaining why
  (`wholeModuleImports` means there's no dotted FQN for a source-root
  strip to resolve in the first place — structurally unaffected by this
  fix, confirmed by direct measurement, not assumed).
- `KNOWN_ZERO_TESTASSOC_LANGUAGES` (Kotlin/Swift/csharp) — **untouched**,
  different set, different mechanism (test association, not dependents);
  csharp's half is being fixed concurrently elsewhere (#1040).

### Caveat/honesty work (#1005's suggested step 1)

Already shipped by prior work (`hasDependentAttributionBlindSpot` /
`dependentAttributionIncomplete` / MCP `attributionCaveat`, already wired
for Java/Kotlin/Swift via `samePackageTestConvention`/
`sameUnitAccessWithoutImport`/`wholeModuleImports`) — verified above via a
live MCP call that it still fires correctly post-fix. No new work needed
here.

## Gates

`npm run format:check`, `npm run lint` (0 errors, pre-existing warnings
only), `npm run typecheck`, `npm run build`, `npm test` (all packages
green), `lien delta` (no complexity-affecting changes) — all pass.

---

### Lien Review

| 🧠 mental load | 1 | +0 |

> [!WARNING]
> **3 issues found** · Low Risk
>
> This PR adds a JVM source-root resolver that maps Java/Kotlin dotted fully-qualified import specifiers (e.g., `com.squareup.javapoet.TypeName`) to workspace-relative file paths via conventional Maven/Gradle source-set directories. Resolution is existence-gated and falls back to the original specifier when no real `.java`/`.kt` candidate exists, mirroring the existing PHP PSR-4 / Rust crate-map / Python src-layout pattern. The change is wired into `resolveImportSpecifier` as a fourth manifest-root step and is scoped only to Java/Kotlin through a refactored per-language `MANIFEST_ROOT_BUILDERS` dispatch table in `chunker.ts`. No regressions are expected: unresolved specifiers (external libraries, wildcard package imports, Kotlin function-level imports, non-standard layouts) pass through unchanged, and the existing `matchesFile` exact-match path handles resolved slash paths without any new fuzzy-matching strategy.
>
> - Added `packages/parser/src/jvm-source-root.ts` with `resolveJvmSourceRoots` and `resolveJvmSourceRootImport`, including per-workspace-root caching and Unicode-aware dotted-FQN validation.
> - Registered Java and Kotlin in `MANIFEST_ROOT_BUILDERS` (`packages/parser/src/ast/chunker.ts`) and wired `resolveJvmSourceRootImport` into `resolveManifestRoot` (`packages/parser/src/ast/symbols.ts`).
> - Added unit tests (`jvm-source-root.test.ts`) and end-to-end integration tests (`jvm-source-root-integration.test.ts`) covering single/multi-module layouts, mixed Java/Kotlin, basename-collision safety, wildcard non-resolution, and anti-fabrication cases.
> - Updated E2E floors in `real-projects.test.ts` to reflect measured Java/Kotlin dependency edges and clarified the remaining same-package-without-import gap as out of scope.

**Findings (3)**

- 🟡 **bug** in `resolveImportSpecifier` — `packages/parser/src/ast/symbols.ts:152`
  The touched JSDoc for `resolveImportSpecifier` describes step 4 as handling only 'PHP PSR-4, Go module prefix' manifest roots, but the implementation also resolves Python src-layout and JVM source roots, leaving callers misinformed about active strategies.
  💡 *Update the step 4 description to list all supported strategies: 'Manifest-declared import roots (PHP PSR-4, Go module prefix, Python src-layout, JVM source roots) against `manifestRoots`.'*

- 🟡 **bug** in `resolveManifestRoot` — `packages/parser/src/ast/symbols.ts:180`
  The touched JSDoc for `resolveManifestRoot` says it applies 'step 3 (manifest-root resolution)' of `resolveImportSpecifier`, but manifest-root resolution is actually step 4 in the parent function's numbered list.
  💡 *Change the comment to 'Apply step 4 (manifest-root resolution) of `resolveImportSpecifier`.'*

- 🟡 **bug** — `.changeset/fix-1046-jvm-source-root.md:9`
  The changeset claims 'No regressions across the other 9 real-project corpora' but enumerates 11 names including JavaPoet and Klaxon, contradicting the 'other 9' count.
  💡 *Reword to 'No regressions across the other 9 real-project corpora (requests, zod, express, monolog, anyhow, chi, mediatr, sinatra, swiftyjson)' or 'No regressions across all 11 real-project corpora (...).'*

<sup>Reviewed by [Lien Review](https://lien.dev) for commit d576c8c. Updates automatically on new commits.</sup>

---

Attested: degraded:budget_starved · 2/2 comments posted · 602K/763K tokens


alfhen:
Rebased onto `origin/main` (picked up #1044 re-export-BFS determinism and
#1050/#1051 worktree fixes). Re-ran the full verification from the rebased
tip:

- Full gate chain (`format:check`, `lint`, `typecheck`, `build`, `test`,
  `lien delta`) — all green.
- Full E2E suite across all 11 real-project corpora — **198/198 passing**.
- Java/Kotlin numbers are **unchanged** from the pre-rebase measurement
  (JavaPoet: 18 edges/93.0% orphan; Klaxon: 4 edges/97.0% orphan) — expected,
  since #1044 only affects transitive re-export-chain resolution, and
  neither corpus exercises that path. This confirms this fix and #1044 are
  independent, as expected.
- Other languages' numbers moved as #1044 predicted (e.g. Zod: 336 edges
  on this run's 100-file sample, up from whatever the pre-#1044 flap
  produced) — not a regression from this PR, and not something this PR's
  floors touch.

No changes needed to this PR's own floors (JavaPoet 9, Klaxon 2) — both
still comfortably pass against the rebased numbers.

alfhen:
Fixed in f4586fbf: `DOTTED_FQN_PATTERN` is now Unicode-aware (`\p{L}`/`\p{N}` with the `u` flag) per the review finding on line 134. Added a regression test with a Unicode package/class name (`例.クラス.Foo`). Note this pattern only gates whether a specifier is even attempted for resolution — the existence check on disk is what actually determines a match either way, so this was a coverage gap (real files silently unresolved), not a fabrication risk.

alfhen:
Fixed in d576c8c6: updated `resolveImportSpecifier`'s step-4 doc comment to list all four manifest-root strategies (PHP PSR-4, Go module prefix, Python src-layout, JVM source roots), and corrected `resolveManifestRoot`'s self-reference from "step 3" to "step 4" to match the parent's numbered list. Both were real doc-drift (the step-numbering one pre-dates this PR; the missing-strategies one pre-dates this PR for Python and is now also missing JVM).

alfhen:
Triaging the latest review pass (commit d576c8c, attested `degraded:budget_starved`) — all 3 findings are stale/incorrect, not new issues:

1. **`resolveImportSpecifier` step-4 comment** — this is exactly the finding I already fixed in d576c8c6 itself. Current content at `symbols.ts:155-156`:
   ```
   * 4. Manifest-declared import roots (PHP PSR-4, Go module prefix, Python
   *    src-layout, JVM source roots) against `manifestRoots`.
   ```
   Already lists all four strategies. Dismissing as stale (re-surfaced from the pre-fix commit, consistent with the budget-starved degradation).

2. **`resolveManifestRoot` step 3→4 comment** — same commit, same fix. Current content at `symbols.ts:191`:
   ```
   /** Apply step 4 (manifest-root resolution) of `resolveImportSpecifier`. */
   ```
   Already says "step 4". Dismissing as stale.

3. **Changeset "other 9" vs "11 names"** — checked directly: the changeset's parenthetical lists exactly 9 names (`requests, zod, express, monolog, anyhow, chi, mediatr, sinatra, swiftyjson`), not 11, and neither JavaPoet nor Klaxon appear in it (correctly — they're the two languages this PR *does* change, not part of the "other" set). The finding's premise is factually wrong. Dismissing.

No code changes needed for this pass. Merging on green CI + this triage.

Issue #1072, opened 2026-08-03
mcp: search_code asserts dependentCount as fact with no caveat machinery (#1014 inverted)


alfhen:
## Summary

`search_code` returns `dependentCount` in every result's metadata, documented as *"informational and unconditional"*, and **has no `attributionCaveat` machinery on that path at all**. `get_dependents` has it; `search_code` does not.

Consequence today: on any C#, Java, Kotlin, Swift, Go or Rust repository, **every `search_code` result an agent sees carries `dependentCount: 0`, asserted as fact, on a healthy freshly-indexed store** — because the underlying resolver is broken (companion issue). A bare `0` reads as *"nothing depends on this file"*.

This is **#1014 inverted**: not a caveat that over-fires and gets trained out as noise, but one that never fires at all.

## Where

- Emitted: `packages/cli/src/mcp/utils/metadata-shaper.ts:44`, allowlisted at `:82`, documented as informational and unconditional.
- Handler: `packages/cli/src/mcp/handlers/search-code.ts` — no `attributionCaveat`, no `note`, no degraded marker on any branch.
- Contrast: `packages/cli/src/mcp/handlers/get-dependents.ts` carries the full caveat vocabulary (`attributionCaveat`, `dependentAttributionIncomplete`, five `AttributionCaveatReason` values).

## Why this is a policy violation, not just a gap

`CLAUDE.md`'s Index-State Honesty section states the constraint directly — *"A read-only, index-backed command MUST NEVER produce a confident answer when it has no data, or the wrong data, to answer from"* — and its disposition table's **MCP tool** row requires an explicit `note`/`attributionCaveat`, *"never a bare empty result"*.

`dependentCount: 0` is precisely a confident answer produced from the wrong data. And per `docs/architecture/index-state-honesty.md`'s own framing, the failure is not the zero — it is that **the zero is indistinguishable from a real one**.

## This is independent of the resolver fix, and must be fixed even after it

Worth stating explicitly, because it's tempting to treat this as a duplicate of the companion issue: **fixing the resolver does not close this.**

After the resolver works, some files will *genuinely* have zero dependents. A consumer still cannot distinguish:

1. "resolved successfully; nothing depends on this file" — a real, useful answer
2. "could not resolve this language's import specifiers" — a blind spot
3. "this path isn't indexed" — an index-state problem

Those need different consumer behaviour, and today all three render as `0`.

## Suggested fix

Two acceptable shapes, in preference order:

[...]

Issue #1067, opened 2026-08-03
refactor(parser): 'bare root import credits a whole directory' is now fixed independently in Go and Rust — name the class before a third language reproduces it


alfhen:
Filed by the coordinator of the 2026-08-03 resolution wave, after watching two agents independently discover and fix the same bug shape in two languages within the same hour.

## The pattern

A language lets you import a *container* (a package, a crate) without naming the file inside it. The resolver, having no file to point at, credits **everything** in that container — producing an identical dependent list for every file in it. That is a false hub: confidently wrong data, not a gap.

Three confirmed instances, all in the #928 family:

| # | language | mechanism | symptom | fixed by |
|---|---|---|---|---|
| #1008 | Rust | `mod x;` resolution | 5 fabricated self-edges on anyhow | #1020 (exact-equality self-edge guard) |
| **#1039** | **Go** | bare root-module self-import (`import "github.com/go-chi/chi/v5"` from a subpackage) | every root file credited to every subpackage importer | **#1064** (`go-root-package-signals.ts`) |
| **#1056** | **Rust** | bare crate-root import (`use serde_derive::Serialize;`) | `de.rs` and `dummy.rs` both returned the identical 144-file list | **#1065** (`rust-crate-exports.ts`) |

## Both independent fixes converged on the same solution

Neither agent saw the other's code. Both concluded that the tempting fix — resolve the bare specifier to the container directory, the way a qualified subpackage/submodule import already resolves — **must be rejected** because it fabricates. Both instead built a narrowly-gated **export-lookup recovery**: find which single file in the container actually declares the imported symbol, and when that cannot be determined, **emit nothing and let the honest-caveat machinery report the blind spot.**

The Go fix's own test is literally named for the Rust issue (`"the #1056 failure shape, checked explicitly"`), and the Rust fix's PR body cross-references the Go one. They arrived at:

- resolve via export lookup, never by crediting the container
- require the *symbol* to be found, not merely the container to exist
- refuse to guess on ambiguity (Go: a name declared by two root files recovers nothing for either; Rust: honest null on a miss)
- gate to a fallback tier that only fires when the primary import graph found nothing
- **acceptance criterion is a distinctness check, not an edge total** — a count cannot distinguish "fixed" from "fabricating differently"

That last point is the most transferable: both PRs added an explicit "two unrelated files return disjoint dependent lists" test.

## Why file this rather than refactor now

[...]

Issue #1039, opened 2026-08-01
fix(parser): resolveGoModuleImport never resolves a bare root-module self-import — Chi's 94% orphan rate is a real gap, not leaf-heaviness


alfhen:
## Summary

Chi's dependency-edge orphan rate (93-94%, ~11-20 edges across 86 files, flagged as an open anomaly in #1004/#1023) is **not** (only) genuine leaf-heaviness. A real, narrow, previously-unfixed gap in `resolveGoModuleImport` (`packages/parser/src/go-module.ts`) means **any Go file that imports its own module's ROOT package** (as opposed to a subpackage) never resolves to the file(s) at the repo root that define what it imports.

## Repro (go-chi/chi)

`go.mod`: `module github.com/go-chi/chi/v5`. `middleware/clean_path.go`:

```go
import (
	"net/http"
	"path"

	"github.com/go-chi/chi/v5"
)

func CleanPath(next http.Handler) http.Handler {
	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		rctx := chi.RouteContext(r.Context())
		...
```

The indexed chunk's `imports` metadata retains the raw specifier unchanged: `["github.com/go-chi/chi/v5"]`. `context.go` (which exports `RouteContext`) shows **zero** dependents despite 7 middleware files (`clean_path.go`, `compress.go`, `get_head.go`, `profiler.go`, `strip.go`, `url_format.go`, `supress_notfound.go`) and most of `_examples/*`'s `main.go` files importing the bare module path this way. A repo-wide grep of Chi finds **35** occurrences of the literal bare-root import `"github.com/go-chi/chi/v5"` — none of them resolve.

Measured via `findDependents` (the same engine `get_dependents`/the CLI E2E suite use), sweeping all 86 indexed files: **11 total edges, 81 orphans (94.2%)**. The handful of edges that DO resolve are all within one self-contained example (`_examples/versions/...`) that uses purely relative-feeling subpackage imports — every root-package (`chi.go`, `mux.go`, `tree.go`, `chain.go`, `context.go`) and every `middleware/*.go` file shows exactly 0 dependents.

## Root cause

`resolveGoModuleImport` (added by #867 to fix Go test-association's manifest-root mapping) strips a module's own prefix from a subpackage import (`github.com/go-chi/chi/v5/middleware` → `middleware`), but is a deliberate no-op when the specifier **equals** the module prefix exactly (a root-package import, no trailing path segment):

```ts
[...]

Issue #1015, opened 2026-07-31
get_dependents symbol-level usages are silently 0 for class/struct/interface names across languages (methods and functions are exact)


alfhen:
## Summary

`get_dependents({filepath, symbol})` returns `totalUsageCount: 0` and empty per-dependent `usages` whenever `symbol` names a **type** (class, struct, interface, enum) — with **no caveat**. For functions and methods the same tool is exact.

An agent asking *"who uses class X before I rename it"* gets a confident zero.

## Evidence (clean scratch index, not the shared/overlay one)

| query | dependents | totalUsageCount | caveat | ground truth |
|---|---|---|---|---|
| TS interface `User` (`typescript/src/types.ts`) | 7 | **0** | none | 59 `User` references outside `types.ts` |
| PHP class `PricingService` | 2 | **0** | none | 12 (`use App\Services\PricingService;`, constructor type hints, 9 `$this->pricingService->` calls) |
| PHP class `PaymentService` | 1 | **0** | none | non-zero |
| Rust struct `Config` | 0 | **0** | none | non-zero |
| Rust enum `AnalyzerError` | 0 | **0** | none | non-zero |
| Python class `Record` | 7 | **3** | none | 4 dependents report 0 |

Contrast — function/method symbols are accurate:

| query | totalUsageCount | matches ground truth |
|---|---|---|
| PHP `formatPrice` | 5 (ProductController 4 / OrderController 1) | exact — lines 31, 79, 80, 89 and 67 |
| PHP `applyDiscount` | 2 | exact |
| PHP `calculateOrderTotal` | 1 | exact |
| TS `getUser` | 6 | exact |
| Rust `generate_report` | 1 | exact |

Note the shape: `dependents` is often **right** while `usages` is empty — so the file-level answer is fine and only the symbol-level detail collapses. That makes it easy to miss, because the response looks populated.

## Why

Usage attribution is call-site driven. Nothing "calls" a class by its class name: types appear as constructor invocations (`new Order()`), type hints, `extends`/`implements` clauses, static access, and generic parameters — none of which surface as a `callSite` naming the type.

This is the same root shape as the review-side gap fixed in #1011 (`buildImportOnlyEdges`) and the one described in #1013 for Ruby: **the exported unit is a type, the called unit is a method.** Here it bites the parser/CLI side rather than the review side.

## Why it matters

[...]

PR #1086, opened 2026-08-04
fix(parser,cli,review): name the fallback behind an inferred dependent (#1018)

### LINKED ISSUE #1018: doc-truth: #1011 claims EdgeProvenance is surfaced in get_dependents JSON — it is review-only and structurally cannot reach the CLI
## Summary

#1011 (Phase 5) introduced `EdgeProvenance` — a six-level tier (`same-file` / `import-verified` / `import-only` / `symbol-name-match` / `oop-method-import` / `namespace-inferred`) letting a consumer weight a verified import edge above a name-matched guess.

Its PR description and summary state this is surfaced **"as a new Confidence column in `<blast_radius>` and in `get_dependents`'s JSON."**

The first half is true. **The second half is false**, and not by a small margin — it is structurally impossible as the packages are currently wired.

## Verification

```
$ grep -rn "provenance\|EdgeProvenance" packages/cli/src/mcp/
(no matches)

$ grep -rn "EdgeProvenance" packages/review/src/index.ts
packages/review/src/index.ts:76:  type EdgeProvenance,

$ grep -rn "@liendev/review" packages/cli/src/
(no matches)
```

`cli` never imports `review`, so a type defined and exported only from `packages/review` cannot appear in `get_dependents`'s output.

What `get_dependents` actually emits per dependent is unchanged (`packages/cli/src/mcp/tools.ts:152`):

```
dependents[]: { filepath, isTestFile, usages[]?, confidence?: "inferred" }
```

That `confidence: 'inferred'` is the **pre-existing** single-bit marker from C# type-reference recovery (#930), not the new six-level tier. It predates #1011 entirely.

## Why it matters

The distinction is exactly the one #994 Phase 5 set out to fix: an agent calling `get_dependents` still cannot tell a verified import edge from a weaker guess beyond the one C#-specific bit. Anyone reading #1011's summary would reasonably conclude that gap is closed for MCP consumers. It is closed for the **review engine** only.

There is a second-order cost too: this claim is now in the merged PR body and will be read as history.

## Two separable pieces of work

1. **Correct the record (small).** Fix the claim in #1011's PR body / any changeset or doc that repeats it. The review-side improvement is genuine and substantial — it just needs stating accurately.
2. **Actually plumb it (larger, optional).** Getting provenance into `get_dependents` means either lifting the tier into `@liendev/parser` (where `findDependents` lives after #992, and which `cli` already imports) or widening parser's existing `confidence` field. The parser is the natural home — `review` and `cli` would then share one vocabulary instead of two, which is the same consolidation argument the whole #994 campaign rests on.

Worth noting: parser already has adjacent vocabulary (`confidence: 'inferred'`, `attributionCaveat` reasons, `symbolAttributionDegraded`). A third independent confidence vocabulary in `review` is itself the N-sites pattern — see #995.

## Provenance

Found by the post-refactor dogfood sweep for #994, then verified directly against the source. I had also relayed the inaccurate claim myself before checking it. Related: #1011, #930, #995, #1017.

alfhen:
Closes #1018.

## Verdict first: the three vocabularies were not merged, and the reason is a shipped bug they were hiding

#1018 asked for one canonical provenance/confidence vocabulary in `parser` that `cli` and `review` consume instead of parallel dialects. **I'm declining that**, with reasons below — and the investigation turned up a live, agent-facing falsehood that the merge framing had obscured. That defect is what this PR fixes.

## The canonical type

```ts
// packages/parser/src/inferred-dependent-mechanisms.ts
export type InferredDependentMechanism = 'csharp-type-reference' | 'go-root-package-export';

export const INFERRED_DEPENDENT_MECHANISMS: Record<
  InferredDependentMechanism,
  { languageLabel; importGraphBlindSpot; recovery; residualRisk }
> = { ... };
```

plus `DependentInfo.inferredVia?: InferredDependentMechanism`, written only by `inferredDependent()` — the single constructor for the `confidence`/`inferredVia` pair.

This is a **refinement inside one existing value**, not a fourth dialect. `confidence?: 'inferred'` is byte-identical in meaning and value space; all that's added is *which* fallback produced it, a fact `parser` always knew and threw away at its own boundary.

## The bug, reproduced on `origin/main`

`confidence: 'inferred'` was single-valued, so every prose surface had to name a mechanism from memory. #930 shipped one (C#), so six surfaces hard-coded C#. #1039/#1064 added Go's root-package export lookup — same marker, same `dependent-attribution-partial` reason — and updated none of them, because **nothing forced it**: `Record<AttributionCaveatReason, string>` guards the set of *reasons*, and no reason was added. Only an existing reason's mechanism coverage widened.

Real `go-chi/chi` clone, real `lien serve` over stdio, `origin/main` (`3d78d6b0`):

```
===== context.go =====
dependentCount: 11
attributionCaveat.reason: dependent-attribution-partial
attributionCaveat.note: The import graph found zero import-based dependents for context.go
  (its language, C#, lets real callers use its exports with no per-file import naming it at
  all — "global using" / implicit enclosing-namespace member access), but 11 of the 11
  dependent(s) below were recovered by matching a uniquely-declared type name against other
  files' source text ...
```

Both halves false for Go. 24 of 24 recovered edges across `context.go`/`mux.go`/`chain.go`. The tool description carried the same claim (`C#'s global-using gap`), as did the docs page (`today only for C#`), the `AttributionCaveatReason` doc comment, and — via interpolation — the server instructions every MCP client reads on connect.

### After

```
===== context.go =====
dependentCount: 11
attributionCaveat.note: ... for context.go (its language, Go, gives a file outside the root
  package no way to spell a reference to it except the module's own bare import path
  ("github.com/org/mod"), which names no specific file), but 11 of the 11 dependent(s) below
  were recovered by looking up which single root-level file actually exports a distinctive
  symbol the importing file's own call sites name (#1039) — only files that already import
  the root package are considered, and an export declared by two root files is never guessed
  at ... this heuristic cannot distinguish a genuine root-package reference from an unrelated
  call that merely shares the same distinctive name ...
inferred dependents: [ 'middleware/clean_path.go [inferredVia=go-root-package-export]', ... ]
```

**Dependent lists are byte-identical before and after** (11 / 5 / 8 across the three files, same paths) — the change is prose accuracy plus one additive field, not semantics.

C# preserved exactly, real serilog clone, 254 files:

```
===== src/Serilog/Parsing/Alignment.cs =====
dependentCount: 6
note: ... (its language, C#, lets real callers use its exports with no per-file import naming
  it at all — "global using" / implicit enclosing-namespace member access), but 6 of the 6
  ... recovered by matching a uniquely-declared type name against other files' source text
  (#930) — the name is only trusted when exactly one C# file project-wide declares it ...
inferred: [ '.../MessageTemplateTextFormatter.cs [inferredVia=csharp-type-reference]', ... ]
```

Negative control (Go subpackage files, genuine resolved zero) — gains nothing:

```
===== middleware/middleware.go =====   dependentCount: 0   reason: undefined   inferred: []
===== middleware/logger.go =====       dependentCount: 0   reason: undefined   inferred: []
```

## The mapping, and why the merge was refused

There are **two axes and three scopes**, not one question asked three times. ADR-016 has the full argument.

| # | vocabulary | axis | scope | disposition |
|---|---|---|---|---|
| 1 | `confidence?: 'inferred'` (parser) | A: how was this edge resolved | per-edge, **file-level** | kept; refined by `inferredVia` |
| 2 | `EdgeProvenance` (review, 7 tiers) | A: how was this edge resolved | per-edge, **symbol-level** | kept as review's internal ranking detail |
| 3 | `AttributionCaveatReason` (cli, 5) | B: why can't this *answer* be trusted | per-answer | kept; already parser-derived |

**Axis A's two members must stay separate — merging makes one wrong.** `require-only` is the proof: a Ruby `require_relative` verifiably resolves the target *file*, so parser's file-level `findDependents` correctly treats that dependent as import-verified (no `confidence` marker), while naming no *symbol* at all, so review's symbol-level graph correctly ranks it imprecise (#1030's argued boundary). One shared bucketing predicate has to contradict one of them.

Separately, 4 of `EdgeProvenance`'s 7 tiers (`import-only`, `require-only`, `symbol-name-match`, `oop-method-import`) are distinctions that only arise *in* a symbol-level analysis — each records something different about what could be established for the specific symbol, a question parser's file-level pipeline never asks. (Not a claim that all four *verify* a symbol: `require-only` and `symbol-name-match` explicitly don't, which is why `SYMBOL_VERIFIED_BY_PROVENANCE` marks them `false`. The point is a file-level pipeline can't draw the distinction in either direction.) Hoisting the enum into `parser` publishes tiers nothing in `parser` emits — dead members, worse than two honest vocabularies.

**Axis B is already consolidated where it matters.** `AttributionCaveatReason` is not an independent taxonomy: `buildAttributionCaveat` derives all five reasons from fields `parser` already owns on `FindDependentsResult` (`targetIndexed`, `symbolAttributionDegraded`, `symbolFoundInFile`, `typeSymbolAttributionIncomplete`, `dependentAttributionPartial`, `dependentAttributionIncomplete`). The structural facts are in `parser`; `attribution-caveat-reasons.ts` is the prose + mutual-exclusion layer. Moving model-facing prose into the AST package gains nothing.

### The four blocked consumers, re-checked

- **#1072** — needed whole-store and whole-language scope. That's **scope 3**, neither axis; `note` + omission was the right call and #1078's reasoning holds. Its one genuinely shared piece (`hasDependentAttributionBlindSpot`) was already in `parser`. Nothing to do.
- **#1067** — **the real one, and now unblocked.** Its stated prerequisite (a first-class home for *"unique among indexed declarations in this scope; a reference may still resolve to an unindexed stdlib/third-party declaration of the same name"*, currently prose in four signal-module doc comments) is one new entry in `INFERRED_DEPENDENT_MECHANISMS`. Adding it makes all six surfaces correct with no further edits, and the compiler refuses the tier until its prose exists.
- **#1015** — needs `type-symbol-attribution-incomplete`'s prose updated when real type-reference counting lands (it currently says usages are "often 0"). A prose update then, not a vocabulary gap now.
- **Coverage-derived test associations** — the consumer motivating a "stronger than import-verified" direction. **That feature was evaluated and rejected** (no coverage artifact in 40 of 40 real checkouts). Building vocabulary for it would be YAGNI, so the "stronger" direction is deliberately not built.

## Forcing functions — both kept, and they now compose

- `Record<AttributionCaveatReason, string>` (#984) unchanged.
- `Record<InferredDependentMechanism, Descriptor>` is new and guards the axis the first one structurally cannot see. That composition is the fix: one guards the set of reasons, the other the set of mechanisms each reason covers.
- `isPreciseProvenance` was a hand-written three-way `||`, so an eighth `EdgeProvenance` tier would default silently to "not precise". Now `Record<EdgeProvenance, boolean>` with a per-tier rationale — ADR-015's move applied here. **Returns exactly what it returned for all seven tiers**; the existing table-driven test covers all seven and passes unchanged.

## Two doc-truth defects found while mapping the surfaces

Both on `packages/site/docs/guide/mcp-tools.md`:

1. `dependent-attribution-partial` documented as `today only for C#`.
2. **A whole fictional vocabulary.** The page documented `testAssociations[]` as `{ testFile, confidence: "high"|"medium"|"low", method }` with a *"Confidence Levels"* section (`high`/`medium`/`low`) and a *"Detection Methods"* section — none of which exists in code. The real field is `string[]`. It was also attributed to `search_code`, which never emitted the field at all (`grep testAssociations` over `search-code.ts` and `metadata-shaper.ts`: no hits). Replaced with the four real static mechanisms and an explicit "there are no per-association confidence levels".

Worth noting because that phantom shape is exactly the "stronger/weaker than import-verified" vocabulary #1018 assumed didn't exist — it existed only in docs, describing something never built.

## Tests

- `inferred-dependent-mechanisms.test.ts` (11 new): prose non-empty/distinct per mechanism; ids derived from the table; **`not.toContain('global using')` / `not.toContain('source text')` on Go's descriptor** — the regression pinned at the source; dedup of a repeated mechanism; empty-list degradation.
- `get-dependents.test.ts` (3 new): the Go note must contain `its language, Go` and its real mechanism, and must **not** contain `C#`, `global using` or `source text`; the C# note must still contain C#'s; both mechanisms together render `C# and Go`.
- `dependency-analyzer.test.ts` (parser): the pairing invariant **in both directions** — `confidence === 'inferred'` iff `inferredVia !== undefined`, and an import-verified dependent carries neither; the #1039 chi repro now also asserts `inferredVia`.

## Gates

`format:check` ✅ · `lint` **0 errors** (41 pre-existing warnings) ✅ · `typecheck` ✅ · `build` ✅ · `npm test` **5,790 tests / 240 files, all green** ✅ · `docs:build` ✅ (rendered page read) · `lien delta` **exit 0** ✅

`lien delta` reports **4 improved, 0 worsened, 0 new violations**:

```
packages/cli/src/mcp/handlers/get-dependents.ts
  · new       buildPartialRecoveryNote cyclomatic – → 1
  ✓ improved  buildAttributionCaveat   cognitive 8 → 7
packages/parser/src/dependency-analyzer.ts
  ✓ improved  enrichWithCSharpTypeReferenceDependents time 40m → 35m
  ✓ improved  enrichWithGoRootPackageDependents       time 39m → 35m
packages/review/src/dependency-graph.ts
  ✓ improved  isPreciseProvenance      cognitive 1 → 0
✓ 4 improved · 11 files · 236 ms
```

Relevant to the `review` gate (#1070), which is delta-aware and blocks on a *worsened* pre-existing violation even without a threshold crossing: nothing here worsens. The one crossing this PR did produce mid-development (`buildAttributionCaveat` halstead 38→65, threshold 60) was caught by the hook and fixed by extracting `buildPartialRecoveryNote`, not silenced with `--soft`.

## Harness evidence

No prompt, rule, or fixture files touched — nothing under `packages/review/src/plugins/agent/` or `packages/review/test/harness/` appears in the diff. The only `review` change is `dependency-graph.ts`'s `isPreciseProvenance` table.

```
 .changeset/brave-pandas-consolidate.md             |  40 ++++
 .../0016-dependency-attribution-honesty-axes.md    | 255 +++++++++++++++++++++
 docs/architecture/decisions/README.md              |   1 +
 packages/cli/src/mcp/attribution-caveat-reasons.ts |  39 +++-
 .../src/mcp/handlers/dependency-analyzer.test.ts   |   2 +
 .../cli/src/mcp/handlers/get-dependents.test.ts    |  67 ++++++
 packages/cli/src/mcp/handlers/get-dependents.ts    |  50 +++-
 packages/cli/src/mcp/tools.ts                      |  13 +-
 packages/parser/src/dependency-analyzer.test.ts    |  39 ++++
 packages/parser/src/dependency-analyzer.ts         |  66 ++++--
 packages/parser/src/index.ts                       |  15 ++
 .../src/inferred-dependent-mechanisms.test.ts      | 130 +++++++++++
 .../parser/src/inferred-dependent-mechanisms.ts    | 164 +++++++++++++
 packages/review/src/dependency-graph.ts            |  62 +++--
 packages/site/docs/guide/mcp-tools.md              |  53 +++--
 15 files changed, 908 insertions(+), 88 deletions(-)
```

## ADR

Yes — **ADR-016**, following ADR-015's precedent (a repeatedly-relitigated cross-cutting decision turned into a documented contract with a compiler-forced field, recording the rejected proposal alongside the accepted one). It documents the two-axis/three-scope model, the `require-only` proof, the routing rule for a new honesty signal, #1067's landing spot, and an explicit out-of-scope list: `lien annotate`'s four-tier test-coverage prose (the next site, needs a data-shape change), per-association test provenance (rejected consumer), the "stronger" direction (no consumer), #1015's floor prose (later), and renaming the overloaded `confidence` (no benefit).

## What I deliberately left out

- **Merging the three vocabularies** — argued above and in ADR-016.
- **Surfacing a 7-tier provenance in `get_dependents`' JSON** (#1018's part 2 as literally drafted). The tiers are review's symbol-level pipeline; exporting them to MCP would publish distinctions parser can't produce and make the consumer-facing field vaguer, which is the opposite of the goal.
- **`lien annotate`'s test-coverage tiers** — the same Axis-A decision at another surface, over four parallel arrays. Needs a `testAssociations` shape change; nothing is blocked on it. Named in the ADR as the next site.
- **A sixth `AttributionCaveatReason` for #1067** — it would be dead vocabulary until the tier producing it ships. #1067 adds its table entry together with its producer.

---

### Lien Review

> [!NOTE]
> **Low Risk**
>
> This PR fixes a real shipped bug where Go files recovered by the root-package export fallback were described with C#-specific prose. It introduces `DependentInfo.inferredVia` and a parser-owned `INFERRED_DEPENDENT_MECHANISMS` table, then drives all consumer-facing prose (caveat note, tool description, server instructions, docs) from that single source. The change is additive (`confidence` is untouched), the pairing invariant is enforced by a single constructor, and the new code is well-covered by tests. No breaking changes or silent failure modes were found.
>
> - Added `InferredDependentMechanism` vocabulary and `INFERRED_DEPENDENT_MECHANISMS` table in `@liendev/parser`
> - Added `DependentInfo.inferredVia`, always paired with `confidence: 'inferred'` via `inferredDependent()`
> - Updated `get_dependents` caveat note, tool description, and server instructions to derive mechanism prose from the table
> - Converted `isPreciseProvenance` to a `Record<EdgeProvenance, boolean>` to prevent silent defaults for new tiers

> [!WARNING]
>

<sup>Reviewed by [Lien Review](https://lien.dev) for commit 980c9db. Updates automatically on new commits.</sup>

---

Attested: degraded:budget_starved · 702K/820K tokens

### DIFF
--- a/.changeset/brave-pandas-consolidate.md
+++ b/.changeset/brave-pandas-consolidate.md
@@ -1,0 +1,40 @@
+---
+'@liendev/parser': patch
+'@liendev/lien': patch
+'@liendev/review': patch
+---
+
+fix(parser,cli,review): name the fallback behind an inferred dependent (#1018)
+
+`get_dependents`' `dependent-attribution-partial` caveat described C#'s
+type-reference fallback in every case, because `confidence: 'inferred'` was
+single-valued and the mechanism identity was discarded at the parser boundary.
+When #1039 added Go's root-package export lookup — same marker, same caveat
+reason — every recovered Go file was told *"its language, C#, lets real callers
+use its exports with no per-file import naming it at all"* and that its
+dependents came from *"matching a uniquely-declared type name against other
+files' source text"*. Both false; measured on a real `go-chi/chi` clone, 24 of
+24 recovered edges across `context.go`/`mux.go`/`chain.go`.
+
+`@liendev/parser` now owns `INFERRED_DEPENDENT_MECHANISMS`, a `Record`-guarded
+table of the non-import recovery fallbacks and their canonical prose, and
+`DependentInfo.inferredVia` names the mechanism per dependent. Every
+consumer-facing surface — the caveat note, the caveat-reason text, the server
+instructions, the tool description and the docs page — derives from the table
+instead of restating it, so a third fallback is a compile error until its prose
+exists and then correct everywhere at once.
+
+`DependentInfo.confidence` is unchanged and still marks exactly what it did;
+`inferredVia` is additive. `review`'s `isPreciseProvenance` returns exactly what
+it returned for all seven tiers, now via a `Record<EdgeProvenance, boolean>` so
+an eighth tier can't default silently.
+
+Also fixes two doc-truth defects on the MCP tools page found while mapping the
+surfaces: `dependent-attribution-partial` was documented as C#-only, and
+`testAssociations[]` was documented as `{ testFile, confidence, method }` with a
+"Confidence Levels" section — a shape and vocabulary that exist nowhere in the
+code (the real field is `string[]`), attributed to a tool that never emitted the
+field.
+
+ADR-016 records why the three vocabularies #1018 named were not merged into one,
+and the routing rule for where a new honesty signal belongs.
--- a/docs/architecture/decisions/0016-dependency-attribution-honesty-axes.md
+++ b/docs/architecture/decisions/0016-dependency-attribution-honesty-axes.md
@@ -104,12 +104,16 @@ natural home.
 
 `confidence: 'inferred'` was single-valued, so the mechanism identity was
 discarded at `parser`'s boundary and every prose surface had to name a
-mechanism from memory. When #930 shipped there was one (C#), so six surfaces
-hard-coded C#. #1039/#1064 then added Go's root-package export lookup —
-same marker, same caveat reason — and updated none of them, because nothing
-forced it: `Record<AttributionCaveatReason, string>` guards the set of
-*reasons*, and no reason was added. Only an existing reason's mechanism
-coverage widened.
+mechanism from memory. When #930 shipped there was one (C#), so all six
+surfaces described C#'s mechanism specifically — four of them naming C# by
+name (the runtime caveat note, the `AttributionCaveatReason` doc comment, the
+tool description, the docs page), the other two saying "text-matching
+fallback" (`ATTRIBUTION_CAVEAT_REASON_TEXT` and, by interpolation, the server
+instructions), which is equally wrong for Go. #1039/#1064 then added Go's
+root-package export lookup — same marker, same caveat reason — and updated
+none of them, because nothing forced it:
+`Record<AttributionCaveatReason, string>` guards the set of *reasons*, and no
+reason was added. Only an existing reason's mechanism coverage widened.
 
 Measured on a real `go-chi/chi` clone against `origin/main`, every recovered
 Go root file was told *"its language, C#, lets real callers use its exports
--- a/docs/architecture/decisions/0016-dependency-attribution-honesty-axes.md
+++ b/docs/architecture/decisions/0016-dependency-attribution-honesty-axes.md
@@ -82,11 +82,18 @@ shared bucketing predicate would have to make one of those two answers
 wrong.
 
 Independently: four of `EdgeProvenance`'s seven tiers (`import-only`,
-`require-only`, `symbol-name-match`, `oop-method-import`) are symbol-level
-distinctions `parser`'s file-level `findDependents` cannot produce. Hoisting
-the enum into `parser` would publish tiers nothing in `parser` emits — a
-vocabulary with dead members, which is worse than two honest ones. So
-`EdgeProvenance` stays `review`'s internal ranking detail.
+`require-only`, `symbol-name-match`, `oop-method-import`) are distinctions
+that only arise *in* a symbol-level analysis — each records something
+different about what could be established for the specific symbol, which is a
+question `parser`'s file-level `findDependents` never asks. Note this is not a
+claim that those four tiers all verify a symbol: `require-only` and
+`symbol-name-match` explicitly do not, which is why
+`SYMBOL_VERIFIED_BY_PROVENANCE` marks them `false`. The point is that a
+file-level pipeline has no way to *draw* the distinction at all, in either
+direction. Hoisting the enum into `parser` would therefore publish tiers
+nothing in `parser` emits — a vocabulary with dead members, which is worse
+than two honest ones. So `EdgeProvenance` stays `review`'s internal ranking
+detail.
 
 ### Why Axis B was already consolidated
 
--- a/docs/architecture/decisions/0016-dependency-attribution-honesty-axes.md
+++ b/docs/architecture/decisions/0016-dependency-attribution-honesty-axes.md
@@ -1,0 +1,255 @@
+# ADR-016: Dependency-Attribution Honesty Has Two Axes and Three Scopes — Route, Don't Merge
+
+**Status**: Accepted
+**Date**: 2026-08-05
+**Deciders**: Core Team
+**Related**: #1018 (proposal and the refusal recorded here), PR for #1018
+(implementation) · #930, #940, #951, #980, #984, #994, #1005, #1011, #1013,
+#1014, #1015, #1026, #1030, #1039, #1064, #1067, #1072, #1078
+
+## Context and Problem Statement
+
+#1018 observed that three vocabularies answer what looks like one question —
+*how much should a consumer trust this dependency answer?* — and proposed
+consolidating them into a single canonical type owned by `parser`, which
+`core`, `cli` and `review` would consume instead of maintaining parallel
+dialects:
+
+| vocabulary | home | shape |
+|---|---|---|
+| `confidence?: 'inferred'` | `packages/parser/src/dependency-analyzer.ts` | single-valued, per dependent |
+| `EdgeProvenance` | `packages/review/src/dependency-graph.ts` | 7 ranked tiers, per caller edge |
+| `AttributionCaveatReason` | `packages/cli/src/mcp/attribution-caveat-reasons.ts` | 5 reasons, per response |
+
+The pressure was real. Four consumers had been blocked or had bent around
+the gap: #1072 needed to express whole-store and whole-language properties
+and fell back to the `note` channel plus omission; #1067's declaration-index
+tier named a missing caveat reason as an explicit prerequisite, its concept
+currently living as prose in four separate signal-module doc comments; #1015
+wants to say a count is a floor distinctly from the existing caveat; and an
+evaluation of coverage-derived test associations found `testAssociations:
+string[]` carries no provenance at all.
+
+The decision needed was not "merge or don't" but *what is actually one
+decision here, and what is several*.
+
+## Decision Drivers
+
+- A consumer-facing honesty signal must get **more** specific under
+  consolidation, never vaguer. #1014 is the cost of a signal that fires
+  wrongly; a signal that fires vaguely is the same failure with extra steps.
+- Whatever replaces or extends an existing vocabulary must keep its
+  build-breaking forcing function (#984's `Record<>` pattern). Prose that can
+  drift silently is the defect class this repo bleeds.
+- `review` depends on `parser` only, never `core` (ADR-012). `parser` is the
+  one package all three consumers already import.
+- Preserve shipped behaviour: #1026's caveat reasons, #1030's
+  verified/inferred bucketing, #1072's note-and-omission.
+
+## Decision Outcome
+
+**The three vocabularies were NOT merged.** They sit on two different axes,
+and one of the three is already downstream of the other's canonical facts.
+What *was* consolidated is the one thing genuinely decided in more than one
+place and demonstrably wrong in production: **the identity of the recovery
+mechanism behind an inferred dependent.**
+
+### The model: two axes, three scopes
+
+**Axis A — how was this edge resolved?** A property of an individual edge
+that is present. `parser`'s `confidence` and `review`'s `EdgeProvenance` both
+live here.
+
+**Axis B — why can't this answer be trusted as a verified clear?** A property
+of an answer, usually about what is *absent*. `cli`'s
+`AttributionCaveatReason` lives here, and nothing else does.
+
+Orthogonal to both, a signal has one of three **scopes**: per-edge,
+per-answer (about a caller-supplied `filepath`), or per-store/per-language
+(a capability of the index or the language, independent of any query).
+
+### Why Axis A's two members must stay separate
+
+They are the same axis at **different granularities**, and each is correct
+for its own.
+
+`require-only` is the proof. A Ruby `require_relative` verifiably resolves
+the target *file*, so `parser`'s file-level `findDependents` correctly treats
+that dependent as import-verified and attaches no `confidence` marker. The
+same statement names no *symbol* at all, so `review`'s symbol-level graph
+correctly ranks it imprecise — the boundary #1030 argued from evidence. One
+shared bucketing predicate would have to make one of those two answers
+wrong.
+
+Independently: four of `EdgeProvenance`'s seven tiers (`import-only`,
+`require-only`, `symbol-name-match`, `oop-method-import`) are symbol-level
+distinctions `parser`'s file-level `findDependents` cannot produce. Hoisting
+the enum into `parser` would publish tiers nothing in `parser` emits — a
+vocabulary with dead members, which is worse than two honest ones. So
+`EdgeProvenance` stays `review`'s internal ranking detail.
+
+### Why Axis B was already consolidated
+
+`AttributionCaveatReason` is not an independent taxonomy. `buildAttributionCaveat`
+derives all five reasons from fields `parser` already owns on
+`FindDependentsResult`: `targetIndexed`, `symbolAttributionDegraded`,
+`symbolFoundInFile`, `typeSymbolAttributionIncomplete`,
+`dependentAttributionPartial`, `dependentAttributionIncomplete`. The
+structural facts are in `parser`; `attribution-caveat-reasons.ts` is the
+prose and mutual-exclusion layer over them. Moving model-facing prose into
+the AST package would gain nothing and cost the #984 forcing function its
+natural home.
+
+### What was actually broken, and fixed
+
+`confidence: 'inferred'` was single-valued, so the mechanism identity was
+discarded at `parser`'s boundary and every prose surface had to name a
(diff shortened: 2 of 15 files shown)


alfhen:
Review triage — 4 CodeRabbit findings, 2 fixed, 2 rejected with reasons. (Fixed in `45f09bb6`.)

**Fixed**

1. **`tools.ts` import order** — valid, and a straight violation of `CLAUDE.md`'s stated order (Node built-ins → external → internal). `@liendev/parser` now sits above the internal imports.
2. **`ADR-0016` → `ADR-016`** — valid. The file is `0016-*.md` but the record's own title and the index entry both read `ADR-016`. Corrected in `packages/parser/src/index.ts` and `packages/review/src/dependency-graph.ts`.

**Rejected**

3. **ADR date `2026-08-05` vs "on August 4, 2026"** — timezone artifact, not an error. The commit is `2026-08-05T00:16:32+02:00`; `git log` renders author-local time by default, so anyone browsing this repo's history sees `2026-08-05`. Changing the ADR to `2026-08-04` (the UTC date) would make it *disagree* with the commit date a reader correlates it against. Repo is authored from Copenhagen and ADR-015 follows the same convention.

4. **Use `collect.js` for the map/filter/dedupe chain in `buildPartialRecoveryNote`** — rejected on two grounds. `CLAUDE.md`'s rule is conditional: *"Use `collect.js` for readable data transformations (groupBy, countBy, chained map/filter/sort). **Prefer native `.map()`, `.filter()` for simple single operations**"* — this is a 2-step map/filter plus a `Set`, not the groupBy/countBy shape the rule targets. More importantly, the `.filter()` here is a **TypeScript type guard** narrowing `InferredDependentMechanism | undefined` → `InferredDependentMechanism`; `collect.js`'s `filter` does not carry that narrowing through, so adopting it would replace a compile-time-checked narrowing with an unchecked cast. That is a net loss on a path whose whole purpose is honesty about which mechanism ran.

**Note on check status:** CodeRabbit's check reads `pass` on the latest SHA but the reason is `Review rate limited` — i.e. it did not re-run. The 4 findings above come from its earlier run against `7b59497`, fetched from both `/comments` and `/reviews`. Lien Review ran to completion on both SHAs: `verdict: delivered`, all 5 passes `starved: false`, 0 findings, 0 complexity issues.

Gates re-run after the fixes: `format:check` ✅ · `lint` 0 errors ✅ · `typecheck` ✅ · `build` ✅ · `npm test` 5,790 green ✅ · `lien delta` exit 0 ✅

alfhen:
Triage of the `review` failure on `45f09bb6` — the doc-truth finding was **correct and is fixed** in `980c9dbe`.

🔴 **doc-truth (error)**: *"The ADR calls `require-only` a 'symbol-level distinction', but the code's `EdgeProvenance` doc comment and `SYMBOL_VERIFIED_BY_PROVENANCE` table explicitly classify `require-only` as 'file-level proof only'."*

**Accepted.** My ADR said four tiers "are symbol-level distinctions", which reads as *these verify the symbol* — and that flatly contradicts the table I added in the same PR, where `require-only` and `symbol-name-match` are `false` with the comment "file-level proof only". Doubly bad in a PR whose whole subject is prose that drifted from code, and doubly confusing because the same ADR uses `require-only` as its central proof that the two granularities differ.

The intended claim was about where a distinction can be **drawn**, not what it verifies. Reworded in both the ADR and the PR body, with the non-verifying tiers named explicitly so the two readings cannot be conflated again.

**Attestation disclosure** (per #1077 — the badge is not the attestation):

| pass | ran | stopReason | starved | deferred |
|---|---|---|---|---|
| `main` | yes | `completed` | no | 0 |
| `doc-truth` | yes | `completed` | no | 0 |
| `stale-duplicate-loop` | yes | `budget` | **yes** (21791/18458) | 0 |
| `incomplete-handling-loop` | yes | `completed` | no | **6** |

Run verdict was `degraded:budget_starved`. Two things worth stating plainly rather than leaving in the log:

- The finding that failed the gate came from `doc-truth`, which ran to **completion and was not starved** — so it is a healthy finding, not an artifact of truncation.
- `incomplete-handling-loop` **deferred 6 candidates**, including `summarizeInferredDependentMechanisms` and `inferredVia` — i.e. two of this PR's own new surfaces were not examined by that pass. `stale-duplicate-loop` was starved outright. Neither is a blocker (both are auxiliary passes and `main` completed clean), but it means coverage on this diff is not total, and I would rather say so than let a green badge imply otherwise.

Also carried forward from the previous SHA: CodeRabbit's check reads `pass` only because it was **rate limited** and never re-ran; its 4 findings from `7b59497` are triaged in the comment above (2 fixed, 2 rejected with reasons).

alfhen:
Pre-merge verification, head `980c9dbe`:

**Checks: 22/22 pass**, `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`. Rollup re-verified against the current head SHA (matches local HEAD exactly).

**Lien Review on the final SHA — 0 error findings, `conclusion=success`:**

| pass | ran | stopReason | starved | deferred |
|---|---|---|---|---|
| `main` | yes | `completed` | no | 0 |
| `doc-truth` | yes | `completed` | no | 0 |
| `stale-duplicate-loop` | yes | `completed` | no | 0 |
| `incomplete-handling-loop` | yes | `budget` | **yes** (95287/67498) | **6** |

The run verdict still reads `degraded:budget_starved`, and I am not going to let the green badge stand in for that. Being precise about what it does and does not mean here:

- The three passes that could produce an error-severity finding — `main`, `doc-truth`, `stale-duplicate-loop` — all ran to **completion, unstarved**. The doc-truth pass is the one that caught the real ADR contradiction on the previous SHA, so it is demonstrably working on this diff, and it is clean now.
- `incomplete-handling-loop` is the sole degraded pass, and it **deferred 6 candidates: `deduplicateResults`, `formatNoIndexNote`, `hasData`, `shapeResults`, `summarizeInferredDependentMechanisms`, `inferredVia`.** Two of those are this PR's own new surfaces. That pass therefore did **not** examine them, and I am recording that as a known, non-total coverage result rather than implying a clean sweep.

**Findings ledger across all four SHAs** (swept on both `/comments` and `/reviews`, paginated):

| finding | source | disposition |
|---|---|---|
| ADR `require-only` mislabelled "symbol-level" | Lien doc-truth, `45f09bb6` | **Fixed** in `980c9dbe` — the one inline comment still on the PR is against the old line and is stale; current run reports 0 errors |
| `tools.ts` import order | CodeRabbit | **Fixed** in `45f09bb6` |
| `ADR-0016` → `ADR-016` | CodeRabbit | **Fixed** in `45f09bb6` |
| ADR date UTC vs author-local | CodeRabbit | **Rejected** — timezone artifact; commit is `+02:00`, `git log` renders author-local |
| `collect.js` for the map/filter chain | CodeRabbit nitpick | **Rejected** — `CLAUDE.md` prefers native for simple ops, and `collect.js` would drop the TS type-guard narrowing for a cast |

No unanswered findings remain. Merging.

Issue #1071, opened 2026-08-03
core: dependentCount is 0 for 100% of files in 6 of 8 languages — search_code's structural ranking boost is a no-op


alfhen:
## Summary

`search_code`'s structural ranking boost is **provably the identity function in 6 of 8 measured languages**. `dependentCount` is `0` for **100% of files** — not "mostly zero", the **maximum value is 0** — across 1,962 files in C#, Kotlin, Swift, Java, Go and Rust.

The cause is not the orphan rates tracked elsewhere. `packages/core/src/vectordb/sqlite/dependent-counts.ts` is a **third, independent dependency resolver** that consults none of the working ones, and it has **exactly one commit in its history** — the one that created it.

## Measured

Real `computeDependentCounts` / `normalizeFileForCounts` from `origin/main`, run against indexed corpora:

| corpus | lang | files | `dependentCount == 0` | **max** | max boost multiplier |
|---|---|---|---|---|---|
| serilog | C# | 216 | **216 (100%)** | **0** | **1.000** |
| Exposed | Kotlin | 775 | **775 (100%)** | **0** | **1.000** |
| TCA | Swift | 796 | **796 (100%)** | **0** | **1.000** |
| javapoet | Java | 39 | **39 (100%)** | **0** | **1.000** |
| gin | Go | 99 | **99 (100%)** | **0** | **1.000** |
| anyhow | Rust | 37 | **37 (100%)** | **0** | **1.000** |
| flask | Python | 80 | 59 (73.8%) | 17 | 1.434 |
| lien | TypeScript | 450 | 229 (50.9%) | 55 | 1.604 |

`applyStructuralBoost(ratio, 0)` = `ratio · (1 + 0.15·ln(1))` = **`ratio`**. Exactly identity, every time.

**The median multiplier is 1.000 in all eight corpora — TypeScript included.** So for a median file in this repo the boost also does nothing.

## Mechanism — two independent bugs in ~40 lines

**1. Only relative specifiers are resolved.** `dependent-counts.ts:49`:

```ts
if (!specifier.startsWith('./') && !specifier.startsWith('../')) return specifier;
```

Everything else passes through unchanged and can then only match if a file path *literally equals* the specifier string. Actual indexed values:

```
javapoet  UtilTest.java         ["org.junit.Assert.assertEquals", "org.junit.Test"]
serilog   AllocationsBench.cs   ["Serilog.Core.Enrichers"]
[...]

PR #772, opened 2026-07-15
feat(cli,plugin): plan-time complexity nudge before edits

alfhen:
## Why (the wedge, step 2)

Lien already computes `complexityHeadroom` (functions at/near their complexity budget) and returns it from `get_files_context`, and `lien delta` flags crossings **after** code is written (the write-time hook, `delta-write.sh`). The missing moment is **before** the agent writes: when it reads or is about to edit a file, near-budget functions should be unmissable so the agent steers around them or extracts instead of piling on. A prior behavioral A/B in this repo proved a well-placed warning changes agent output (#672/#682 line) — this extends that to pre-write.

## Channel decision (read first: `docs/architecture/claude-code-hook-channels.md`)

The obvious design is a new `PreToolUse:Edit|Write` hook mirroring `delta-write.sh`. **Rejected** — per the repo's own hook-channels doc, `PreToolUse`'s only channel that delivers model-visible content is `hookSpecificOutput.updatedInput.prompt`, and that's specific to the subagent-launch tool (`augment-explore-task.sh` rewrites a launched subagent's own prompt). `Edit`/`Write`'s `tool_input` has no `prompt` field to rewrite. The other `PreToolUse` channel, `exit 2` + stderr, does reach the model — but by *blocking* the tool call, which would turn an advisory nudge into a hard stop and break every existing hook's fail-open contract.

**Fallback (shipped):** enrich the existing `annotate-read.sh` (`PostToolUse:Read`, `additionalContext` — the channel proven to reach the model). It already fires right before the mandatory `get_files_context` → `Edit` sequence in the normal agent workflow, so it lands at the same "before you write" moment through infrastructure already proven to work. TTL suppression (per file per session) is inherited for free — no new dedup logic needed. Fail-open is unaffected: I didn't touch hook error-handling, only the CLI output it wraps.

## What shipped

- **`get_files_context`**: new optional `complexityHeadroomWarning` string field, spread into the response *before* `complexityHeadroom` so it's the first thing the agent reads in the serialized JSON — an imperative line instead of just data. `complexityHeadroom` itself is unchanged (no schema break).
- **`computeComplexityHeadroom`** (in `get-files-context.ts`): parameter type widened from `SearchResult[]` to a minimal `{ metadata: ChunkMetadata }` shape, so the exact same computation is reusable by the CLI (no re-implementation, no drift risk between the MCP tool and the CLI).
[...]

PR #1088, opened 2026-08-04
fix(parser): extract call sites from module-level code, not just function bodies (#1087)

### LINKED ISSUE #1087: parser: call-site extraction shares the complexity gate, so module-level code contributes no references
## Summary

`packages/parser/src/ast/chunker.ts` gates call-site extraction behind
`shouldCalcComplexity` — the *same* flag that gates cognitive/Halstead
metrics:

```ts
const shouldCalcComplexity = symbolInfo?.type && COMPLEXITY_SYMBOL_TYPES.has(symbolInfo.type);
// COMPLEXITY_SYMBOL_TYPES = new Set(['function', 'method'])
...
const callSites = shouldCalcComplexity ? extractCallSites(node, language) : undefined;
```

Complexity metrics genuinely only mean something for a function or a method.
Call sites do not have that restriction, and sharing the flag makes any code
that isn't a function or method body contribute **no call-site evidence at
all**:

- `export const client = createClient(loadConfig())` — a top-level
  declaration that holds no function is not a `'function'`/`'method'` symbol,
  so `createChunk` emits nothing for it.
- Bare top-level statements — route registration, DI wiring, `app.use(...)`,
  config objects assembled from factory calls — reach `createChunkFromRange`,
  which never calls `extractCallSites` at all (it has no `SyntaxNode` to call
  it with).
- A Vitest/Jest file is *almost entirely* bare top-level statements, so a test
  calling the function it tests registers no call site.

## Measured

Emitted call sites vs. call expressions actually present in the tree, this
repo (`packages/`, 617 AST-parsed files):

| | call expressions | emitted | coverage |
|---|---|---|---|
| production source (276 files) | 11,236 | 10,608 | 94.4% |
| test files (341 files) | 52,611 | 1,510 | **2.9%** |
| **all** | **63,847** | **12,118** | **19.0%** |

Built as an identifier-join graph (references × declarations, same shape as
the evaluation that surfaced this), this repo is the weakest of 12 measured
corpora:

| | before |
|---|---|
| files with ≥1 reference to a locally-declared identifier | **53.6%** |
| joined edges/file | **8.02** |

For comparison the same measurement on foreign corpora: JavaPoet 100%, chi
94.9%, monolog 88%. The two corpora below us are express (12.1%) and zod
(26.9%) — both, like us, JS/TS codebases where the module level does the
composition.

## Why it matters beyond the number

Anyone evaluating ranking or graph quality *on this repo* is measuring the
corpus this gate depresses most, and will conclude a working change doesn't
work. That nearly happened this week.

## Scope

Widening is a behaviour change, not a flag flip, and needs care in two
places:

1. `createChunkFromRange` has no AST node — module-level call sites have to
   come from the file's `rootNode`, filtered by the ranges the other chunks
   already cover.
2. A class chunk's line range *contains* its method chunks' ranges, and
   nothing downstream dedupes call sites across a file's chunks
   (`dependency-analyzer.ts`'s `extractSymbolUsagesFromChunks`, `review`'s
   `buildCallerEdges`). Extracting for a container would report every
   method's calls twice, inflating `totalUsageCount` and caller-edge counts.

Complexity metrics must not move.

alfhen:
Closes #1087.

## The gate

`ast/chunker.ts` extracted call sites behind `shouldCalcComplexity` — the same
flag as cognitive/Halstead. Complexity genuinely only means something for a
function or method. Call sites do not have that restriction, and sharing the
flag left three kinds of code contributing no reference evidence at all:

| shape | why it produced nothing |
|---|---|
| `export const client = createClient(loadConfig())` | a top-level declaration holding no function isn't a `'function'`/`'method'` symbol, so `createChunk` skipped it |
| bare top-level statements — route/DI registration, `app.use(...)` | they reach `createChunkFromRange` from a raw line range, which has no `SyntaxNode` to extract from |
| almost all of any Vitest/Jest file | it is bare top-level statements end to end, so a test never registered as calling what it tests |

Measured on this repo, that silence covered **81%** of the call expressions
actually in the tree (12,118 of 63,847 emitted). Split by file kind it was
94.4% coverage in production source and **2.9% in test files**.

## What changed

Call sites are now extracted for every chunk whose line range no other chunk
overlaps, plus the module-level chunks that previously had no node to extract
from (`withModuleLevelCallSites`, which works from the file's `rootNode` minus
the ranges the top-level chunks already claim).

Containers stay excluded, and that is the load-bearing constraint: a class
chunk's range *contains* its method chunks', and neither
`dependency-analyzer.ts`'s `extractSymbolUsagesFromChunks` nor `review`'s
`buildCallerEdges` dedupes call sites across a file's chunks, so extracting
there would report every method's calls a second time and inflate
`totalUsageCount` and caller-edge counts.

The container predicate checks `getContainerBody() !== null` as well as
`shouldExtractChildren`, because a language may state the intent to descend and
then decline per node. Python's `decorated_definition` is in `containerTypes`
but returns no body for a decorated *function*. Getting this wrong showed up as
a **regression**, not a missing improvement: psf/requests lost 384 of 2580
references on the first pass, which is how it was caught.

## Consumer map — what moves and what does not

| consumer | reads `callSites`? | effect |
|---|---|---|
| precomputed `dependentCount` (`dependent-count-index.ts`, #1071) | **no** — imports/`importedSymbols` only | **none.** Verified: the `dependent_counts` table is identical before/after (314 rows, total 1701, `diff` empty) |
| `search_code` structural ranking (`applyStructuralBoost`) | no — consumes the count above | **none** |
| test associations (`test-associations.ts`) | **no** — imports + filename conventions | **none** |
| `get_dependents` symbol-level (`findSymbolUsages`) | yes — `fileImportsSymbolFromAny` fallback + `extractSymbolUsagesFromChunks` | more usages, and more symbol-level dependents. Still gated on the file importing from the target first |
| `review`'s caller graph (`buildCallerEdges`) | yes — 5 of the 7 provenance tiers | more call-site-backed edges; the `import-only`/`require-only` fallback tiers shrink correspondingly (a precision improvement — they exist for exactly the absence this fixes) |
| `get_files_context` | yes (only tool with `callSites` in its `metadata-shaper` allowlist) | more call sites returned |
| `search_code` / `list_functions` / `find_similar` | no — not in their allowlists | **none** |
| complexity metrics | n/a | **none** — see below |

## Complexity metrics did not move

`shouldCalcComplexity` still gates cognitive/Halstead exactly as before; only
call-site extraction widened. Verified row-for-row against two isolated indexes
of the same tree:

```
== before
chunks|10579
complexity_sum|10646|cognitive_sum|9220
halstead_volume_sum|1791020.91|halstead_effort_sum|39396968.0
boundary_digest|10579
callsite_entries|14262
== after
chunks|10579
complexity_sum|10646|cognitive_sum|9220
halstead_volume_sum|1791020.91|halstead_effort_sum|39396968.0
boundary_digest|10579
callsite_entries|66345

== row-level diff of (file, startLine, endLine, type, symbolName, complexity,
   cognitiveComplexity, halsteadVolume) across all 10,579 chunks
(diff lines:        0)
```

## Before/after — 12 corpora, 11 of them the ones the E2E suite names

Identifier-join graph (references × declarations), the same shape as the
evaluation that surfaced this. `conn%` is the share of files referencing an
identifier declared in the same corpus; `dupes` is call sites attributed to
more than one chunk of the same file.

```
corpus           files             refs         join%         conn%      edges/file  dupes
MediatR (C#)       154       2578->2578    25.3->25.3    56.5->56.5      4.23->4.23      0
SwiftyJSON          23       1815->1923    39.5->40.9    91.3->95.7    31.17->34.22      0
anyhow (Rust)       37        997->1006    51.6->51.2    75.7->75.7    13.89->13.92      0
chi (Go)            78       3508->3518    63.7->63.6    94.9->94.9    28.63->28.67      0
express (JS)       141       243->11282      14->15.4    12.1->84.4     0.24->12.33      0
javapoet (Java)     39       5402->5402        45->45      100->100    62.31->62.31      0
klaxon (Kotlin)    101       1982->2137    52.9->50.9    79.2->81.2    10.38->10.76      0
monolog (PHP)      216       5793->5799    43.9->43.9        88->88    11.78->11.78      0
requests (Py)       37       2580->2650    56.9->55.8    62.2->70.3       39.68->40      0
sinatra (Ruby)     147       2215->8347    56.3->45.8    42.9->82.3     8.48->26.02      0
zod (TS)           405      3390->34065      65.1->50    26.9->85.4     5.45->42.04      0
lien (self)        617     12129->63838    40.8->28.7    53.6->87.5     8.02->29.66      0
```

No corpus regressed on any measure. The class-based languages (C#, Java, PHP,
Go, Rust) barely move — they put everything in methods, which already worked.
The movers are exactly the corpora where the module level does the composition:
express, zod, sinatra, and us.

**This repo was the second-weakest of the twelve and is now mid-pack**, which
is the point of the issue: anyone evaluating ranking or graph quality on this
repo was measuring the corpus the gate depressed most.

### False-hub check (#928 family) — looked for specifically, not found

Per-reference join precision *dilutes* in the corpora that gain a lot (zod
65.1% → 50%, us 40.8% → 28.7%). That is the expected shape, not a hub: the
newly-added references skew toward member-property names, and the ones that
don't name a declared identifier (`expect`, `it`, `toBe`, `describe` — 20,000+
on this repo alone) cannot join anything, so they cost bytes and create no
edges.

The one measured hub candidate was `readFile`: `fs.readFile` resolves to the
property name, and `review/src/plugins/agent/agent-tools.ts` exports a
`readFile`. 462 new module-level references across 15 files. Real
`get_dependents` output, same query against both indexes:

```
BEFORE  dependentCount: 2   totalUsageCount: 1
AFTER   dependentCount: 2   totalUsageCount: 4
```

`dependentCount` unchanged. The symbol path still requires the caller file to
import from the target (`fileImportsSymbolFromAny`), so `fs.readFile`
everywhere else never becomes a caller. The three added usages are the real
ones, in the one test file that does import it.

Structurally: **0 duplicate attributions** across all 12 corpora
(65k+ entries), and a unit test pins it.

## Index cost, measured

Isolated `LIEN_HOME` per run, two runs each, 766 files:

| | before | after | delta |
|---|---|---|---|
| wall time | 3116 / 3030 ms | 3407 / 2870 ms | within noise (~+2% on means) |
| `chunks.db` | 26.52 / 26.26 MB | 29.29 / 29.51 MB | **+11.2%** |
| whole index dir | 26084 / 25828 KiB | 28792 / 29004 KiB | **+11.6%** (+2.9 MiB) |

4.65x the stored call sites for +11% bytes and no measurable time cost —
`callSites` is a `TEXT` column on `chunks` and is not FTS-indexed, so nothing
else grows.

## Dogfood — real MCP server, real index, before vs after

`lien serve` over stdio against two isolated indexes of this tree.

`get_dependents` for a symbol whose test callers were previously invisible:

```
BEFORE  dependentCount: 1  totalUsageCount: 0
        dependents: [ annotate-cmd.test.ts ]   (no usages at all)

AFTER   dependentCount: 3  totalUsageCount: 5
        packages/cli/test/integration/index-state-matrix.test.ts
          (module-level)  L1217  const result = await handleGetDependents({ filepath: 'math.ts' }, makeCtx(vectorDB));
          (module-level)  L1229  { filepath: 'does/not/exist.ts' },
          (module-level)  L1243  const result = await handleGetDependents({ filepath: 'math.ts' }, makeCtx(vectorDB));
        packages/cli/src/mcp/handlers/get-dependents.test.ts
          (module-level)  ...
```

`get_files_context` on `cli/src/mcp/schemas/dependents.schema.ts` — a file that
is one `z.object({...})` and nothing else, i.e. the issue's first example:

```
BEFORE  0 call sites
AFTER   28 call sites
```

## A third instance of the same root cause, found by the review gate

Lien Review flagged two findings on the first push, both in
`packages/review/src/dependent-context.ts`: review's own copy of the
`callerSymbol` fallback and of the snippet arithmetic, neither of which the
second commit had touched. Both correct. Fixed in the third commit by
extracting the mechanism into `parser`'s `findChunkLineIndex` and having both
call sites use it, rather than fixing the same decision twice.

The second one mattered more than its 🟡 suggested. Dogfooded through
`assembleDependentContext` on this repo's chunks, symbol `chunkByAST`:

```
BEFORE  **Dependent Usage Context** (no call-site data available):
        - 54 file(s) depend on `chunkByAST`

AFTER   **Dependent Usage Context** (top 3 of 54 dependents):
        // packages/parser/src/ast/chunker.test.ts:56 — in (module-level)
              const chunks = chunkByAST('test.ts', content);
              expect(chunks.length).toBeGreaterThan(0);
```

The review agent was getting **no code at all** for a symbol with 54
dependents, falling back to a filename list, because every caller is
module-level. `formatDependentContext` was also rendering `in (module-level)()`,
which claims module-level code is a function; the `()` is now dropped for it.

## Two knock-on fixes (second commit)

Both are output that only becomes common once module-level code carries call
sites, and both were found by reading the dogfood output rather than by tests:

- `callerSymbol` fell back to `'unknown'` for any chunk with no symbol name,
  which module-level chunks never have. Now `'(module-level)'` for a `'block'`
  chunk, matching the wording `review`'s `dependency-graph.ts` already uses.
- `snippet` was computed as `callLine - startLine`, exact only when a chunk's
  content starts on the line `startLine` names. `createChunkFromRange` trims a
  range's leading blank lines out of the content while `startLine` keeps naming
  the untrimmed start, so the arithmetic landed a line late and returned the
  neighbouring statement (`") as ToolResult;"` for a call to `readFile`). Now
  located by finding the nearby line that mentions the symbol. Deliberately
  corrects the lookup rather than narrowing the chunk's `startLine`/`endLine`
  to match its content — `isValidChunk` measures `minChunkSize` from exactly
  those fields, so shrinking them would silently drop small module-level chunks
  (the #772 failure mode). The `line` a usage reports was always correct.

## E2E floors — no floor and no tripwire moves, verified per corpus

`expectedMinDependencyEdges` is computed by `computeDependencyStats` →
`countDependents` → `findDependents`, which resolves *imports*;
`expectedMinTestAssociations` comes from `findTestAssociationsFromChunks`,
which reads no call sites. Rather than assert that from the code, I replicated
both computations (same 100-file sweep, same helpers) against both parser
builds on all 11 cloned corpora:

```
                     totalEdges   orphans   testAssociations
MediatR      before/after   393/393     39/39      194/194
SwiftyJSON   before/after       0/0     23/23          0/0   <- exact-zero tripwire
anyhow       before/after     32/32     23/23          6/6
chi          before/after     64/64     68/68        20/20
express      before/after 2239/2239     66/66    1605/1605
javapoet     before/after     18/18     36/36        16/16
klaxon       before/after       4/4     97/97          0/0   <- exact-zero tripwire
monolog      before/after   186/186     84/84        32/32
requests     before/after   350/350       3/3      136/136
sinatra      before/after     77/77     85/85        45/45
zod          before/after   545/545     73/73          5/5
```

Every number identical, including both `toBe(0)` tripwires
(`KNOWN_ZERO_EDGE_LANGUAGES` for Swift edges,
`KNOWN_ZERO_TESTASSOC_LANGUAGES` for Swift/Kotlin associations) — those assert
*exact* zero, so a call-site-driven leak into either would have failed CI
rather than merely shifted a floor. Nothing to update.

Worth noting for #1053's family: there is no E2E assertion on call-site volume
at all, so this metric could have collapsed to near-zero silently — as it in
fact was, at 19% coverage, while the suite stayed green. Filed as #1090 rather
than bolted onto this PR, with the measured per-corpus values to derive floors
from.

## Gates

| gate | result |
|---|---|
| `format:check` | pass |
| `lint` | 0 errors (41 pre-existing warnings) |
| `typecheck` | 0 errors |
| `build` | pass |
| `npm test` | 5794 tests pass (1845 parser / 1766 / 1684 / 431 / 41 / 27) |
| `lien delta` | exit 0 — new functions at cognitive 6, 5, 2, 2 |

New coverage:

- 7 tests in `parser/src/ast/chunker.test.ts` — the gate being changed had
  **zero** `callSites` assertions before this. Covers the
  declaration-without-function shape, bare top-level statements, a test file
  with no top-level declaration, the no-double-attribution invariant,
  complexity staying untouched, and both Python decorated-function and
  decorated-class cases.
- 7 tests in `parser/src/chunk-line-lookup.test.ts` on the shared lookup,
  including both misalignment directions and the search-radius boundary.
- 1 snippet-alignment regression test in
  `cli/src/mcp/handlers/dependency-analyzer.test.ts`.
- 3 in `review/test/dependent-context.test.ts` — module-level caller label,
  trimmed-offset centring, and the overshoot-past-content case that used to
  return `null` and drop the snippet.

---

### Lien Review

> [!NOTE]
> **Low Risk**
>
> This PR widens call-site extraction from function/method bodies to module-level code in the AST chunker, adds a shared `findChunkLineIndex` lookup to correct snippet alignment for trimmed module-level chunks, and unifies caller labeling as `(module-level)` across parser and review. The core logic is well-tested (new tests cover the declaration-without-function shape, bare top-level statements, test files, no-double-attribution, complexity preservation, Python decorated functions/classes, Ruby module bare calls, and snippet alignment), and the two issues flagged in the prior review run are both addressed in the current diff: `dependency-graph.ts` now uses `callerSymbolFor`, and `chunker.ts` handles transparent-container gaps. No output-breaking or caller-breaking bugs remain.
>
> - Call sites extracted for all non-container AST chunks plus module-level uncovered ranges
> - Shared `findChunkLineIndex` lookup used by both `get_dependents` snippets and review dependent context
> - `callerSymbolFor` exported and used consistently to label module-level callers as `(module-level)`

<sup>Reviewed by [Lien Review](https://lien.dev) for commit 16bea4f. Updates automatically on new commits.</sup>

---

### DIFF
--- a/.changeset/lucky-pandas-listen.md
+++ b/.changeset/lucky-pandas-listen.md
@@ -1,0 +1,24 @@
+---
+'@liendev/parser': patch
+---
+
+Extract call sites from module-level code, not just function and method bodies (#1087)
+
+`ast/chunker.ts` gated call-site extraction behind `shouldCalcComplexity` — the
+same flag as cognitive/Halstead metrics — so anything that wasn't a function or
+method body contributed no call-site evidence at all: `export const client =
+createClient(loadConfig())`, route/DI registration, config objects built from
+factory calls, and (because a test file is almost entirely bare top-level
+statements) every test that calls the function it tests.
+
+Call sites are now extracted for every chunk whose line range no other chunk
+overlaps, plus the module-level "uncovered" chunks that previously had no AST
+node to extract from. Containers stay excluded, because a class chunk's range
+contains its methods' and nothing downstream dedupes across a file's chunks.
+
+Measured across 12 real corpora, the share of files referencing an identifier
+declared in the same corpus: this repo 53.6% → 87.5%, zod 26.9% → 85.4%,
+express 12.1% → 84.4%, sinatra 42.9% → 82.3%, requests 62.2% → 70.3%. Zero
+duplicate attributions anywhere. Complexity metrics, chunk boundaries and
+precomputed `dependentCount` are byte-identical before and after; index size
+grows ~11% and index time is unchanged.
--- a/.changeset/lucky-pandas-listen.md
+++ b/.changeset/lucky-pandas-listen.md
@@ -22,3 +22,10 @@ express 12.1% → 84.4%, sinatra 42.9% → 82.3%, requests 62.2% → 70.3%. Zero
 duplicate attributions anywhere. Complexity metrics, chunk boundaries and
 precomputed `dependentCount` are byte-identical before and after; index size
 grows ~11% and index time is unchanged.
+
+Two knock-on fixes for output that only becomes common once module-level code
+carries call sites: `get_dependents` usages from module-level code now report
+`callerSymbol: "(module-level)"` rather than `"unknown"` (matching the wording
+`review` already uses), and their `snippet` is located by finding the line that
+mentions the symbol instead of computing it, which was landing a line late for
+any chunk whose leading blank lines were trimmed out of its content.
--- a/.changeset/wild-otters-agree.md
+++ b/.changeset/wild-otters-agree.md
@@ -1,0 +1,22 @@
+---
+'@liendev/parser': patch
+'@liendev/review': patch
+---
+
+Share one chunk-line lookup between `get_dependents` usage snippets and review's dependent context (#1087)
+
+Both computed a snippet's position as `callSiteLine - chunk.metadata.startLine`,
+which is only exact when a chunk's content starts on the line `startLine` names.
+A module-level chunk's does not — `createChunkFromRange` trims its range's
+leading blank lines out of the content while `startLine` keeps naming the
+untrimmed start — so the subtraction overshoots: `get_dependents` returned a
+neighbouring statement as the snippet, and review's `extractSnippetWindow`
+either centred its window a line late or, once the overshoot exceeded the
+content, returned `null` and dropped the snippet from the review prompt
+entirely.
+
+Both now go through `findChunkLineIndex`, which locates the line by finding the
+nearby one that mentions the symbol. `review`'s module-level callers are also
+labelled `(module-level)` rather than `unknown`, matching what parser already
+reports, and the `()` suffix is dropped for them since module-level code is not
+a function.
--- a/packages/cli/src/mcp/handlers/dependency-analyzer.test.ts
+++ b/packages/cli/src/mcp/handlers/dependency-analyzer.test.ts
@@ -15,6 +15,7 @@ function createChunk(
     callSites: Array<{ symbol: string; line: number }>;
     symbolName: string;
     symbolType: 'function' | 'method' | 'class' | 'interface';
+    type: 'function' | 'class' | 'block';
     content: string;
     startLine: number;
     endLine: number;
@@ -359,6 +360,39 @@ describe('findDependents', () => {
         callerSymbol: 'handleRequest',
         line: 3,
         snippet: 'doWork(data);',
+      });
+    });
+
+    it('finds the right snippet line for a module-level chunk whose content was trimmed (#1087)', async () => {
+      // A module-level chunk's range starts on the blank line after the
+      // preceding chunk, and `createChunkFromRange` trims that line out of
+      // `content` while `startLine` keeps naming it -- so `line - startLine`
+      // lands one line late. The reported `line` is the true file line either
+      // way; the snippet has to be found, not computed.
+      const chunk = createChunk('src/wiring.ts', {
+        imports: ['src/target.ts'],
+        importedSymbols: { 'src/target': ['doWork'] },
+        callSites: [{ symbol: 'doWork', line: 11 }],
+        symbolName: undefined,
+        type: 'block',
+        startLine: 10,
+        endLine: 13,
+      });
+      // Range was lines 10-13, but line 10 was blank and got trimmed away, so
+      // content[0] is really file line 11.
+      chunk.content = 'export const wired = doWork(config);\n\nexport const other = 1;';
+
+      mockDB.scanAll.mockResolvedValue([
+        createChunk('src/target.ts', { exports: ['doWork'] }),
+        chunk,
+      ]);
+
+      const result = await findDependents(mockDB as any, 'src/target.ts', mockLog, 'doWork');
+
+      expect(result.dependents[0].usages![0]).toEqual({
+        callerSymbol: '(module-level)',
+        line: 11,
+        snippet: 'export const wired = doWork(config);',
       });
     });
   });
--- a/packages/parser/src/ast/chunker.test.ts
+++ b/packages/parser/src/ast/chunker.test.ts
@@ -1411,6 +1411,34 @@ class Service:
       expect(entries).toHaveLength(new Set(entries).size);
       expect(symbolsOf(chunks).has('helper')).toBe(true);
     });
+
+    it('records a bare call directly inside a transparent container (Ruby module)', () => {
+      // Review finding on #1087/#1088: `emitsChildChunks` deliberately voids a
+      // transparent container's OWN call sites (extractCallSites recurses
+      // unscoped, so extracting on the module node would double-count its real
+      // children), but that container's full range used to sit in
+      // `coveredRanges` too -- so a bare statement directly inside the module,
+      // wrapped in no further declaration of its own, was silently dropped:
+      // neither the module's chunk nor any child chunk claimed it.
+      const content = `module Foo
+  configure()
+
+  class Bar
+    def method1
+      helper_call()
+    end
+  end
+end
+`;
+      const chunks = chunkByAST('foo.rb', content);
+      const symbols = symbolsOf(chunks);
+
+      expect(symbols.has('configure')).toBe(true);
+      expect(symbols.has('helper_call')).toBe(true);
+
+      const entries = allCallSites(chunks);
+      expect(entries).toHaveLength(new Set(entries).size);
+    });
   });
 
   describe('error handling', () => {
--- a/packages/parser/src/ast/chunker.test.ts
+++ b/packages/parser/src/ast/chunker.test.ts
@@ -1267,6 +1267,152 @@ test('does something', () => {
     });
   });
 
+  describe('module-level call sites (#1087)', () => {
+    /** Every (symbol, line) pair the whole file's chunks report. */
+    const allCallSites = (chunks: ReturnType<typeof chunkByAST>) =>
+      chunks.flatMap(c => (c.metadata.callSites ?? []).map(cs => `${cs.symbol}:${cs.line}`));
+
+    const symbolsOf = (chunks: ReturnType<typeof chunkByAST>) =>
+      new Set(chunks.flatMap(c => (c.metadata.callSites ?? []).map(cs => cs.symbol)));
+
+    it('records call sites in a top-level declaration that contains no function', () => {
+      // The shape this fix exists for: `export const x = createThing(...)` is
+      // not a 'function'/'method' symbol, so sharing the complexity gate meant
+      // it contributed no call-site evidence at all.
+      const content = `import { createClient } from './client.js';
+import { loadConfig } from './config.js';
+
+export const client = createClient(loadConfig());
+`;
+      const chunks = chunkByAST('wiring.ts', content);
+
+      expect(symbolsOf(chunks)).toEqual(new Set(['createClient', 'loadConfig']));
+    });
+
+    it('records call sites in bare top-level statements', () => {
+      const content = `import { app } from './app.js';
+import { authRoutes, userRoutes } from './routes.js';
+
+app.use(authRoutes());
+app.use(userRoutes());
+`;
+      const chunks = chunkByAST('server.ts', content);
+      const symbols = symbolsOf(chunks);
+
+      expect(symbols.has('use')).toBe(true);
+      expect(symbols.has('authRoutes')).toBe(true);
+      expect(symbols.has('userRoutes')).toBe(true);
+    });
+
+    it('records call sites in a test file with no top-level declaration at all', () => {
+      const content = `import { describe, it, expect } from 'vitest';
+import { chunkByAST } from './chunker.js';
+
+describe('chunker', () => {
+  it('chunks', () => {
+    expect(chunkByAST('a.ts', 'x')).toHaveLength(1);
+  });
+});
+`;
+      const chunks = chunkByAST('chunker.test.ts', content);
+
+      expect(symbolsOf(chunks).has('chunkByAST')).toBe(true);
+    });
+
+    it('never attributes the same call site to two chunks of one file', () => {
+      // Nothing downstream dedupes call sites across a file's chunks, so an
+      // overlapping attribution would inflate get_dependents' totalUsageCount
+      // and review's caller-edge counts.
+      const content = `import { helper, decorate } from './util.js';
+
+export const table = { a: helper(1), b: helper(2) };
+
+export class Service {
+  private dep = helper(3);
+
+  run() {
+    return helper(4);
+  }
+
+  also() {
+    return decorate(helper(5));
+  }
+}
+
+register(table);
+`;
+      const chunks = chunkByAST('service.ts', content);
+      const entries = allCallSites(chunks);
+
+      expect(entries).toHaveLength(new Set(entries).size);
+    });
+
+    it('leaves complexity metrics untouched', () => {
+      const content = `import { helper } from './util.js';
+
+export const wired = helper(configure());
+
+export function branchy(n: number): number {
+  if (n > 1) {
+    for (let i = 0; i < n; i++) {
+      if (i % 2 === 0) return i;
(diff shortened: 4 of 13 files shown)


alfhen:
Both review findings triaged — **accepted and fixed** in 9efc2bf1.

They were the same defect this PR had already fixed once in `dependency-analyzer.ts` while leaving `review`s parallel copy alone, so rather than fix it twice I extracted the mechanism: `packages/parser/src/chunk-line-lookup.ts`s `findChunkLineIndex`, now the single implementation behind both `get_dependents` usage snippets and reviews dependent context.

**One correction to the second findings reasoning.** It states `relativeLine` is *negative* for a module-level caller with leading blanks. It **overshoots**, not undershoots: the content loses its leading lines while `startLine` keeps naming the untrimmed start, so `callSiteLine - startLine` is too *large*. The consequence described is right either way, and worse than it looked — once the overshoot exceeds the content, `extractSnippetWindow` returns `null` and the snippet is dropped from the review prompt silently, rather than merely being off by a line.

Also fixed while there: `formatDependentContext` was rendering `in (module-level)()`, which claims module-level code is a function. The `()` suffix is now dropped for that caller.

Dogfooded through `assembleDependentContext` on this repos own chunks, symbol `chunkByAST` over `packages/parser/src/ast`:

```
BEFORE  **Dependent Usage Context** (no call-site data available):
        - 54 file(s) depend on `chunkByAST`

AFTER   **Dependent Usage Context** (top 3 of 54 dependents):
        ```
        // packages/parser/src/ast/chunker.test.ts:56 — in (module-level)
              `.trim();

              const chunks = chunkByAST('test.ts', content);

              expect(chunks.length).toBeGreaterThan(0);
        ```
```

The review agent was previously getting **no code at all** for a symbol with 54 dependents, because every one of its callers is module-level — the degraded filename-only fallback. That is a second, independent instance of the same root cause this PR fixes, found only because the widening made it visible.

New coverage: 7 unit tests on `findChunkLineIndex`, plus 3 in `review/test/dependent-context.test.ts` (module-level caller label, trimmed-offset centring, and the overshoot-past-content case that used to return `null`).

alfhen:
## Merge-readiness state

**21 of 22 checks pass**, including the full 11-corpus E2E matrix, `build-and-test`, and `lien delta --base`. The one red check is **`review`, and it is an operational failure, not a finding.**

From the job log (`gh run view 30957905674 --log`):

```
##[warning][agent] chat request failed after retries (API error (402):
  "This request requires more credits, or fewer max_tokens. You requested up to
   24576 tokens, but can only afford 4827 ... limit_source: openrouter_credits")
##[warning][agent] Review never ran (stopReason=error, no verdict/summary after 0 completed turn(s))

"verdict":"failed:provider_never_ran"   "spentTokens":0   "neverRan":true
```

The OpenRouter key is out of credits, so **every** provider request 402d and zero tokens were spent. I re-ran the job once; it failed identically. The check`s own error text is explicit that this is not a clean review — and it will fail this way on every PR in the repo until credits are topped up, so waiting does not resolve it.

**What was reviewed.** The previous commit `9efc2bf1` — which contains every line of code in this PR — was reviewed successfully and came back **0 code findings**:

> ✅ **Good** — No complexity issues found.
> **Low Risk** ... The changes are well-isolated: complexity metrics remain gated as before, containers are excluded to avoid double attribution, and the new logic is covered by targeted unit tests including the Python decorated-function regression case. No behavioral bugs or breaking changes were identified.

The only delta since that clean review is `f2425afa`, which changes **one sentence of prose inside a comment** and no code.

**Findings triage is complete.** Both endpoints checked with `--paginate`: `/pulls/1088/comments` → 0, `/pulls/1088/reviews` → 1 (the two 🟡 findings on `a1a13832`, both accepted and fixed in `9efc2bf1`, answered above). No untriaged findings remain.

**Not merging on my own authority**, since the repo rule is to wait for CI rather than merge past a red check. Unblocking is a one-step owner action: add OpenRouter credits and re-run the `review` job.

One thing worth noting for its own sake: the `9efc2bf1` run reported `degraded:provider_partial` with the doc-truth pass at `stopReason: error, neverRan: true` — so doc-claim verification never actually ran on this PR. I self-audited the factual claims in my own comments as a result and found one wrong (a comment said 94% of call expressions went unextracted; the measured figure is 81%, 12,118 of 63,847). That is what `f2425afa` fixes. Both provider failures are the same root cause.

alfhen:
Updated: merged current `main` (`206127df`, the #1018 vocabulary work from #1086) into this branch at `c9548d3e`, so CI now tests the real combined state rather than a stale base. No conflicts, and no semantic clash either — #1018 also touched `parser/src/dependency-analyzer.ts` and `review/src/dependency-graph.ts`, and all six gates pass on the merged tree:

| gate | result |
|---|---|
| `format:check` | pass |
| `lint` | 0 errors |
| `typecheck` | 0 errors |
| `build` | pass |
| `npm test` | **5808** pass (1857 parser / 1766 / 1686 / 431 / 41 / 27) |
| `lien delta` | exit 0 |

Re-verified the load-bearing invariant on the merged code: **0 duplicate call-site attributions** across 660 files / 65,698 entries, and call-site coverage still ~100% (64,142 of 64,162).

`review` still fails identically — `"verdict":"failed:provider_never_ran"`, `code:402`, `"spentTokens":0`. Unchanged cause, unchanged remedy: add OpenRouter credits, then re-run that one job. Everything else is green and findings triage is complete (0 inline comments; the single review is on `a1a13832`, both findings fixed in `9efc2bf1`).

PR #1089, opened 2026-08-04
fix(cli): refuse to index home directory or filesystem root without explicit override (#1025)

### LINKED ISSUE #1025: lien index has no guard against indexing a home directory — swept Library/Keychains, Claude caches and npm logs into a 10.5 GB store
## Summary

`lien index` run from a home directory produced an index containing macOS keychain databases, Claude CLI caches and npm debug logs. There is no project-boundary check and no fallback exclusion set when the target has no VCS root or `.gitignore` to filter against.

Found on a developer machine where it had grown to **10.5 GB** and was the largest single contributor to a disk-full condition.

## What was actually indexed

Index `~/.lien/indices/alfhenderson-785a27b5`, `structural.db` = **10,543,894,528 bytes** from only **462 files** — so individual binary blobs were being chunked wholesale.

```
top-level dirs swept:
   219  Library
   186  .npm
    56  .claude
     1  .claude.json
```

13 paths matched sensitive-name patterns, including:

```
Library/Keychains/login.keychain-db
Library/Containers/com.1password.browser-support/Data/Library/Keychains/login.keychain-db
Library/Containers/com.apple.siri.*/Data/Library/Keychains/login.keychain-db
```

plus `Library/Caches/claude-cli-nodejs/…` (agent conversation transcripts) and `.npm/_logs/*-debug-0.log`.

**Scope note, stated fairly:** macOS keychain files are encrypted at rest, so this is not plaintext credentials in a database. The problems are that (a) these files have no business being indexed at all, (b) chunking multi-GB binaries produces a store two orders of magnitude larger than the file count implies, and (c) a user has no reason to expect `lien index` to reach into `Library/`.

## Why the existing protections don't fire

Lien's exclusion strategy leans on `.gitignore` (`gitignore.ts`) and ecosystem presets (`ecosystem-presets.ts`) — both of which assume a **project**. A home directory has no `.gitignore` and matches no ecosystem preset, so effectively nothing is excluded. `ALWAYS_IGNORE_PATTERNS` evidently doesn't cover `Library/`, `.npm/`, `.claude/` or `*.keychain-db`.

## Suggested fixes, in order of value

1. **Refuse to index a home directory or filesystem root** unless explicitly forced. `lien index` at `$HOME`, `/`, or `/Users` is essentially never intentional. A clear error with an override flag costs nothing and prevents the whole class.
2. **Extend `ALWAYS_IGNORE_PATTERNS`** with OS/user-cache paths that are never source: `Library/`, `AppData/`, `.npm/`, `.cache/`, `.claude/`, `*.keychain-db`, `*.keychain`, `.ssh/`, `.aws/`, `.gnupg/`. These should apply regardless of `.gitignore`, since the point is that no `.gitignore` exists.
3. **Skip or cap oversized binaries.** 462 files producing 10.5 GB means something is chunking blobs it should have rejected. A size ceiling and a binary sniff would have prevented the disk impact independently of the path filtering.
4. Consider surfacing index size in `lien status` and having `lien gc` flag anomalously large or stale indices — this one sat at 10.5 GB for three days without any signal.

(1) and (3) are independent safety nets and both are worth having: (1) stops the intent error, (3) bounds the damage when some other path filter is missed.

## Provenance

Found while diagnosing a 100%-full disk during the #994 follow-up work. The index has been deleted (regenerable cache); reclaiming it took `~/.lien/indices` from 11 GB to 1.0 GB.

Related: the broader disk issue is agent worktrees — 151 of them holding 115 GB, since each runs `npm ci` for its own ~800 MB `node_modules` and nothing prunes them. Filing separately.

alfhen:
## Summary

Closes #1025. `lien index` run from `$HOME` had no guard and swept macOS Keychain databases, `.npm` debug logs, and Claude Code caches into a 10.5 GB index on a maintainer's machine. This PR implements 3 of the 4 fixes the issue suggested, and skips the 4th with reasoning below.

**1. Hard refusal (the important one).** `lien index` now refuses when `process.cwd()` resolves to the home directory or a filesystem root (`/`, `C:\`, a Windows user-profile root), unless `--allow-unsafe-root` is passed. Implemented in `packages/cli/src/cli/unsafe-root.ts`. The home check delegates to a symlink-safe (`realpath`-compared) `isHomeDirectory` in `@liendev/parser` — needed because e.g. macOS's `/tmp` → `/private/tmp` symlink can make a naive string comparison silently miss a true match (a false negative in a safety check is worse than a false positive; caught this via manual reproduction, see Verification below).

**2. Extended ignore patterns — scoped, not universal.** `HOME_ROOT_ONLY_IGNORE_PATTERNS` (`Library/`, `AppData/`, `.npm/`, `.cache/`, `.claude/`, `.ssh/`, `.aws/`, `.gnupg/`, `*.keychain`/`*.keychain-db`) is layered on top of the existing `ALWAYS_IGNORE_PATTERNS` **only when the indexed root IS the home directory itself** (`packages/parser/src/gitignore.ts`'s `getEffectiveAlwaysIgnorePatterns`/`getEffectiveNeverIndexPatterns`). I scoped by "root IS home" rather than matching these names anywhere in a path, because `Library/` is a legitimate source directory in some ecosystems (Arduino, Unity, some Java layouts) — a universal exclusion would blind those projects. This also means the patterns only ever activate through the deliberate `--allow-unsafe-root` override (or another entry point that happens to resolve to home), never for an ordinary project.

**3. Oversized-file cap.** Indexing now skips (rather than chunks) any single file over 5 MB (`MAX_INDEXABLE_FILE_SIZE_BYTES` in `packages/core/src/constants.ts`), wired into all three read-then-chunk call sites (`indexer/index.ts`'s full-index path, `indexer/incremental.ts`'s single-file and batch paths). This is an independent backstop against the same disk-blowup class regardless of path filtering — it protects an ordinary project against an accidentally-committed multi-GB blob exactly as much as it protects an explicitly-overridden home-root scan against a keychain database. Skipped files are recorded with `chunkCount: 0` (like an empty file) so they aren't silently re-attempted every run.

**4. `lien status` now reports index size** (`Index size:` in text, `indexSizeBytes` in `--format json`) — the 10.5 GB index in the issue sat unnoticed for three days; this is the signal that would have surfaced it.

**Skipped:** the issue's further suggestion that `lien gc` flag anomalously large/stale indices. That's a separate feature (detection logic, thresholds, a new command surface) rather than a natural extension of this fix, and out of scope to keep this PR focused — left as a natural follow-up now that `lien status` exposes the raw number.

## Refusal message

```
Refusing to index your home directory (/private/tmp/lien-1025-fakehome-1).

Indexing this path would sweep OS caches, credential stores (SSH keys, cloud CLI configs, keychains), and every unrelated project underneath it into one index — this is almost never intentional, and it has previously produced a multi-gigabyte index built entirely from OS and agent cache files (see https://github.com/getlien/lien/issues/1025).

If this is really what you want, rerun with --allow-unsafe-root.
```
Exit code 1.

## Verification (manual, against isolated fixtures — never the real home directory or a shared index)

- **Refusal reproduced**: built `dist/index.js`, ran with `HOME`/`cwd` pointed at a constructed fake-home directory under `/tmp` → refused with the message above, exit 1. (This is also what caught the symlink bug: the first attempt silently did **not** refuse, because `/tmp` resolves to `/private/tmp` and the naive comparison missed it — fixed by comparing realpaths.)
- **Override works**: same fixture with `--allow-unsafe-root` → proceeded, indexed only the one legitimate file (`myproject-a/src/index.ts`); `Library/Caches/.../transcript.md`, `.npm/_logs/npm-debug.md`, and `.ssh/notes.md` were all correctly excluded (confirmed via the manifest).
- **No over-refusal**: a project directory living directly under the fake `$HOME` (`myproject-b`, containing its own `Library/Sketchbook/sketch.h`) indexed normally with no refusal — both `src/index.ts` and `Library/Sketchbook/sketch.h` ended up in the manifest, proving a real project's own `Library/` directory is unaffected.
- **`lien status` dogfooded** against both the fixture and this actual repo (read-only, on the existing overlay index — no `--force`, no real-home touch): shows `Index size: 40.6 KB` / `20.9 MB` respectively, and `indexSizeBytes` in JSON output.
- **Full dogfood in this repo**: ran `lien index` (incremental, no `--force`) against this actual worktree — completed in 4.7s, no false refusal, 505 files indexed.

## Test plan

- [x] `npm run format:check` — clean
- [x] `npm run lint` — 0 errors (42 pre-existing warnings, unrelated)
- [x] `npm run typecheck` — clean
- [x] `npm run build` — clean
- [x] `npm test` (parser-native, parser, core, review, cli, action) — 1697+ tests pass across all packages
- [x] `lien delta` — exit 0, no new threshold crossings (several "worsened but still under threshold" entries from the added branching, all informational)
- [x] New/updated tests: `unsafe-root.test.ts` (new), `index-cmd.test.ts`, `gitignore.test.ts`, `scanner.test.ts`, `watcher/index.test.ts`, `incremental.test.ts`, `index.test.ts` (core indexer), `dir-size.test.ts`, `status.test.ts`
- [x] Changeset added (`@liendev/parser`, `@liendev/core`, `@liendev/lien`, patch)

---

### Lien Review

✅ **Improved!** Complexity reduced by 53. 5 pre-existing issues remain in touched files.
🔗 **Impact**: 1 high-risk file(s) with 14 total dependents
| 🔀 test paths | 1 | +0 |
| 🧠 mental load | 1 | -17 |
| ⏱️ time to understand | 3 | -36 |

> [!NOTE]
> **Low Risk**
>
> This PR adds well-scoped safety guards for `lien index`: refusing to run from the home directory or filesystem root unless `--allow-unsafe-root` is passed, applying root-anchored OS/credential ignore patterns only when the indexed root IS the home directory, capping single files at 5 MB with `chunkCount: 0` recording across full/incremental/batch/overlay paths, and reporting index size in `lien status`. The implementation matches the stated intent and the added tests cover the key boundaries. No code paths were found that produce wrong output, break callers, or silently swallow errors.
>
> - CLI refuses home/root indexing without --allow-unsafe-root, with symlink-safe comparison via @liendev/parser
> - Home-root-only ignore patterns are root-anchored and do not affect nested project directories
> - 5 MB oversized-file cap applied consistently across full, incremental, batch, and overlay indexing paths
> - lien status reports on-disk index size in text and JSON output

> [!WARNING]
>

<sup>Reviewed by [Lien Review](https://lien.dev) for commit d812a06. Updates automatically on new commits.</sup>

---

## Summary by CodeRabbit

* **New Features**
  * `lien index` now protects home-directory and filesystem-root scans by default, with an explicit override available.
  * Files larger than 5 MB are skipped safely while indexing continues.
  * Status output now reports index storage usage in text and JSON formats.
* **Bug Fixes**
  * Home-directory scans now exclude sensitive credentials, caches, and system folders.
  * Improved handling of empty, malformed, and oversized files prevents indexing interruptions.
* **Documentation**
  * Added release information for updated parser, core, and CLI packages.

---

Attested: degraded:budget_starved · 655K/722K tokens

### DIFF
--- a/.changeset/silly-mangos-jog.md
+++ b/.changeset/silly-mangos-jog.md
@@ -1,0 +1,13 @@
+---
+"@liendev/parser": patch
+"@liendev/core": patch
+"@liendev/lien": patch
+---
+
+`lien index` now refuses to index your home directory or a filesystem root (`/`, `C:\`, a Windows user-profile root) unless explicitly overridden with `--allow-unsafe-root`. This closes the incident behind #1025: running `lien index` from `$HOME` swept macOS Keychain databases, `.npm` debug logs, and Claude Code agent caches into a 10.5 GB index with no warning. The refusal names the exact path and the override flag; a genuine reason to index an unusual root is always one flag away.
+
+As defense in depth, an extra set of OS/credential exclusions (`Library/`, `AppData/`, `.npm/`, `.cache/`, `.claude/`, `.ssh/`, `.aws/`, `.gnupg/`, `*.keychain`/`*.keychain-db`) now applies whenever the indexed root IS the home directory itself — scoped so an ordinary project is never affected, even one with its own legitimate `Library/` directory (Arduino, Unity, some Java layouts) or one that simply lives directly under `$HOME` (`~/myproject`).
+
+Indexing also now skips any single file over 5 MB instead of chunking it whole — a backstop against the same disk-blowup class independent of path filtering, for a legitimately huge binary in an otherwise ordinary project just as much as for an overridden home-root scan.
+
+`lien status` now reports the index's on-disk size (`Index size:` in text output, `indexSizeBytes` in `--format json`), so an anomalously large index is visible instead of sitting unnoticed.
--- a/packages/cli/src/cli/index-cmd.test.ts
+++ b/packages/cli/src/cli/index-cmd.test.ts
@@ -1,4 +1,5 @@
 import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest';
+import os from 'os';
 
 // Mock @liendev/core
 const mockIndexCodebase = vi.fn();
@@ -152,5 +153,53 @@ describe('indexCommand', () => {
         onProgress: expect.any(Function),
       }),
     );
+  });
+
+  describe('unsafe-root refusal (#1025)', () => {
+    afterEach(() => {
+      vi.restoreAllMocks();
+    });
+
+    it('refuses to index the home directory and never calls indexCodebase', async () => {
+      vi.spyOn(os, 'homedir').mockReturnValue('/Users/fakehome');
+      vi.spyOn(process, 'cwd').mockReturnValue('/Users/fakehome');
+
+      await indexCommand({});
+
+      expect(processExitSpy).toHaveBeenCalledWith(1);
+      expect(mockIndexCodebase).not.toHaveBeenCalled();
+      expect(consoleErrorSpy).toHaveBeenCalledWith(expect.stringContaining('your home directory'));
+      expect(consoleErrorSpy).toHaveBeenCalledWith(expect.stringContaining('--allow-unsafe-root'));
+    });
+
+    it('refuses to index a filesystem root and never calls indexCodebase', async () => {
+      vi.spyOn(process, 'cwd').mockReturnValue('/');
+
+      await indexCommand({});
+
+      expect(processExitSpy).toHaveBeenCalledWith(1);
+      expect(mockIndexCodebase).not.toHaveBeenCalled();
+      expect(consoleErrorSpy).toHaveBeenCalledWith(expect.stringContaining('filesystem root'));
+    });
+
+    it('proceeds when --allow-unsafe-root overrides a home-directory refusal', async () => {
+      vi.spyOn(os, 'homedir').mockReturnValue('/Users/fakehome');
+      vi.spyOn(process, 'cwd').mockReturnValue('/Users/fakehome');
+
+      await indexCommand({ allowUnsafeRoot: true });
+
+      expect(processExitSpy).not.toHaveBeenCalled();
+      expect(mockIndexCodebase).toHaveBeenCalled();
+    });
+
+    it('does not refuse an ordinary project directory that lives under $HOME (no over-refusal)', async () => {
+      vi.spyOn(os, 'homedir').mockReturnValue('/Users/fakehome');
+      vi.spyOn(process, 'cwd').mockReturnValue('/Users/fakehome/myproject');
+
+      await indexCommand({});
+
+      expect(processExitSpy).not.toHaveBeenCalled();
+      expect(mockIndexCodebase).toHaveBeenCalled();
+    });
   });
 });
--- a/packages/cli/src/cli/index-cmd.ts
+++ b/packages/cli/src/cli/index-cmd.ts
@@ -6,6 +6,7 @@ import type { IndexingProgress } from '@liendev/core';
 import { showCompactBanner } from '../utils/banner.js';
 import { getIndexingMessage, getModelLoadingMessage } from '../utils/loading-messages.js';
 import { formatDuration } from './utils.js';
+import { checkRootSafety, formatUnsafeRootMessage } from './unsafe-root.js';
 
 /**
  * Clears the existing index and manifest (for --force flag).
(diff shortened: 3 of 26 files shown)


alfhen:
Closing and reopening to force GitHub Actions to pick up the latest push (b77c853d) — that push's `pull_request: synchronize` webhook never reached Actions (0 check-runs / 0 workflow_runs for that SHA after 20+ minutes, while CodeRabbit's separate integration ran fine).

alfhen:
Triaged both CodeRabbit findings from the last review round:

**Fixed** — the symlink-to-filesystem-root bypass in `checkRootSafety` (`unsafe-root.ts`). `path.resolve()` doesn't resolve symlink targets, so a `cwd` that's a symlink to `/` would slip past the `resolved === fsRoot` check, while the home-directory half of the same guard (`isHomeDirectory`) was already symlink-safe via `realpathSync`. Exported parser's `toComparablePath` and pointed the filesystem-root check at it too. Added a regression test that creates a real symlink to `/` and confirms it's refused — verified it fails without the fix.

**Dismissed, not fixed** — "don't fingerprint skipped oversized files." `computeContentHash` already internally `stat`s the file and takes a bounded 16KB head+tail fingerprint for anything over 1MB (`LARGE_FILE_THRESHOLD`), which every 5MB+ "oversized" file always is — via positional `fs.read` at explicit offsets, not a sequential read. So no oversized file is ever fully read into memory regardless of whether `computeContentHash` runs before or after the explicit `isOversizedForIndexing` check; the actual safety guarantee (`#1025`'s "no gigabyte reads") already holds. This matches what `incremental.ts`'s own doc comments already state ("never the whole file"). Restructuring to skip hashing entirely, as suggested, would need to replace it with some other change-detection signal for oversized files and wasn't clearly justified as a net improvement.

Also confirmed the remaining Lien Review inline comments on this PR (the incremental.ts ordering bugs, index.ts manifest recording, overlay-index.ts cap, and the HOME_ROOT_ONLY_IGNORE_PATTERNS `**/` over-broad match) are stale — all already fixed in earlier commits on this branch, verified directly against the current source.

Issue #1057, opened 2026-08-01
get_dependents' type-symbol caveat wrongly reassures that dependentCount is reliable for Java/Kotlin/Swift (contradicts the file-level caveat for the same file)


alfhen:
## Summary

`get_dependents`'s type-symbol `attributionCaveat` (reason
`type-symbol-attribution-incomplete`, added by #1026) reassures the caller
that "`dependentCount`/`dependents` ... remain reliable" for the file-level
part of the answer. For Java/Kotlin/Swift — languages in
`KNOWN_ZERO_EDGE_LANGUAGES` (the already-documented #1005 gap, where
file-level dependency resolution is itself broken, not just symbol
attribution) — that reassurance is false: the identical file's plain
file-level query (no `symbol` param) returns a *different* caveat,
`dependent-attribution-incomplete`, whose text says the opposite ("don't
treat this as a verified clear").

## Repro

`get_dependents(filepath: ".../TypeSpec.java", symbol: "TypeSpec")` →
`dependentCount: 0`, `attributionCaveat.reason:
"type-symbol-attribution-incomplete"` — text implies the 0 file-level count
can be trusted, only the type-symbol usage count can't.

`get_dependents(filepath: ".../TypeSpec.java")` (no symbol) → returns a
*different* caveat, `dependent-attribution-incomplete`, whose own text
warns the caller not to treat the 0 as verified — the real, already-known
#1005 gap (Java dotted-FQN imports never resolve).

Both are describing the same underlying file/count; only one caveat reason
can win.

## Root cause

`buildAttributionCaveat` (`packages/cli/src/mcp/handlers/get-dependents.ts`)
checks `typeSymbolAttributionIncomplete` before
`dependentAttributionIncomplete` and returns on the first match. For a
Java/Kotlin/Swift type-symbol query, `typeSymbolAttributionIncomplete` is
true (correctly), so the function returns immediately — the file-level
`dependentAttributionIncomplete` condition (also true, for the same file) is
never reached, and its warning is silently dropped from the type-symbol
response even though it's just as applicable.

## Why this matters

A caller who queries a Java/Kotlin/Swift type symbol gets a caveat that
tells them the file-level dependent count is fine to trust — when it is
exactly as unreliable as the file-level-only query (which gets a caveat
saying the opposite) reveals. The five `AttributionCaveatReason` values are
documented as mutually exclusive (#980), which is true by construction
(`buildAttributionCaveat`'s first-match return), but that construction is
[...]

PR #1099, opened 2026-08-06
fix(cli,parser): recover real import-only evidence for type-symbol get_dependents queries (#1015, #1057)

### LINKED ISSUE #1015: get_dependents symbol-level usages are silently 0 for class/struct/interface names across languages (methods and functions are exact)

alfhen:
## Summary

Fixes #1015 (fix direction 2) and #1057.

`get_dependents({filepath, symbol})` returns `totalUsageCount: 0` for any
class/struct/interface/enum symbol, because usage attribution is
call-site-driven and nothing "calls" a type by its own name. Fix direction
1 (the `type-symbol-attribution-incomplete` caveat) already shipped in
#1026 — verified against current `main` before starting, and it is
comprehensive (it even has a text-based fallback, `isTypeDeclarationLine`,
for bare data structs with no dedicated chunk). This PR is fix direction 2:
real, non-fabricated evidence instead of just a caveat.

It also folds in #1057, filed against the exact same caveat: the note
asserted "dependentCount/dependents ... remain reliable" unconditionally,
which is false for C#/Java/Kotlin/Swift (the #1005 same-unit,
import-invisible access shape) — the identical file's file-level query
gets a *different*, contradictory caveat for those languages. #1098 (open,
unmerged, touches the same function) fixes a sibling gap — symbol-scoped
queries never getting `dependentAttributionIncomplete` at all — but by
design keeps `type-symbol-attribution-incomplete` as the sole caveat, so it
does not touch this wording problem. This PR only changes the note's text,
gated on `hasDependentAttributionBlindSpot`; it does not touch
`checkDependentAttributionIncomplete`'s gating, so it should not conflict
with #1098 beyond a routine rebase.

## What was added

A new optional response field, **`importedBy: string[]`** — present
whenever the `type-symbol-attribution-incomplete` caveat fires (even as
`[]`, so its presence always means "checked"). It lists files (always a
subset of `dependents`) that `@liendev/parser`'s call-site-level dependency
graph (`buildDependencyGraph`, lifted into parser by #1095 earlier the same
night specifically for this) can verify import the symbol via a
non-call-site edge — a constructor call, a type hint, an
`extends`/`implements` clause.

**No line number is attached.** An earlier version of this fix (analyzed,
then adversarially reviewed, before I wrote any code) proposed filling
`dependents[].usages[]`'s existing `{callerSymbol, line, snippet}` shape
from the graph's `import-only` tier. The review found that fatal: the
graph's `import-only`/C#-fallback edges fill `callSiteLine` with a
representative chunk's line, not a real call site, and testing it against
6 real cases found it wrong 6/6, 3 of them literally line 0. `importedBy`
is a deliberately separate, file-level-only field with no line field at
all — nothing to fabricate.

## Provenance tiers: included vs excluded

`EdgeProvenance` has 7 tiers; `isPreciseProvenance` marks 3 as precise:
`same-file`, `import-verified`, `import-only`. `importedBy` uses a
narrower set — `isImportOnlyEvidenceTier` in `get-dependents.ts`:

- **Included: `import-only`, `import-verified`.** Both require the
  caller's import specifier to verifiably resolve to the target file (via
  parser's guarded `importMatchesTarget`) AND the target file's own
  `exportIndex` entry to name the exact symbol. `import-verified` also has
  a real call site (so in practice it rarely adds anything beyond what
  `totalUsageCount` already found — included anyway since it's precise and
  exercises the same code path safely).
- **Excluded: `same-file`.** `findDependents` never counts an intra-file
  caller as a dependent; surfacing it here would contradict that.
- **Excluded: `require-only`.** Despite being `false` under
  `isPreciseProvenance` already, named explicitly in the filter (not just
  relying on that function) — it isn't language-gated to Ruby despite being
  Ruby-motivated, and the earlier analysis measured it fabricating 68 edges
  for a TypeScript interface (`BugCallerInfo`) actually used in exactly one
  file.
- **Excluded: `symbol-name-match`.** Also `false` under
  `isPreciseProvenance`, also named explicitly — measured resolving to the
  *wrong* file 4 of 5 times when multiple languages declare a same-named
  class.
- **Excluded (implicitly, already `false`): `oop-method-import`,
  `namespace-inferred`.**

## The subset-property invariant, enforced not just believed

`packages/parser/src/graph/dependency-graph.test.ts` now has 4 new tests
proving: every file the graph surfaces via a safe tier is already present
in `findDependents`'s own `dependents` (both mechanisms verify the same
import specifier via the same `importMatchesTarget` primitive, but the
graph additionally requires an exact `exportIndex` symbol match — a
strictly narrower, never wider, condition). Two more tests prove
`require-only` and `symbol-name-match` are excluded even when they are the
*only* thing the graph found for a seed (not coincidentally absent because
something stronger masked them — `getCallers` returns the first non-empty
tier, so those two tiers had to be tested in isolation; combining them with
`import-only`/`import-verified` in one fixture is architecturally
impossible, since a real call-site edge or `symbol-name-match` edge for a
key suppresses `import-only`'s fallback entirely for that same key — this
took two failed fixture attempts to discover and is now documented inline).

## The canonicalization gotcha, verified and tested

`buildDependencyGraph`'s `getCallers` keys on the RAW `chunk.metadata.file`
string; `get-dependents.ts` now calls `getCanonicalPath(filepath,
workspaceRoot)` before passing the target file into the graph. A dedicated
test (`get-dependents.test.ts`) passes a backslash-separated relative path
(the realistic case — the MCP schema rejects absolute paths and `..`
outright, so an absolute-path test would never reach this code) and
asserts `getCallers` receives the canonicalized, forward-slash form.

## Where the code lives

- `packages/cli/src/mcp/handlers/dependency-analyzer.ts`: `graphCache`, a
  sibling to the existing `scanCache`, keyed by the same `indexVersion`
  (the graph is a pure function of the same chunks). `getOrBuildDependencyGraph`
  builds it lazily — most `findDependents` calls never need it.
  `clearDependencyCache()` now clears both.
- `packages/cli/src/mcp/handlers/get-dependents.ts`: `isImportOnlyEvidenceTier`,
  `computeImportOnlyEvidence`, `buildTypeSymbolCaveatNote` — kept as small,
  standalone functions so `handleGetDependents`/`buildAttributionCaveat`
  stay lean (`lien delta` shows both only moved +1 cognitive complexity).
- `packages/cli/src/mcp/attribution-caveat-reasons.ts` +
  `packages/site/docs/guide/mcp-tools.md`: updated the single-source
  caveat prose and the public docs page to describe `importedBy` and the
  conditional reliability wording, so the static tool-description surfaces
  stay in sync with the dynamic per-response note.

## Dogfood evidence (verbatim, real MCP server over stdio, real corpora)

**This repo's own source**, freshly indexed, queried via a real `lien
serve` process over stdio (not the mocked unit tests):

`get_dependents({filepath: "packages/parser/src/graph/dependency-graph.ts", symbol: "CallerEdge"})`:
```json
{
  "dependentCount": 1,
  "totalUsageCount": 0,
  "importedBy": ["packages/parser/src/index.ts"],
  "attributionCaveat": {
    "reason": "type-symbol-attribution-incomplete",
    "note": "\"CallerEdge\" is a class/struct/interface/enum declaration ... totalUsageCount/usages below are a partial, best-effort floor ... not a verified total. 1 file(s) among the dependents below verifiably import \"CallerEdge\" per the dependency graph even though no literal call site names it (see `importedBy`) — real usage there is likely, just not call-site-attributable. dependentCount/dependents (which files import \"CallerEdge\") remain reliable. Verify with grep before concluding \"CallerEdge\" is unused or safe to rename."
  }
}
```
**Before this fix** (verbatim from the source as read prior to any edit —
the exact same query would have returned no `importedBy` field at all, and
the note read: `"...not a verified total; dependentCount/dependents (which
files import \"CallerEdge\") remain reliable. Verify with grep..."` — a
confident-looking 0 with zero evidence attached).

`get_dependents({filepath: "packages/cli/src/mcp/handlers/get-dependents.ts", symbol: "AttributionCaveat"})`
— the negative control, a genuinely orphaned local interface:
```json
{
  "dependentCount": 0,
  "totalUsageCount": 0,
  "importedBy": [],
  "attributionCaveat": { "reason": "type-symbol-attribution-incomplete",
    "note": "... The dependency graph found no non-call-site import evidence either (checked — `importedBy` is empty), which is a genuine absence of signal, not a confirmed 0 usages. ..." }
}
```

**Real JavaPoet clone** (Java — a #1005 blind-spot language), freshly
cloned and indexed:

`get_dependents({filepath: "src/main/java/com/squareup/javapoet/ClassName.java", symbol: "ClassName"})`:
```json
{
  "dependentCount": 1,
  "totalUsageCount": 0,
  "importedBy": ["src/test/java/ClassNameNoPackageTest.java"],
  "attributionCaveat": { "reason": "type-symbol-attribution-incomplete",
    "note": "... 1 file(s) among the dependents below verifiably import \"ClassName\" ... dependentCount/dependents (which files import \"ClassName\") aren't a verified clear either — this file's language has an import-invisible same-unit access shape (e.g. C#'s enclosing-namespace access, Java/Kotlin's same-package access, or Swift's whole-module access) the import graph can't see, so a real caller could exist with no import naming \"ClassName\" at all. ..." }
}
```
`TypeSpec` on the same corpus: `dependentCount: 0`, `importedBy: []`, same
blind-spot hedge — the #1057 fix firing correctly on real, unmodified Java
source (JavaPoet is essentially one package, so most real usages are
same-package with no import at all).

**Real Monolog clone** (PHP), freshly cloned and indexed:

`get_dependents({filepath: "src/Monolog/Logger.php", symbol: "Logger"})`:
```json
{
  "dependentCount": 15,
  "totalUsageCount": 0,
  "importedBy": [
    "src/Monolog/Handler/AbstractHandler.php", "src/Monolog/Handler/DeduplicationHandler.php",
    "src/Monolog/Handler/FilterHandler.php", "src/Monolog/Handler/FingersCrossed/ChannelLevelActivationStrategy.php",
    "src/Monolog/Handler/FingersCrossed/ErrorLevelActivationStrategy.php", "src/Monolog/Handler/FingersCrossedHandler.php",
    "src/Monolog/Handler/NullHandler.php", "src/Monolog/Handler/PushoverHandler.php", "src/Monolog/Handler/TestHandler.php",
    "src/Monolog/Processor/GitProcessor.php", "src/Monolog/Processor/IntrospectionProcessor.php",
    "src/Monolog/Processor/MercurialProcessor.php", "src/Monolog/Test/MonologTestCase.php",
    "tests/Monolog/Handler/PHPConsoleHandlerTest.php", "tests/Monolog/Handler/SymfonyMailerHandlerTest.php"
  ],
  "attributionCaveat": { "reason": "type-symbol-attribution-incomplete", "note": "... 15 file(s) ... dependentCount/dependents (which files import \"Logger\") remain reliable. ..." }
}
```
This is the exact shape #1015 reported for real: 15 of Monolog's own
dependents, all previously `totalUsageCount: 0` with a bare-trust note, now
have real, verifiable (not fabricated, no line numbers) evidence attached.

## Gate chain (all local, all green)

- `npm run format:check` — clean
- `npm run lint` — 0 errors (42 pre-existing warnings, none in touched files)
- `npm run typecheck` — clean
- `npm run build` — clean
- `npm run test -w @liendev/parser` — 1910/1910 (45 new/changed in `dependency-graph.test.ts`)
- `npm run test -w @liendev/lien` — 1715/1715 (81 in CLI's `dependency-analyzer.test.ts`, 55 in `get-dependents.test.ts`)
- `npm test` (full) — parser 1910, cli 1715, review 1725, action 41, all green
- `npm run docs:build` — clean
- `node packages/cli/dist/index.js delta --base origin/main` — **exit 0**, no new crossing. Six functions moved (3 new: `isImportOnlyEvidenceTier` cognitive 2, `computeImportOnlyEvidence` cognitive 3, `buildTypeSymbolCaveatNote` cognitive 2, `getOrBuildDependencyGraph` cognitive 3; 3 existing worsened by ≤1: `buildAttributionCaveat` 7→8, `buildDependentsResponse` 4→5, `handleGetDependents` 3→4) — all advisory, nowhere near the 15 threshold.

## `index-state-matrix.test.ts`

No new row needed — checked and reasoned through explicitly. This change
adds no new `createVectorDB` call site and no new MCP tool handler; it
enriches the *existing* `get_dependents` entry point's "ok" (real data)
response with an additional evidence field for one query shape. It doesn't
change how S0/S1/S2/S3 states are classified or handled. The completeness
guard (which fires on a new `toolHandlers` entry or `createVectorDB` call
site) is unaffected, and all 49 tests in that file still pass unchanged.

## Note on the GitHub Actions outage

Per instructions, I have not waited for or attempted to merge through CI —
this PR is opened ready (non-draft) with a fully green local gate chain and
is queued for whoever picks it up once Actions recovers.

## Summary by CodeRabbit

* **New Features**
  * Type-symbol dependency results now include verified importing files through `importedBy`, even when no direct call site is available.
  * Dependency analysis results are cached and refreshed automatically when indexed content changes.
* **Bug Fixes**
  * Improved attribution accuracy by excluding unverified import evidence and avoiding fabricated line numbers.
  * Clarified reliability limitations for languages with import-invisible access patterns.
* **Documentation**
  * Updated `get_dependents` guidance to explain `importedBy`, usage counts, and attribution limitations.

---

### Lien Review

> [!NOTE]
> **Low Risk**
>
> This PR adds a new optional `importedBy` field to `get_dependents` responses for type-symbol queries, backed by a lazy, cached call-site-level dependency graph. The implementation is careful: it intersects graph candidates against `analysis.dependents` to enforce the subset invariant, gates `importedBy` on the final decided caveat reason (not the raw flag), excludes unsafe provenance tiers (`same-file`, `require-only`, `symbol-name-match`), canonicalizes filepaths before graph lookup, and updates docs/caveat prose consistently. No wrong-output paths, breaking changes, or unvalidated input flows were found.
>
> - Added `importedBy?: string[]` to `get_dependents` responses when `type-symbol-attribution-incomplete` fires, populated from safe `import-only`/`import-verified` graph edges.
> - Added lazy `graphCache` and `getOrBuildDependencyGraph` in `dependency-analyzer.ts`, invalidated alongside `scanCache`.
> - Updated caveat prose to conditionally hedge `dependentCount` reliability for C#/Java/Kotlin/Swift blind-spot languages (#1057).

<sup>Reviewed by [Lien Review](https://lien.dev) for commit c9af9bf. Updates automatically on new commits.</sup>

---


alfhen:
GitHub Actions incident (stuck ARC runner pods) resolved at 02:04 UTC. Closing/reopening to force a fresh trigger for the commit that never got picked up during the outage.

alfhen:
## Follow-up: addressed CodeRabbit's 2 actionable findings + 2 of 3 nitpicks

**Finding 1 (subset guarantee argued, not enforced):** `computeImportOnlyEvidence` now takes `analysis.dependents` and intersects the graph's candidate callers against it before returning, rather than trusting the two mechanisms to keep agreeing. Added a test (`get-dependents.test.ts`) that mocks the graph reporting a caller `findDependents` did *not* return, and asserts it's dropped from `importedBy` — plus a second assertion checking the subset property directly on the parsed response.

**Finding 2 (gated on the wrong thing):** Extracted `decideAttributionCaveatReason` from `buildAttributionCaveat` so the priority-order decision has one source of truth. `handleGetDependents` now calls it first and only computes/attaches `importedBy` when the reason it decides is actually `'type-symbol-attribution-incomplete'` — not whenever the raw `analysis.typeSymbolAttributionIncomplete` flag is true. Added a test simulating the #927-manifest-vs-#928-chunk-scan disagreement (the concrete way an earlier-priority `unresolved-target` can still win while that flag is set) and asserting `importedBy` stays absent and the graph is never built in that case.

**Nitpicks addressed:**
- Moved `isImportOnlyEvidenceTier` out of the CLI handler and into `packages/parser/src/graph/dependency-graph.ts` (exported, next to `isPreciseProvenance`), so the parser's own subset-property tests import the real predicate instead of a test-local mirror that couldn't detect it drifting.
- Added a test for `getOrBuildDependencyGraph`'s uncached path (`indexVersion` undefined), which the existing versioned-cache tests didn't exercise.

**Nitpick not addressed:** none left outstanding — all three were applied.

Full local gate chain (format/lint/typecheck/build/parser tests 1910/1910/cli tests 1718/1718/full `npm test`/`docs:build`/`lien delta --base origin/main`) re-run clean after each round of fixes.

PR #1100, opened 2026-08-07
feat(parser): resolve Java/Kotlin same-package dependents (#1005 Mechanism 3, Phase 1)

alfhen:
## Summary

Implements Phase 1 of #1005's Mechanism 3: a new resolver
(`packages/parser/src/jvm-same-package-signals.ts`) that recovers real
`get_dependents` edges for Java/Kotlin files the import graph reports as
orphaned purely because same-package visibility (JLS §6.5.5.1 / Kotlin's
identical rule) needs no `import` statement at all.

Six gates, modeled on `csharp-type-reference-signals.ts`'s build/resolve
split:
- **G1' (package-local uniqueness)** — replaces the originally-planned
  corpus-wide uniqueness gate, which 3-way adversarial review measured to be
  simultaneously inert (0% effect on javapoet/klaxon) and permeable (passes
  exactly in the fabrication case, since a competing declaration in an
  unindexed jar is invisible to a corpus-wide count by construction).
- **G2** — exact package-string equality (no C#-style enclosing-namespace
  walk; Java/Kotlin have no such rule).
- **G3** — undeterminable package ⇒ non-participation.
- **G4** — candidate doesn't declare its own (nested-inclusive) same-named
  type.
- **G5** — class/interface declarations only.
- **G6 (new)** — import-shadowing exclusion: a candidate whose own
  single-type/single-static import binds the simple name to something else
  is excluded. Import-on-demand and Kotlin aliases are handled per JLS
  §6.4.1/§7.5.1/§7.5.3 (verified against real fabrication cases below).
- **G7** — source-set direction (non-test ↛ test), kept as free insurance;
  measured to catch zero edges G6 doesn't already catch.

Wired into `findDependents`'s enrichment chain (`enrichWithJvmSamePackageDependents`),
the batch `dependent-count-index.ts` (fresh-per-call index, not a
module-level cache — `jvm-source-root.ts`'s cache pattern would go stale
under `lien serve`), and `inferred-dependent-mechanisms.ts` (`confidence:
'inferred'`, `inferredVia: 'jvm-same-package'`). Phase 2 (`NO_DIRECTORY_NAMESPACE_LANGS`
in `graph/dependency-graph.ts`, feeding the call-graph consumer) and Phase 3
(Kotlin `typealias`/annotations, top-level `fun`/`val`) are explicitly out
of scope here.

## Real (not spike) measurements

All numbers below are from **this PR's actual parser/chunker/resolver**
(`performChunkOnlyIndex` + `findDependents`/`computeDependentCountsFromChunks`),
run against the same real clones the plan review used — not the filesystem
spike.

| Corpus | Edges before → after | Orphans before → after |
|---|---|---|
| JavaPoet | 18 → 152 | 40/43 (93.0%) → 23/43 (53.5%) |
| Klaxon | 4 → 228 | 100/103 (97.1%) → 69/103 (67.0%) |
| Retrofit | 645 → 974 | 337/392 (86.0%) → 223/392 (56.9%) |
| OkHttp | 467 → 1458 | 612/682 (89.7%) → 416/682 (61.0%) |
| SwiftyJSON | 0 → 0 | 27/27 (100%) → 27/27 (100%) — unaffected, as required (Swift out of scope) |
| lien-review-testbed (no JVM fixtures) | 93 → 93 | unaffected |

**Contradiction found vs. the plan's stated bar:** criterion 1 asked for
"orphan rate falls ≥50% relative" on JavaPoet/Klaxon. Measured relative
falls are 42.5% (JavaPoet) and 31.0% (Klaxon) — under the bar. I dug into
*why*: every remaining orphan on both corpora is either a non-source file, a
file with no top-level class/interface (top-level `fun`/`val` are
deliberately out of Phase-1 scope), or a real `*Test.java`/`*Test.kt` JUnit
class — and JUnit test classes are structurally zero-in-degree in every
language this repo's E2E suite covers (see Sinatra's own "expected
in-degree-0 targets" precedent). Concretely: **every JavaPoet production
file with a top-level type declaration now has at least one resolved
dependent** (0 production misses); Klaxon has exactly 2 (one is a
genuinely-unreferenced annotation type confirmed via corpus-wide grep, the
other is a real cross-package edge unreachable by *any* mechanism because
it's reached only through `import com.beust.klaxon.*`, an on-demand import).
The resolver is correct; the 50% bar as measured against the *whole* file
set undershoots because these two corpora are unusually test/leaf-heavy.
Raised `real-projects.test.ts` floors to the real measured numbers (see
below) rather than the original plan's now-outdated 9/2.

## Hand-audit (criterion 2): 44 sampled edges, Retrofit + OkHttp

24 Retrofit + 20 OkHttp real, newly-recovered edges read individually
against source. **42/44 correct (95.5%)**, clearing the ≥95% bar. Two false
positives found and worth naming honestly:
- `Phone.java` → `CrashingPhone.java` (retrofit wire converter test fixtures):
  matched only via a `toString()` debug-string literal (`"Phone{"`), not a
  real code reference — a plain identifier-boundary text match can't
  distinguish a string literal from code, the same category of residual risk
  `csharp-type-reference-signals.ts`'s own doc comment already accepts.
- `Request.kt` → `DnsValuesTest.kt` (okhttp): matched via `Dns.Request(...)`,
  which resolves to `Dns.kt`'s *own* nested `class Request`, not the
  top-level `okhttp3.Request`. G4 only excludes a candidate's own
  declaration; it doesn't (and structurally can't, from text alone) protect
  against a THIRD file's nested type sharing a target's simple name. Noting
  this as an explicit residual-risk shape alongside the on-demand-import one
  already in `inferred-dependent-mechanisms.ts`.

Both are exactly the class of miss `confidence: 'inferred'` exists to hedge,
not silent fabrication.

## Import-shadowing regression (criterion 4), real corpora

```
okhttp3/Cache.kt dependents (69 present) — ResponseBodyTest.kt present? false  ✓ ABSENT
```
`RxJava{,rxjava2,rxjava3}CallAdapterFactory.java → HttpException.java`:
checked against all 6 real `HttpException.java` declarations in the
corpus — **absent in every case** ✓, while the other 25 real HttpException
dependents (5 on `retrofit2.HttpException` + 6/6/4/2/2 on the five adapter
modules' own `HttpException`) all **remain present** ✓ — this dual
assertion is what actually proves G6 works rather than just firing on
everything.
`retrofit2/HttpException.java` ← `HttpExceptionTest.java`: **present** ✓
(the recall proof — package-locally unambiguous, corpus-wide ambiguous,
exactly the case the original corpus-wide gate would have dropped).

## Property tests (criteria 3 & 6)

- **P1** (brute-force oracle): `resolveJvmSamePackageDependentsBruteForce`
  scans the whole corpus with G2 checked explicitly rather than via the
  pre-narrowed `filesByPackage` index; asserted to agree *exactly* (not just
  "never drops a match") with the fast path across a multi-gate fixture.
- **P2** (post-hoc re-derivation, mandatory per the plan — the only property
  that would catch a dropped/reordered gate): for every produced edge, all
  six gates are independently re-derived from raw chunk content in the test
  file (not from the resolver's internal state) and reconfirmed.
- **P3** (subset containment): every result is a member of its own
  package-local candidate set.

## Other verification

- **Wall-clock (criterion 7):** `computeDependentCountsFromChunks` on OkHttp
  (682 indexed files — this repo's `performChunkOnlyIndex` count, vs. the
  spike's 514 `.kt`/`.java`-only count): ~60ms baseline → ~260ms with all
  three recovery tiers (the JVM tier is effectively the whole ~200ms delta,
  since OkHttp has no C#/Go files to trigger the other two). Real end-to-end
  `lien index` on OkHttp: **8.4s** for 682 files — the added cost is
  immaterial against that total.
- **Live MCP dogfood (criterion 8):** real `lien index` + real `get_dependents`
  call via the MCP protocol (stdio) against a fresh OkHttp clone —
  `okhttp3/Request.kt` → 69 dependents, `confidence: "inferred"`,
  `inferredVia: "jvm-same-package"`, and the caveat note correctly says
  *"its language, Java/Kotlin"* — not C# (the exact #1018 regression shape
  this was checked against verbatim).
- **Package-private corpus (criterion 9, no pass/fail bar):** Retrofit
  11/11 (100%) package-private top-level files got a derived package;
  OkHttp 361/362 (99.7%) — the one miss (`build-logic/.../BndBuildAction.kt`)
  has no `package` line in the source at all, confirmed via grep, not a
  chunker-coverage gap.
- **`index-state-matrix.test.ts` (criterion 10):** ran as-is; guard does not
  fire (49/49 pass) — this PR adds no new `createVectorDB` call site or MCP
  tool handler, so no update was needed.

## Gate chain

`npm run format:check && npm run lint && npm run typecheck && npm run build
&& npm test && lien delta` — all pass. `lien delta`: 0 new/crossed
complexity violations (2 pre-existing functions worsened but stayed under
threshold — `resolveDependents` time 18m→21m, `recoverDependentsForFile`
cognitive 2→4 — neither crosses).

## Test plan
- [x] `jvm-same-package-signals.test.ts` — 25 unit tests covering all six
  gates individually, plus P1/P2/P3 property tests and the
  content-vs-path-derived package cross-check (criterion 10's integration
  site 7), including the Kotlin-blindness of `javaPackageRelativePath`
  documented rather than unified (out of scope here).
- [x] Full parser/cli/core/review/action suites green (no regressions).
- [x] `inferred-dependent-mechanisms.test.ts` updated for the third
  mechanism's forcing-function string.
- [x] `real-projects.test.ts` floors raised to real measured numbers for
  JavaPoet (76, ~50% of 152) and Klaxon (112, ~50% of 224).

## Summary by CodeRabbit

* **New Features**
  * Improved dependency analysis for Java and Kotlin same-package relationships.
  * Added recovery when standard import-based analysis cannot identify dependents.
  * Added clearer attribution for partially resolved dependencies.

* **Bug Fixes**
  * Improved handling of package-local references, imports, aliases, nested declarations, and test sources.

* **Documentation**
  * Documented Java and Kotlin fallback behavior, resolution coverage, and limitations.

* **Tests**
  * Expanded coverage for JVM dependency resolution and updated project thresholds.

---

### Lien Review

| 🧠 mental load | 1 | +0 |

> [!NOTE]
> **Low Risk**
>
> This PR adds a well-scoped Java/Kotlin same-package dependent recovery resolver (`jvm-same-package-signals.ts`) and wires it into `findDependents` and batch dependent counting. The implementation mirrors the existing C#/Go recovery tiers, keeps the JVM index fresh per call (avoiding `lien serve` staleness), and marks recovered edges with `confidence: 'inferred'`. The six gates (package-local uniqueness, exact package match, derivable package, no nested same-name declaration, class/interface targets only, import-shadowing exclusion, and source-set direction) are implemented consistently with the documented JLS/Kotlin semantics, and the property tests (P1/P2/P3) plus real-corpus validation provide strong correctness evidence. No bugs producing wrong output or breaking callers were found in the changed code paths.
>
> - New `jvm-same-package-signals.ts` resolver with build/resolve split and brute-force oracle
> - Wired into `findDependents` via `enrichWithJvmSamePackageDependents` and into `dependent-count-index.ts` batch counts
> - Updated `inferred-dependent-mechanisms.ts` prose and exports in `index.ts`
> - Raised `real-projects.test.ts` dependency-edge floors for JavaPoet/Klaxon based on measured real-corpus numbers

> [!WARNING]
>

<sup>Reviewed by [Lien Review](https://lien.dev) for commit 46efb86. Updates automatically on new commits.</sup>

---

Attested: degraded:budget_starved · 695K/805K tokens


alfhen:
Addressed both findings from triage:

**1. Fixed (mandatory) — CRLF/trailing-comment blind spot in `SINGLE_IMPORT_LINE_RE`.**
CodeRabbit's diagnosis and proposed fix were exactly right. Widened the
trailing anchor to `[ \t]*;?[ \t]*(?:\/\/.*)?[ \t\r]*$`. Added two new
fixtures to `jvm-same-package-signals.test.ts` (a CRLF file, a trailing
`// comment` on the import line) that exercise the production regex via
`findJvmSamePackageDependents` (not a re-declared copy, per CodeRabbit's
note that the existing P2 property test's independent oracle couldn't
catch this by construction). Verified both new tests fail against the old
regex and pass against the fix (reverted locally, confirmed failure,
restored). Commit 46efb868.

**2. Investigated, deferred with a filed issue — per-call index rebuild in
`enrichWithJvmSamePackageDependents`.**
Confirmed reachable: `api-delta-cmd.ts`'s `enrichDeltas` loops over every
changed file *and* every changed exported symbol in one process
invocation, so a diff touching N zero-dependent Java/Kotlin symbols pays N
full `JvmSamePackageIndex` rebuilds. Measured cost: ~37-40ms/rebuild on a
real 682-file/9033-chunk corpus (OkHttp) — not severe per call, but linear
in diff size and avoidable.

Important context: this is **not new to this PR** — `enrichWithCSharpTypeReferenceDependents`
and `enrichWithGoRootPackageDependents` (already shipped, #930/#1039) have
the *identical* per-call rebuild pattern via their own single-target
wrappers (`findCSharpTypeReferenceDependents`/`findGoRootPackageDependents`).
The new JVM mechanism is at exact parity with its two siblings, not worse.

A proper fix means threading a shared, batch-scoped recovery-index bag
through `findDependents`'s public signature (or `ScanContext`) — mirroring
`dependent-count-index.ts`'s `RecoveryIndexes` precedent — which touches
all three enrichment functions, `findDependents`'s signature, and every
existing caller (`get-dependents.ts`, `annotate-cmd.ts`, `api-delta-cmd.ts`,
plus their tests). That's a genuine multi-call-site restructuring of a
widely-used core API, not a Phase-1-scoped fix, so I filed
#1101 rather than attempt it under this PR's scope, and scoped the issue
to cover all three mechanisms symmetrically (fixing only the JVM one here
would leave C#/Go with the same cost and just move the inconsistency
around).

Full local gate chain (format/lint/typecheck/build/test/`lien delta`)
re-run clean after the fix: parser 1945 tests, cli 1737, core+review 1725
(all +2 from the new fixtures, 0 regressions), delta exit 0.

Issue #1114, opened 2026-08-08
Kotlin AST extractor doesn't set parentClass on nested classes, inflating jvm-same-package-signals.ts's top-level declaration index


alfhen:
Noticed while implementing #1005 Phase 2 Item 1 (call-graph same-package tier). `jvm-same-package-signals.ts`'s `collectTopLevelDeclarations` restricts to top-level class/interface declarations via `!chunk.metadata.parentClass` — same-package visibility only applies to top-level types, per JLS/Kotlin semantics.

The Kotlin AST extractor, however, doesn't currently set `parentClass` on a nested class's own chunk the way the Java extractor does. This means a Kotlin nested class (e.g. inside a sealed hierarchy or a companion-adjacent inner class) is currently indistinguishable from a real top-level declaration in `jvm-same-package-signals.ts`'s index — it gets treated as eligible for G1'/G5 resolution when it structurally shouldn't be.

This is a real but small, out-of-scope-for-#1005-Phase-2 precision issue (not something the same-package tier's own logic can fix — it's upstream, in the Kotlin extractor itself). Filing to track it separately. Likely fix: audit `KotlinSymbolExtractor` (or wherever nested-class chunks are built) for parity with the Java extractor's `parentClass` assignment, then add a regression test to `jvm-same-package-signals.test.ts` asserting a Kotlin nested class is excluded from `collectTopLevelDeclarations`.

Issue #1113, opened 2026-08-08
Revisit blast-radius.ts's maxNodes/depth defaults now that JVM call-graph recall has improved


alfhen:
`packages/review/src/blast-radius.ts`'s `computeBlastRadius` defaults to `maxNodes: 30`, `depth: 2`. #1005 Phase 2 Item 1 added a Java/Kotlin same-package tier to the call graph (`dependency-graph.ts`'s `getCallers`), which measurably increases fan-in for popular declared types in JVM codebases — e.g. a real OkHttp dogfood run: `Cache`'s `getCallersTransitive(depth: 2)` went from 19 edges (not truncated) to 30 edges (truncated) after that change, purely from real same-package callers the call graph could not see before.

A new test (`packages/review/test/blast-radius.test.ts`, "AC10 ... a very-high-fan-in JVM type seed saturates the default maxNodes at hop 1") pins the CURRENT behavior: once hop-1 fan-in exceeds `maxNodes`, the walk truncates before reaching hop 2 at all for that seed, even for a hop-1 node that itself has a real, further caller one hop away.

This is a disclosed, not-fixed limitation, not a bug — the existing default trades recall for a bounded walk, and #1005 Phase 2 doesn't change that policy. Filing to ask: should `maxNodes`/`depth` be revisited for JVM specifically (or across the board) now that same-package recall has improved and can realistically saturate the budget on real, popular types? This is a policy question for `blast-radius.ts`'s owners (review-engine cost/precision trade-offs), not something to solve inside the same-package tier itself.

Issue #1112, opened 2026-08-08
findDependents's JVM same-package recovery: make additive now that #1111 removed the rebuild-cost argument


alfhen:
`enrichWithJvmSamePackageDependents` in `packages/parser/src/dependency-analyzer.ts` only fires when `symbol` is undefined AND the import graph found literally zero dependents (mirroring the C#/Go recovery tiers' precondition). That gate predates #1111, which made the JVM same-package index build-once-per-batch via `RecoveryIndexes` — the original cost argument for "only ever try this as an expensive last resort" is largely gone now.

#1005 Phase 2 Item 1 (see that PR) added an *additive* same-package tier to the call-graph (`dependency-graph.ts`'s `getCallers`), which fires unconditionally for any JVM type-symbol query regardless of whether the import graph already found something — deliberately NOT mirroring `findDependents`'s zero-only gate, because the two mechanisms' preconditions are exact inverses by design (see that PR's AC6).

Worth asking: should `findDependents`'s file-level recovery become additive too (recover same-package dependents even when the import graph already found some), now that the cost argument for "only as a last resort" is weaker? This is a policy question with real trade-offs (e.g. `dependentCount`/`riskLevel` semantics for `get_dependents`, and whatever `#1057`'s dependent-attribution-incomplete caveat wiring assumes about the zero-only precondition) — filing to track the question, not proposing an immediate answer.

PR #1117, opened 2026-08-08
feat(parser,cli): add Kotlin same-package test-association mechanism (#1005 Phase 2, Item 2)

alfhen:
## Summary

Item 2 of #1005 Phase 2 (follows #1115, merged): adds a Kotlin same-package
test-association mechanism, shipped in **both** consumers in this one PR:

- `packages/parser/src/test-associations.ts`'s `findTestAssociationsFromChunks`
  (the shared engine — feeds `lien annotate`, blast-radius's test-coverage
  risk computation, and the agent-review plugin's coverage checks).
- `packages/cli/src/mcp/handlers/get-files-context.ts`'s `findTestAssociations`
  (the `get_files_context` MCP tool's own separate implementation — the same
  parallel-implementation pattern #1040 already established for C#).

Like Java's own same-package test convention (#925) and C#'s
enclosing-namespace one (#1040), a Kotlin test class commonly lives in the
same package as its subject with **no import statement connecting them at
all**. Kotlin's same-package visibility rule needs none.

### Design

- **Mechanism (a) only** — no path-regex extension of Java's mechanism was
  added. The plan's own measurement found that contributes little beyond
  what this content-derived tier alone finds, is 71-100% redundant with it
  wherever it does fire, and never fires on heavily-multiplatform-Kotlin
  layouts. Kotlin gets exactly one mechanism, gated strictly to
  `detectLanguage(filepath) === 'kotlin'` — Java keeps its own, separate
  mechanism untouched.
- **Reuses Phase 1's FILE-LEVEL `resolveJvmSamePackageDependents`** (#1100),
  deliberately **not** Item 1's per-type `resolveJvmSamePackageDependentsForType`
  — "which tests exercise this file" is inherently a file-level question,
  and both consumers (`hasTestCoverage` in blast-radius, `testAssociations`
  in `get_files_context`) are file-keyed, not type-keyed. Stated explicitly
  in both files' doc comments, not left for a future reader to re-derive.
- **Path canonicalization, explicit and tested (AC8)**:
  `resolveJvmSamePackageDependents` requires the caller's filepath to be the
  *exact* `chunk.metadata.file` string used when its index was built.
  `findTestAssociationsFromChunks` is reached with several different path
  forms (some workspace-relative with an explicit `rootDir`, at least one —
  `insights/chunk-complexity.ts` — with **no root passed at all**, defaulting
  to `process.cwd()`). A form mismatch would otherwise silently return `[]` —
  exactly the "clean-looking zero" this repo's index-state-honesty policy
  forbids. **This bug already exists, uncanonicalized, in this same file's
  C# collector** (`collectCSharpNamespaceTests`) — deliberately not copied
  forward:
  - `test-associations.ts`: canonicalizes only the *query* side, via a
    `getCanonicalPath`-keyed `canonical → raw` lookup
    (`buildRawJvmFileByCanonical`), so the module's established raw-form
    output convention (every other tier here returns raw
    `chunk.metadata.file` strings) stays consistent.
  - `get-files-context.ts`: mirrors that file's **already-correct** C#
    precedent (`buildCSharpTestIndexFromChunks`/`collectCSharpNamespaceTestFiles`),
    which canonicalizes the whole chunk array up front — that file's
    established convention is full canonicalization throughout, so the new
    tier's output is canonical-form too, consistent with its siblings.
- **No package-size heuristic.** An earlier concern was that a
  single-package-dominated project would make the package-scoping gate
  nearly vacuous. Measured: fabrication rate stayed ~0% regardless of
  package concentration — the real bound is the underlying resolver's
  word-boundary text match against the production file's declared type
  name, not the candidate pool size. Not added.

### What does NOT change

- Java's existing path-derived mechanism (`java-same-package-tests.ts`) —
  untouched, still Java-only.
- Every other tier in both files (Go, direct-import, C#) — pure addition.

## Acceptance criteria

8. **Path-form honesty** — new tests in both `test-associations.test.ts`
   and `server.get-files-context.test.ts`: a chunk with an ABSOLUTE
   `metadata.file`, queried with a workspace-relative form of the same
   file, still resolves the association (not a silent `[]`). A negative
   control (`test-associations.test.ts`) confirms a genuinely different
   file still honestly returns nothing — the fix isn't "always non-empty."
9. **Index-state-matrix stays green, confirmed not assumed** —
   `packages/cli/test/integration/index-state-matrix.test.ts` (49 tests)
   ran clean; this PR adds no `createVectorDB` call site or MCP handler.

Also verified: the Java-only gating actually discriminates (a test proves
the underlying JVM resolver *would* find a same-package Java association
if the language gate were removed, then proves the gate suppresses it —
not a vacuous "no signal at all" negative).

## Dogfood evidence (verbatim, real Klaxon clone)

Indexed a fresh clone of `cbeust/klaxon` with this branch's built CLI.

**Before this PR** (`git stash`, same corpus, same command):

```
$ lien annotate klaxon/src/main/kotlin/com/beust/klaxon/Klaxon.kt
⚠ Lien: Klaxon.findConverterFromClass cognitive 14/15 — avoid adding complexity here; prefer extraction.
Lien impact for klaxon/src/main/kotlin/com/beust/klaxon/Klaxon.kt:
  • 60 files import this — ...; risk: medium (7 production callers, 7 untested).
  • No test coverage.
```

**After this PR** (same command, same index):

```
$ lien annotate klaxon/src/main/kotlin/com/beust/klaxon/Klaxon.kt
⚠ Lien: Klaxon.findConverterFromClass cognitive 14/15 — avoid adding complexity here; prefer extraction.
Lien impact for klaxon/src/main/kotlin/com/beust/klaxon/Klaxon.kt:
  • 60 files import this — ...; risk: medium (7 production callers, 7 untested).
  • Test coverage: klaxon/src/test/kotlin/com/beust/klaxon/BindingAdapterTest.kt, klaxon/src/test/kotlin/com/beust/klaxon/BindingTest.kt (+51 more).
```

`Klaxon.kt` is the library's central class — before this PR, `lien
annotate` reported it as having **zero** test coverage, even though it is
in fact exercised by essentially the entire test suite. This is exactly the
"clean-looking zero" the index-state-honesty policy is about: not a bug in
the *rendering*, a real blind spot in what the tool could see.

Also confirmed via the real `get_files_context` MCP tool (started via
`lien serve`, queried over stdio with `@modelcontextprotocol/sdk`'s client):
its `testAssociations` field for the same file returns the identical 53
files.

## Local gate chain (all green)

```
npm run format:check   ✅
npm run lint            ✅ (0 errors, 42 pre-existing warnings, none in touched files)
npm run typecheck       ✅
npm run build            ✅
npm test                 ✅ (parser 1962, core 448, review 1726, cli 1739, action 41 — all passing)
lien delta                ✅ exit 0 — no threshold crossings (two advisory Halstead "time" metric
                              upticks on findTestAssociationsFromChunks/findTestAssociations, both
                              well under threshold)
```

## Known, disclosed residual (not a blocker)

Roughly 1-2% of newly-found associations may rest on a single textual
occurrence that happens to sit inside a comment or KDoc block rather than
live code — a real, small limitation shared with Phase 1's underlying
word-boundary-match gate (`jvm-same-package-signals.ts`). Not gated on or
fixed here, per the plan's own explicit scope call.

## Follow-up issues

All three follow-ups from the overall #1005 Phase 2 plan were already filed
against Item 1's PR (#1115): #1112, #1113, #1114. Nothing new to file for
Item 2 specifically.

---

### Lien Review

➡️ **Stable** - 4 pre-existing issues in touched files (none introduced).
| 🔀 test paths | 1 | +0 |
| 🧠 mental load | 2 | +0 |

> [!NOTE]
> **Low Risk**
>
> This PR adds a Kotlin same-package test-association mechanism in two parallel implementations: the shared parser engine (`findTestAssociationsFromChunks`) and the `get_files_context` MCP tool (`findTestAssociations`). It reuses the existing file-level JVM same-package resolver, gates it strictly to Kotlin queries, and addresses path-form mismatches through explicit canonicalization. The change is additive, well-documented, and covered by tests for the happy path, the Kotlin-not-Java gate, import-based composition, and the canonicalization edge case.
>
> - Adds Kotlin same-package test association to `findTestAssociationsFromChunks` with query-side canonicalization and raw-form output preservation.
> - Adds a parallel Kotlin tier to `get_files_context`'s `findTestAssociations`, canonicalizing the full chunk array upfront to match the file's existing C# convention.
> - Updates Kotlin project expectations in `real-projects.test.ts` to reflect recovered test associations.

<sup>Reviewed by [Lien Review](https://lien.dev) for commit 0f83e57. Updates automatically on new commits.</sup>

---

### DIFF
--- a/.changeset/kotlin-test-associations.md
+++ b/.changeset/kotlin-test-associations.md
@@ -1,0 +1,10 @@
+---
+'@liendev/parser': minor
+'@liendev/lien': patch
+---
+
+Add a Kotlin same-package test-association mechanism (#1005 Phase 2, Item 2), shipped in both `@liendev/parser`'s `findTestAssociationsFromChunks` (the shared engine — feeds `lien annotate`, blast-radius test-coverage risk, and the agent-review plugin) and `@liendev/lien`'s `get_files_context` MCP tool (its own separate implementation, mirroring the existing C# tier there).
+
+Like Java's own same-package test convention, a Kotlin test class commonly lives in the same package as its subject with no import connecting them at all — Kotlin's same-package visibility rule needs none. This reuses Phase 1's file-level `resolveJvmSamePackageDependents` (#1100), gated strictly to Kotlin (Java keeps its existing, separate path-based mechanism), and explicitly canonicalizes the query path against the index before resolving — a mismatched path form now resolves correctly instead of silently returning zero associations.
+
+Measured against a real Klaxon (Kotlin) clone: `lien annotate` on `Klaxon.kt`, the library's central class, went from reporting "No test coverage" to 53 real, same-package test files.
--- a/packages/cli/src/mcp/handlers/get-files-context.ts
+++ b/packages/cli/src/mcp/handlers/get-files-context.ts
@@ -20,6 +20,8 @@ import {
   hasEnclosingNamespaceAccess,
   buildCSharpTypeReferenceIndex,
   resolveCSharpTypeReferenceDependents,
+  buildJvmSamePackageIndex,
+  resolveJvmSamePackageDependents,
   MAX_CHUNKS_PER_FILE,
   DEFAULT_COMPLEXITY_DELTA_THRESHOLDS,
   type GoTestCandidate,
@@ -27,6 +29,7 @@ import {
   type JavaTestCandidate,
   type JavaTestDirIndex,
   type CSharpTypeReferenceIndex,
+  type JvmSamePackageIndex,
 } from '@liendev/parser';
 import type { SearchResult, VectorDBInterface } from '@liendev/core';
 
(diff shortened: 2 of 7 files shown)


alfhen:
Follow-up commit (0f83e571): the E2E \"collapse-to-zero detector\" for Klaxon caught this PR closing a documented KNOWN GAP -- \`KNOWN_ZERO_TESTASSOC_LANGUAGES\` asserted Kotlin's test-association count as an exact-zero tripwire, and this PR correctly makes that assertion fail (142 associations found, not 0). Per that test's own doc comment ("if this fails, the gap likely just closed -- update this project's config to a real `expectedMinTestAssociations` floor instead of re-asserting zero"), removed `'kotlin'` from the tripwire set and set Klaxon's floor to 70 (~50% of the measured 142, matching this project's own existing `expectedMinDependencyEdges` floor convention). Also fixed a now-stale doc comment in `ast/languages/kotlin.ts` that claimed no verified Kotlin test-source recovery exists.

Verified locally: Klaxon's full 18-test E2E suite passes (including the new floor), SwiftyJSON's zero-tripwire is unaffected, and JavaPoet (Java) is unaffected.

